PR #16, opened 2019-03-22
Sagonzal/update nimbus oauth2 sdk

sangonzal:
Nimbusds Oauth2 SDK introduced a breaking change in v6.0 to support multi-valued query parameters. This is a fix for that change. Same as https://github.com/AzureAD/azure-activedirectory-library-for-java/pull/259


SomkaPe:
:shipit:

PR #27, opened 2019-04-19
Clean up dev samples. Update scrub logic for telemetry Pii

sangonzal:


src/main/java/com/microsoft/aad/msal4j/Event.java
```diff
@@ -68,10 +69,17 @@ static String scrubTenant(URI uri){
             return null;
         }
 
-        //TODO should be updated when B2C is added, since tenant could be in different place
         String[] segment = uri.getPath().split("/");
         if(segment.length >= 2){
-            segment[1] = TENANT_PLACEHOLDER;
+            if(segment[1].equals("tfp")){
+                segment[2] = TENANT_PLACEHOLDER;
```
henrik-me: >segment[2] = TENANT_PLACEHOLDER; [](start = 16, length = 32)

if length is 2 wouldn't this result in an error?

src/samples/public-client/IntegratedWindowsAuthFlow.java
```diff
@@ -40,16 +40,17 @@ public static void main(String args[]) throws Exception {
...
+        IntegratedWindowsAuthenticationParameters parameters =
+                IntegratedWindowsAuthenticationParameters.builder(
+                        Collections.singleton(TestData.GRAPH_DEFAULT_SCOPE), TestData.USER_NAME)
+                        .build();
 
-        AuthenticationResult result = future.get();
+        Future<AuthenticationResult> future = app.acquireToken(parameters);
 
-        return result;
+        return future.get();
```
henrik-me: >future.get() [](start = 15, length = 12)

is this good for debugging? Wouldn't you prefer to have the result in the variable instead?

src/samples/public-client/UsernamePasswordFlow.java
```diff
@@ -66,20 +65,17 @@ private static void getAccessTokenFromUserCredentials() throws Exception {
 
             } catch (MalformedURLException e) {
                 e.printStackTrace();
+                throw new RuntimeException();
```
henrik-me: Why? can't the original exception be re-thrown? (alternatively could this be made the inner exception)

henrik-me:
Great refactorings!

src/main/java/com/microsoft/aad/msal4j/Event.java
sangonzal: Yes, but we are implicitly checking that segment[2] will not be null by checking that segment[1] == "tfp" (if segment[1] == "tfp" then segment["2"] will never be null). This is a good point though, it is probably better to have the explicit check anyways. I will update.

src/samples/public-client/IntegratedWindowsAuthFlow.java
sangonzal: You're right, changing it back.

src/samples/public-client/UsernamePasswordFlow.java
sangonzal: Without going too much into detail, you [can't throw checked exceptions from a lambda (or streams) without a lot more boilerplate code](https://stackoverflow.com/questions/27644361/how-can-i-throw-checked-exceptions-from-inside-java-8-streams). I didn't think it was worth it here. The simpler alternative is to just throw an unchecked exception.

PR #47, opened 2019-06-01
fixing issues found with spotbugs

SomkaPe:
pom.xml - removing unused dependencies, adding scope "test" to test dependencies, removing profiles, 
changing find bugs plugin to spot bugs (find bugs is not maintained any more), formatting 

fixing some issues found with spot bugs:
- exposure of mutable fields in immutable objects
- relaying of default encoding - specifying UTF-8 explicitly
- having reference to non Serializable object in Serializable one (AuthenticationResult)
- fixing bug with incorrect calculation of base64EncodedSha256Hash   in hashPii function
- fixing integration tests config
- fixing compliance warning


pom.xml
```diff
@@ -1,289 +1,249 @@
...
+                    <execution>
+                        <id>attach-sources</id>
+                        <goals>
+                            <goal>jar</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>com.github.spotbugs</groupId>
```
henrik-me: >com.github.spotbugs [](start = 25, length = 19)

did you validate that we can use this? Also assuming there is automatic component governance setup for this project? Otherwise pls. ensure this component is properly registered and removed components are de-registered.

src/main/java/com/microsoft/aad/msal4j/UserNamePasswordParameters.java
```diff
@@ -46,6 +46,10 @@
     @NonNull
     private char[] password;
 
+    public char[] password(){
```
henrik-me: >password [](start = 18, length = 8)

ClonePassword?

pom.xml
```diff
@@ -1,289 +1,249 @@
...
-			<artifactId>commons-codec</artifactId>
-			<version>1.11</version>
-		</dependency>
-		<dependency>
-			<groupId>org.apache.commons</groupId>
-			<artifactId>commons-lang3</artifactId>
-			<version>3.5</version>
-		</dependency>
-		<dependency>
-			<groupId>com.google.guava</groupId>
```
sangonzal: Guava was added it because of the selenium tests: https://github.com/SeleniumHQ/selenium/issues/3880. Please make sure integration tests still work both locally and on Azure Devops

pom.xml
SomkaPe: it is plugin - dev tool, yes automatic governance is set up

src/main/java/com/microsoft/aad/msal4j/UserNamePasswordParameters.java
SomkaPe: to keep it immutable we should not expose internal representation which is array, so mutable

pom.xml
SomkaPe: i have fixed some issued with integ tests, like usage of wrong redirect url(wrong port) and test with device code flow, but one test failing - acquireTokenWithAuthorizationCode_B2C_Google - it was never successfully executed on azure devops

sangonzal: The B2C_Google test failing on Azure Devops is probably due to Google flagging the test account for "suspicious activity" since the request is coming from completely different place/different IP. When this happens, they ask you to change the password. I'm not sure if there is much that we can do there.

src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
```diff
@@ -305,6 +304,7 @@ private String acquireAuthorizationCodeAutomated(
                 throw new RuntimeException("Could not start TCP listener");
             }
             runSeleniumAutomatedLogin(labUserData, authorityType);
+            String page  = seleniumDriver.getPageSource();
```
sangonzal: Doesn't seem to be used anywhere.

PR #48, opened 2019-06-05
exposing rt grant flow as public api

SomkaPe:


henrik-me:
@sangonzal @SomkaPe : this is not just for migration purposes. It's both for migration and integration scenarios. I believe both Python and .NET have a reasonable description of this, it may be worth looking at the words used there.

src/main/java/com/microsoft/aad/msal4j/IClientApplicationBase.java
```diff
@@ -92,6 +92,15 @@
...
     CompletableFuture<IAuthenticationResult> acquireToken(AuthorizationCodeParameters parameters);
 
+    /**
+     * Acquires a security token from the authority using a Refresh Token previously received.
+     *
+     * @param parameters#refreshToken Refresh Token to use in the refresh flow.
+     * @param parameters#scopes       scopes of the access request
+     * @return A {@link CompletableFuture} object representing the {@link IAuthenticationResult} of the call.
+     */
+    CompletableFuture<IAuthenticationResult> acquireToken(RefreshTokenParameters parameters);
```
henrik-me: acquireTokenByRefreshToken + see comments in other libraries which also have aka.ms links to additional details.

SomkaPe: thanks, for pointing out , have copy pasted

PR #68, opened 2019-06-22
Pesomka/client cred silent

SomkaPe:
- silent flow for client credential grant 
- use tenant id from request authority if id token not returned
- use  AutoCloseable  for cache Aspect application


src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
```diff
@@ -72,8 +72,14 @@ private void assertAcquireTokenCommon(String clientId, IClientCredential credent
...
-        // TODO AuthenticationResult should have an getAccountInfo API
-        // Assert.assertEquals(labResponse.getUser().getUpn(), result.getAccountInfo().getUsername());
+
+        result = cca.acquireTokenSilently(SilentParameters
+                .builder(Collections.singleton(KEYVAULT_DEFAULT_SCOPE))
+                .build())
+                .get();
+
+        Assert.assertNotNull(result);
+        Assert.assertNotNull(result.accessToken());
```
sangonzal: Might also be worth writing a test that has multiple tokens in cache for same clientId and tenant but different scopes and check that the right token is returned.

src/main/java/com/microsoft/aad/msal4j/TokenRequest.java
```diff
@@ -51,12 +52,15 @@
...
+    private final URL tokenEndpointUrl;
     private final MsalRequest msalRequest;
     private final ServiceBundle serviceBundle;
 
-    TokenRequest(final URL url, MsalRequest msalRequest, final ServiceBundle serviceBundle) {
-        this.url = url;
+    TokenRequest(Authority requestAuthority, MsalRequest msalRequest, ServiceBundle serviceBundle)
+            throws MalformedURLException {
+        this.requestAuthority = requestAuthority;
+        this.tokenEndpointUrl = new URL(requestAuthority.tokenEndpoint());
```
henrik-me: Do you need both here, Perhaps authority.tokenEndpointUrl could do this instead?

SomkaPe: good point, changed

src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
SomkaPe: good point - extended to few scopes and found serious bug

PR #69, opened 2019-06-25
Sagonzal/exceptions

sangonzal:
- Split AuthenticationException into AuthenticationClientException and AuthenticationServiceException
- Expose http headers in AuthenticationServiceException- #59 
- Expose suberror code in AuthenticationServiceException (invalid_grant and interaction_required classification) - #61 
- Remove ClaimsChallengeException.  
   - developers should instead check if AuthenticationServiceException contains claims (aligned with .NET)
 
Open question: 
Should we rename:
- AuthenticationException -> MsalException
- AuthenticationServiceException -> MsalServiceException
- AuthenticationClientException -> MsalClientException


src/main/java/com/microsoft/aad/msal4j/AuthenticationClientException.java
```diff
@@ -23,30 +23,24 @@
...
 /**
- * The exception type thrown when a claims challenge error occurs during token acquisition.
+ *  Exception type thrown when and error occurs that is local to the library or the device.
  */
-public class ClaimsChallengeException extends AuthenticationException {
+public class AuthenticationClientException extends AuthenticationException{
 
     /**
-     * claims challenge value
+     * Initializes a new instance of the exception class with a specified error message
```
SomkaPe: probably not "error message" but instance of Throwable

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
```diff
@@ -0,0 +1,115 @@
...
+     */
+    private Integer statusCode;
+
+    /**
+     * Status message returned from the http layer
+     */
+    private String statusMessage;
+
+    /**
+     * An ID that can used to piece up a single authentication flow.
```
SomkaPe: "can be used" probably

src/main/java/com/microsoft/aad/msal4j/OAuthHttpRequest.java
```diff
@@ -94,16 +97,26 @@ HTTPResponse createResponse(final HttpURLConnection conn, final String out)
                     + e.getMessage(), e);
         }
 
-        response.setCacheControl(conn.getHeaderField("Cache-Control"));
-        response.setPragma(conn.getHeaderField("Pragma"));
-        response.setWWWAuthenticate(conn.getHeaderField("WWW-Authenticate"));
+        Map<String, List<String>> headers = conn.getHeaderFields();
+        for(Map.Entry<String, List<String>> header: headers.entrySet()){
+
+            if(header.getKey() == null || StringHelper.isBlank(header.getKey())){
```
SomkaPe: check for null is included into StringHelper.isBlank

sangonzal:
@SomkaPe @henrik-me any opinions on the naming of the exceptions?

src/test/java/com/microsoft/aad/msal4j/TokenRequestTest.java
```diff
@@ -101,13 +80,80 @@ public void executeOAuthRequest_SCBadRequestErrorInteractionRequired_ClaimsChall
...
+        EasyMock.expect(msalOAuthHttpRequest.send()).andReturn(httpResponse).times(1);
+
+        PowerMock.replay(request, msalOAuthHttpRequest);
+
+        try {
+            request.executeOauthRequestAndProcessResponse();
+            Assert.fail("Expected AuthenticationServiceException was not thrown");
+        } catch (AuthenticationServiceException ex) {
+            Assert.assertEquals(claims.replace("\\", ""), ex.claims());
+            Assert.assertEquals(ex.subError(), "");
```
henrik-me: >Assert.assertEquals(ex.subError(), ""); [](start = 12, length = 39)

I dont' see any tests really validating the sub error. Also the subError is not what should be exposed, the exposed property is Classification and it will only be a subset of the suberrors. Please work with Bogdan on these details.

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
```diff
@@ -0,0 +1,115 @@
...
+import lombok.experimental.Accessors;
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Exception type thrown when service returns an error response or other networking errors occur.
+ */
+@Accessors(fluent = true)
+@Getter
```
henrik-me: >Getter [](start = 1, length = 6)

subError is supposed to stay private and a subset of errors should be exposed via the Classification property

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
```diff
@@ -0,0 +1,115 @@
...
+public class AuthenticationServiceException extends AuthenticationException{
```
henrik-me: >AuthenticationServiceException [](start = 13, length = 30)

Some of the errors should probably be categorized as client side / UiRequiredExceptions and not ServiceExceptions.

src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
```diff
@@ -23,20 +23,10 @@
...
-
-    AuthenticationErrorCode(String errorCode){
-        this.errorCode = errorCode;
-    }
-
-    @Override
-    public String toString(){
-        return errorCode;
-    }
+    public final static String AUTHORIZATION_PENDING = "authorization_pending";
```
bgavrilMS: Consider providing comments on these strings, in a format similar to "What Happens?" and "Mitigation".

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
bgavrilMS: Yes, namely "invalid_grant" service exceptions are exposed as UiRequiredServiceExceptions. Currently MSAL.NET treats all "invalid_grant" exceptions as UiRequiredExceptions, but I have a PR to filter some of these out (e.g. suberror "client_mismatch", which is a FOCI related suberror, is not a UiRequiredException).

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
```diff
@@ -0,0 +1,115 @@
...
+
+    /**
+     * The protocol error code returned by the service
+     */
+    private String errorCode;
+
+    /**
+     * More specific error
+     */
+    private String subError;
```
bgavrilMS: As Henrik mentioned, we don't expose this because Evo does not guarantee that these sub errors will remain unchanged.

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
bgavrilMS: Consider having a factory that decides which type of exception needs to be thrown - ServiceException or UiRequiredException.

src/main/java/com/microsoft/aad/msal4j/TokenRequest.java
```diff
@@ -139,40 +130,50 @@ AuthenticationResult executeOauthRequestAndProcessResponse()
...
-                    throw new ClaimsChallengeException(
-                            errorResponse.toJSONObject().toJSONString(),
-                            getClaims(httpResponse.getContent()));
-                } else {
-                    String telemetryErrorCode = StringHelper.isBlank(errorObject.getCode()) ?
-                            AuthenticationErrorCode.UNKNOWN.toString() :
-                            errorObject.getCode();
+                // Some invalid_grant or interaction_required subError codes returned by
+                // the service are not supposed to be exposed to customers
+                if(errorResponse.error() != null &&
```
bgavrilMS: The TokenRequest class does too much, consider moving this logic to another object, for example the ServiceException factory can take on this responsability.

src/main/java/com/microsoft/aad/msal4j/TokenRequest.java
```diff
@@ -139,40 +130,50 @@ AuthenticationResult executeOauthRequestAndProcessResponse()
...
-                }
+                throw new AuthenticationServiceException(
+                        errorResponse,
+                        httpResponse.getHeaderMap());
             }
             return result;
         }
     }
 
+    private ErrorResponse filterSubErrorCode(ErrorResponse errorResponse){
```
bgavrilMS: The spec I've used to do this work mentioned having a "Classification" property that exposes some of the sub-errors. We don't expose sub_error in general. 

Also, for discoverability, the Classification property will return an enum instead of a string constant.

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
sangonzal: @bgavrilMS I believe the service also returns "interaction_required". Are you treating these as UiRequiredExceptions as well?

bgavrilMS: @sangonzal - yes, sorry, let me clarify. There are 2 cases where MSAL gets an error from the server and throws UiRequiredException: 

- Whenever it interacts with Evo and Evo responds with error code `invalid_grant`
- In the authorization flow, when the auth code url contains the param `error=login_required`

The MSAL library also throws `UiRequiredException` based on client-side logic, as follows: 

- in `AcquireTokenSilent` if the account / login_hint is null, or if a login_hint is used but no matching Account / too many Accounts are in the cache for that hint. 
- If no cache is configured 
- Some web UIs also throw this (should not apply to you, the system browser ui doesn't throw this)

bgavrilMS: For the Authorization flow, that error will be sent when either: 

- Prompt.Never was passsed, but interaction is actually required
- An error occurs during silent auth that prevented the authentication to complete in time.

sangonzal:
@bgavrilMS Thanks for all of the detailed feedback. I have addressed all of the main issues you have raised, and have aligned this implementation with the one in .NET. I have a couple of questions which I will follow up with offline.

src/main/java/com/microsoft/aad/msal4j/AuthenticationServiceException.java
henrik-me: +1 : Seems like a missing word as Petro indicates

---
In reply to: [299208121](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/69#discussion_r299208121) [](ancestors = 299208121)

src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
henrik-me: imo, a must as these are public consts. Please document the usages, meaning and mitigations.

---
In reply to: [300341764](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/69#discussion_r300341764) [](ancestors = 300341764)

src/main/java/com/microsoft/aad/msal4j/MsalClientException.java
```diff
@@ -0,0 +1,46 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
+//
```
henrik-me: nit: short form instead?

src/main/java/com/microsoft/aad/msal4j/MsalServiceException.java
```diff
@@ -0,0 +1,107 @@
+// Copyright (c) Microsoft Corporation.
+// All rights reserved.
+//
+// This code is licensed under the MIT License.
```
henrik-me: nit: short form

src/main/java/com/microsoft/aad/msal4j/ServiceExceptionClassification.java
```diff
@@ -0,0 +1,66 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * Details about the cause of an {@link MsalInteractionRequiredException}, giving a hint about the
+ * user can expect when they go through interactive authentication
+ */
+public enum ServiceExceptionClassification {
```
henrik-me: >ServiceException [](start = 12, length = 16)

is this the right name? Why ServiceException ? Why not InteractionRequiredClassification ?

henrik-me:
one naming question which should be thought of before closing

henrik-me:
Not sure which exceptions. IMO, I like calling it InteractionRequired and not UIInteractionRequired, though this may be misleading as some of them we should really show them the system browser... otherwise using names that are used in other libraries makes the most sense to me for the exception naming (seems like you are generally following that pattern)

---
In reply to: [507772688](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/69#issuecomment-507772688) [](ancestors = 507772688)

src/main/java/com/microsoft/aad/msal4j/ServiceExceptionClassification.java
sangonzal: You are right, it should be InteractionRequiredReason, as per Bogdan's latest email.

PR #79, opened 2019-07-13
making token cache thread safe

SomkaPe:


src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -16,12 +18,10 @@
...
-    protected static final int MIN_ACCESS_TOKEN_EXPIRE_IN_SEC = 5*60;
+    static final int MIN_ACCESS_TOKEN_EXPIRE_IN_SEC = 5*60;
+
+    private ReadWriteLock lock = new ReentrantReadWriteLock();
 
-    /**
-     *  Constructor for token cache
-     * @param tokenCacheAccessAspect {@link ITokenCacheAccessAspect}
-     */
     public TokenCache(ITokenCacheAccessAspect tokenCacheAccessAspect) {
```
sangonzal: Please include comments for public constructors

SomkaPe: added

PR #85, opened 2019-07-26
Update README

sangonzal:


README.md
```diff
@@ -16,13 +16,28 @@ Current version - 0.5.0-preview
 You can find the changes for each version in the [change log](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/master/changelog.txt).
 
 The library is currently in preview. During the preview we reserve the right to make changes to the API, cache format, and other mechanisms of this library without notice which you will be required to take along with bug fixes or feature improvements
+
+###Maven
```
jmprieur: You probably need a space between ### and Maven (same below for Graddle)
@sangonzal

sangonzal: @jmprieur You are right - I realized this right after merging and immediately made the change. If you take a look at dev you'll see "maven" and "gradle" both have spaces.

Issue #82, opened 2019-07-17
Fix AT cache miss for silent call with common authority


SomkaPe:
Currently tokens are cached with tenant id which issued them, this is an issue when "common" authority is used - cached AT will not be found during silent call, so RT will be used and redundant call to STS executed.

PR #86, opened 2019-08-01
using tenant from request for caching

SomkaPe:
Change for https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/82

suggested fix  is using for caching tenant id from request Authority, so silent call with same Authority will return previously cached AT.


henrik-me:
This is an important issue to address and deserves an issue describing the scenario you are addressing, as that will allow people to search and fine and comment on the scenario. Also pls. ensure those comments are in the release notes.

SomkaPe:
added link to corresponding issue

PR #118, opened 2019-10-31
Make telemetryConsumer internal

sangonzal:


sangonzal:
> I'm a bit surprised that no tests needs update. Assuming it's because the interface is not used externally?

Yes

Issue #115, opened 2019-10-29
Allow user to pass in a custom HTTP Client


navyasric:

sangonzal:
[Instructions on how to configure Http client](https://github.com/AzureAD/microsoft-authentication-library-for-java/wiki/Configure-Http-Client)

PR #129, opened 2019-11-15
Let users configure HTTP client

sangonzal:
Addresses #115 

There are lot of changes - happy to sit in person to outline what they are and answer any questions.


src/integrationtest/java/com.microsoft.aad.msal4j/ApacheHttpClientAdapter.java
```diff
@@ -0,0 +1,93 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+import org.apache.http.Header;
+import org.apache.http.client.methods.CloseableHttpResponse;
```
henrik-me: nit: license

src/main/java/com/microsoft/aad/msal4j/IHttpClient.java
```diff
@@ -0,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * Interface to be implemented when configuring http client for {@link IPublicClientApplication} or
+ * {@link IConfidentialClientApplication}.
+ */
+public interface IHttpClient {
```
henrik-me: >IHttpClient [](start = 17, length = 11)

Seems like a lot of the methods in the DefaultHttpClient implementation is not available in this interface, is that on purpose?

sangonzal: Yea, those are implementation details . For customers configuring their own IHttpClient, all we care about is that they `send` the `HttpRequest` that we pass to them and they return an `IHttpResponse` - everything else is implementation details we are not concerned with. 

If no HTTP client is configured, we use JDK `HttpsUrlConnection` for making HTTP calls (in `DefaultHttpClient`). Everything other than `send` are implementation details specific to JDKs `HttpsUrlConnection`

src/integrationtest/java/com.microsoft.aad.msal4j/ApacheHttpClientAdapter.java
sangonzal: License header? think you might have missed it, it's already at the top of this file

src/main/java/com/microsoft/aad/msal4j/HttpUtils.java
```diff
@@ -0,0 +1,25 @@
...
+            return null;
+        }
+
+        List<String> headerValue = headers.get(headerName);
+
+        if(headerValue == null || headerValue.isEmpty()){
+            return null;
+        }
+
+        return headerValue.get(0);
```
jianghaolu: This seems a little confusing - in the case of an HTTP header where there are multiple values, e.g.: `x-ms-clitelem: 1,0,0,,` this will only return `1`. 

One way to work around this in my adapter is to put `1,0,0,,` in one header - but if so, why not just take a header parameter of type `Map<String, String>` instead?

sangonzal: @jianghaolu  Thanks for reviewing, this is a good point. I had originally given this some thought and decided to go with `Map<String, List<String>>` as this is what some http clients seem to expose as the type for headers (for example HttpsURLConnection). 

I missed this when testing as it seems that most headers returned by AAD are returned as strings and not lists (for example when testing with HttpsUrlConnection, OkHttpClient, and Apache's http client `x-ms-clitelem: 1,0,0,,` is  returned as `["1,0,0,,"]` and not `["1", "0","0",""]`). But the point still stands, we would be dropping information for any headers returned as lists so this needs to be updated. 

Seems to me that the most straightforward way is to to return something along the lines of ` String.join(",", headerValue);` We could also change the type to `Map<String, String>` as you suggest, but that would would put the work on the developer to convert multi valued headers to strings, and I'm leaning towards just doing it internally. If there are any other reasons why you think type should be `Map<String, String>` let me know and we can discuss.

src/main/java/com/microsoft/aad/msal4j/HttpUtils.java
```diff
@@ -0,0 +1,25 @@
...
+            return null;
+        }
+
+        List<String> headerValue = headers.get(headerName);
+
+        if(headerValue == null || headerValue.isEmpty()){
+            return null;
+        }
+
+        return String.join(",", headerValue);
```
henrik-me: >String.join(",", headerValue); [](start = 15, length = 30)

couldn't there be other separators?

Can you pls. add tests for this?

sangonzal: Not sure I understand your question. To clarify,  `HttpUtils` class is only used when we need to log certain headers returned from HTTP responses for logs or telemetry.  Since headers are returned as a list, what we are doing here is in certain cases where the list has multiple values (only case that I've seen from testing is the "Set-Cookie" header, which we don't even log anywhere), then we join those multiple values into one string, delimited by a comma. We could use any delimiter we want, but comma seemed like a sensible choice.

PR #135, opened 2019-11-26
Add server telemetry with v2 schema

sangonzal:


src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
```diff
@@ -34,7 +34,7 @@ static MsalServiceException fromHttpResponse(HTTPResponse httpResponse){
                 errorResponse.error().equalsIgnoreCase(AuthenticationErrorCode.INVALID_GRANT)) {
 
             if(isInteractionRequired(errorResponse.subError)){
-                throw new MsalInteractionRequiredException(errorResponse, httpResponse.getHeaderMap());
+                return new MsalInteractionRequiredException(errorResponse, httpResponse.getHeaderMap());
```
henrik-me: >return [](start = 16, length = 6)

? - is the exception thrown somewhere else?

src/main/java/com/microsoft/aad/msal4j/ServerSideTelemetry.java
```diff
@@ -0,0 +1,116 @@
...
+
+            middleSegmentBuilder.append(apiId).append(SCHEMA_COMMA_DELIMITER).append(correlationId);
+            errorSegmentBuilder.append(error);
+
+            if(it.hasNext()){
+                middleSegmentBuilder.append(SCHEMA_COMMA_DELIMITER);
+                errorSegmentBuilder.append(SCHEMA_COMMA_DELIMITER);
+            }
+
+            it.remove();
```
henrik-me: Seems like you are deleting the requests one at a time, would it make sense to only clear after a the request is successfully sent? 
also what happens if there are 2 requests going on a the same time? Will the first one here get all the elements and the next request will be waiting until this is complete?

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
sangonzal: Yes. MsalServiceExceptionFactory should only create exceptions. This was a bug.

src/main/java/com/microsoft/aad/msal4j/ServerSideTelemetry.java
sangonzal: Good point. I have updated to keep track of requests in progress. In case of HTTP status code 429 or >500, requests are not deleted. 

If there are 2 requests going on at the same time, for the same client application object, the first one will get all of the elements (up to 4kb of data), and flush them. The second thread will be blocked. (only while the request header is built). If there is any data left, it will flush that. If not, then it would send a blank previous request header.

PR #136, opened 2019-11-27
Add appName and appVersion headers.

sangonzal:
Addresses #134


src/main/java/com/microsoft/aad/msal4j/ClientApplicationBase.java
```diff
@@ -337,7 +347,9 @@ public T executorService(ExecutorService val) {
 
         /**
          * Sets Proxy configuration to be used by the client application for all network communication.
-         * Default is null and system defined properties if any, would be used.
+         * Default is null and system defined properties if any, would be used. If HTTP client is set on
```
SomkaPe: I would add that it does set proxy on default http client and if custom http client is set on ....

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
```diff
@@ -0,0 +1,60 @@
...
+                .applicationVersion("app-version")
+                .build();
+
+        RequestContext requestContext = new RequestContext(app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE);
+
+        HttpHeaders httpHeaders = new HttpHeaders(requestContext);
+
+        Map<String, String> httpHeaderMap = httpHeaders.getReadonlyHeaderMap();
+
+        Assert.assertEquals(httpHeaderMap.get("x-client-SKU"), "MSAL.Java");
```
SomkaPe: probably better use constants from HttpHeaders class

sangonzal: Fixed

src/main/java/com/microsoft/aad/msal4j/ClientApplicationBase.java
sangonzal: I have tried to clarify further. Let me know if you think it can be further improved.

PR #138, opened 2019-12-05
adfs2019 + new lab api

SomkaPe:
- ADFS2019 support 
- integration test switched to new Lab api

few issues found with new lab api:
1. "localhost" redirect URL is not configured for B2C applications, so corresponding tests (B2C auth code flow) are disabled

> 2. upn of guest user returned from lab contains character '#' , so user discovery request 
>     fails(because http GET is used),
>     I have tried to "url encode" upn, but "userrealm" endpoint fails to identify a user
it was fixed - homeUpn should be used for guest users.

Going to discuss mentioned issued with lab team


src/integrationtest/java/infrastructure/TcpListener.java
```diff
@@ -77,7 +77,9 @@ public void run(){
     }
 
     public ServerSocket createSocket() throws IOException {
-        int[] ports = { 3843,4584, 4843, 60000 };
+        //int[] ports = { 3843,4584, 4843, 60000 };
```
sangonzal: What happens if port 8080 is not available. If I remember correctly, this is a possibility and therefore that's why we have a list of ports.

src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
```diff
@@ -63,7 +57,5 @@ public void acquireTokenWithOBO_Managed() throws Exception {
         Assert.assertNotNull(result);
         Assert.assertNotNull(result.accessToken());
         Assert.assertNotNull(result.idToken());
-        // TODO AuthenticationResult should have an getAccountInfo API
-        // Assert.assertEquals(labResponse.getUser().getUpn(), result.getAccountInfo().getUsername());
```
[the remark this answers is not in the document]
SomkaPe: id token is not returned for OBO call, in fact only AT is getting returned. It actually depends on app config.

src/integrationtest/java/com.microsoft.aad.msal4j/AzureEnvironmentIT.java
```diff
@@ -22,41 +20,40 @@ public void setUp() {
...
+    private void assertAcquireTokenCommon(String azureEnvironment) throws Exception{
+        User user = labUserProvider.getUserByAzureEnvironment(azureEnvironment);
+
+        App app = LabService.getApp(user.getAppId());
 
         PublicClientApplication pca = PublicClientApplication.builder(
-                labResponse.getAppId()).
-                authority(TestConstants.ORGANIZATIONS_AUTHORITY).
+                user.getAppId()).
+                authority(app.getAuthority() + "organizations/").
```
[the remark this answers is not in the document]
SomkaPe: here we building environment specific "organizaions" authority,  but TestConstants.ORGANIZATIONS_AUTHORITY is for us cloud.

src/integrationtest/java/infrastructure/TcpListener.java
SomkaPe: adfs on prem send "unregistered redirect url" fo such urls.
so aad accept redirect url with any port but not adfs.

sangonzal: This is also the case for B2C. In that case we should just ask the lab to register redirect URLs with these ports (as they had done previously for B2C). Less of a chance of tests failing due to port being unavailable (which was a problem when I tried just using one port for all tests)

SomkaPe: we can just wait if port is busy, and try few times.

src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
```diff
@@ -24,4 +28,8 @@
 
     public final static String LOCALHOST = "http://localhost:";
     public final static String LOCAL_FLAG_ENV_VAR = "MSAL_JAVA_RUN_LOCAL";
+
+    public final static String ADFS_AUTHORITY = "https://fs.msidlab8.com/adfs/";
+    public final static String ADFS_SCOPE = "user.read";
```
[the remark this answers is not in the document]
SomkaPe: changed

src/integrationtest/java/infrastructure/TcpListener.java
sangonzal: My preference would be to use multiple ports, as that was seems more resilient and has been working perfectly. If you think doing retries is preferable, that's fine with me as well. In that case, you'd need to add the retry logic, as right now if the port is not available it fails and moves on.

SomkaPe: I have added retry logic, will ask lab teams to add few more ports, so we can do both

SomkaPe:
> Looks great. Much cleaner.
> 
> A couple of notes:
> 
> 1. The tests should also work on Azure DevOps. Can you make sure that everything works there?
> 2. There are a couple of unit tests that (I believe they are under AuthorityTests, but there might be others) that had been commented out until we added support for ADFS. Consider uncommenting those and making sure that they pass.


good catch about some disabled adfs unit test 👍 
test on DevOps are green, there was minor issue with chromium driver version

PR #149, opened 2019-12-13
Add instance discovery metadata to token cache

sangonzal:
In the case where a client application object is create in a new process, and they are using a token cache that has been previously written to disk, then they can use the instance discovery metadata from the token cache instead of having to make a network call.


src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscovery.java
```diff
@@ -85,41 +142,28 @@ private static void validate(InstanceDiscoveryResponse instanceDiscoveryResponse
...
-    }
-
-    private static void cacheInstanceDiscoveryMetadata(String host, InstanceDiscoveryResponse instanceDiscoveryResponse) {
+    private static void cacheInstanceDiscoveryMetadata(
+            String host,
+            InstanceDiscoveryResponse instanceDiscoveryResponse) {
         if (instanceDiscoveryResponse.metadata() != null) {
             for (InstanceDiscoveryMetadataEntry entry : instanceDiscoveryResponse.metadata()) {
+
+                // expiry time of 96 hours
```
sangonzal: What should expiry date be?

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscovery.java
```diff
@@ -46,6 +46,60 @@
...
+            ServiceBundle serviceBundle) {
+
+        log.info("Instance Discovery information not found in cache. Sending instance discovery " +
+                "request to AAD");
+
+        InstanceDiscoveryResponse instanceDiscoveryResponse =
+                sendInstanceDiscoveryRequest(authorityUrl, msalRequest, serviceBundle);
+
+        if (validateAuthority) {
+            validate(instanceDiscoveryResponse);
```
bgavrilMS: [Aside] The way you do `validate` is different from how .net does it so it might be a good idea to align here... Can you give me an example where the instance discovery endpoint would fail validation ? I manually tested using a "bogus authority" and the instance discovery endpoint fails with a 500 error I believe.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -20,7 +21,7 @@
 
     protected static final int MIN_ACCESS_TOKEN_EXPIRE_IN_SEC = 5 * 60;
 
-    transient private ReadWriteLock lock = new ReentrantReadWriteLock();
+    transient ReadWriteLock lock = new ReentrantReadWriteLock();
```
bgavrilMS: Expose a getter instead?

src/test/java/com/microsoft/aad/msal4j/InstanceDiscoveryTest.java
```diff
@@ -0,0 +1,155 @@
...
+                "code", new URI("http://my.redirect.com")).build();
+
+        MsalRequest msalRequest = new AuthorizationCodeRequest(
+                parameters,
+                app,
+                new RequestContext(app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE));
+
+        URL authority = new URL(app.authority());
+
+        PowerMock.mockStaticPartial(AadInstanceDiscovery.class, "sendInstanceDiscoveryRequest");
```
bgavrilMS: [Aside] If you use an interface and either make your service a singleton or just pass the instance around, the code would be easier to test without tools for "untestable code"

src/test/resources/Instance_discovery_metadata_cache_data/cache_instance_discovery_metadata_expired.json
```diff
@@ -0,0 +1,122 @@
...
+    "field2": "whats"
+  },
+  "AppMetadata": {
+    "appmetadata-login.windows.net-my_client_id": {
+      "environment": "login.windows.net",
+      "family_id": null,
+      "client_id": "my_client_id"
+    }
+  },
+  "InstanceDiscoveryMetadata": {
```
bgavrilMS: MSAL .net should be ok with this, it is similar to the "unknown entity" above. Otherwise, the cache structure looks good to me, but please open an API review so that this is documented for other teams / new MSALs to come.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscovery.java
sangonzal: We are not validating instance discovery endpoint here, only validating that the response contains "tenant_discovery_endpoint". Authority validation happens here: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/main/java/com/microsoft/aad/msal4j/Authority.java#L88

What validation happens in .net? Agreed that we should align here.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
sangonzal: Good point, will fix.

src/test/resources/Instance_discovery_metadata_cache_data/cache_instance_discovery_metadata_expired.json
henrik-me: +1 : to get this documented

Issue #12, opened 2019-03-18
Gson is used as JSON parser


JonathanGiles:
The azure-sdk-for-java project is trying to standardise on using Jackson as its JSON library of choice. Is it possible to discuss the possibility of this library also standardising on this library, or is Gson a set-in-stone decision now?

sangonzal:
Gson is not set-in-stone, happy to discuss what it would take to move over to Jackson.

SomkaPe:
Lets discuss in teams or by email the reasons for this standardisation.

henrik-me:
Re-opening per discussion with Jonathan

PR #155, opened 2019-12-20
switching to Jackson Json parser

SomkaPe:
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/12

### DIFF
--- a/lombok.config
+++ b/lombok.config
@@ -1,0 +1,1 @@
+lombok.anyConstructor.addConstructorProperties=true
--- a/src/integrationtest/java/labapi/LabService.java
+++ b/src/integrationtest/java/labapi/LabService.java
@@ -3,7 +3,9 @@
 
 package labapi;
 
-import com.google.gson.Gson;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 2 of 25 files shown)


src/integrationtest/java/labapi/LabService.java
```diff
@@ -16,6 +18,17 @@
 
     static ConfidentialClientApplication labApp;
 
+    static ObjectMapper mapper = new ObjectMapper()
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+
+    static <T> T convertJsonToObject(final String json, final Class<T> clazz) {
```
sangonzal: Are you able to use JsonHelper here? Would avoid having duplicate code

lombok.config
```diff
@@ -0,0 +1 @@
+lombok.anyConstructor.addConstructorProperties=true
```
sangonzal: I'd prefer to have the annotations and constructor than an extra config file.

src/integrationtest/java/labapi/LabService.java
SomkaPe: will have to make it public

lombok.config
SomkaPe: adding those constructors might be very error prone because it should depend on json properties - so when you add new or remove json property you also need update constructor and keep in in sync

PR #156, opened 2019-12-20
Add option to pass in instance discovery response

sangonzal:
After further discussion, we decided not to merge #149 due to: 
- Storing instance discovery metadata in token cache increases the number of times that an application needs to access token cache. This increases the possibility of lock contention. 
- Adds complexity to token cache format, which multiple libraries need to agree on.
- Same functional requirements can be met with simpler approach

This PR adds the option for developer to pass in instance discovery metadata, and meets the same functional requirements.  Accompanying wiki post at https://aka.ms/msal4j-instance-discovery


sangonzal:
> in description you mentioned that metadata can be requested from https://{host}/common/discovery/instance and provided to app, but as i understand form previous conversation this endpoint is not available in region specific configs.

@SomkaPe This is a good point. I will update the wiki page with instructions for region specific configuration.

src/main/java/com/microsoft/aad/msal4j/ClientApplicationBase.java
```diff
@@ -77,6 +75,10 @@
     @Getter
     private String applicationVersion;
 
+    @Accessors(fluent = true)
+    @Getter
+    private InstanceDiscoveryResponse instanceDiscoveryResponse;
```
SomkaPe: I would rename to AADInstanceDiscoveryResponse because it is only relevant for AAD authorities, 
so not for B2C or ADFS

SomkaPe:
If Instance Discovery Response set manually, during construction of application object we might want to check "validate authority" flag, and throw Exception if it is true.

"Authority validation" feature is essentially check that authority is known to AAD (we do it by issuing instance discovery request to hardcoded AAD endpoint).
So, if we use provided by developer instance discovery response, we do no validation at all.

Also we should discuss option of throwing exception if application object is configured with ADFS or B2C authority and "validate authority" flag is set as true.
@henrik-me, @bgavrilMS, @navyasric, @rayluo   it would be nice to hear yours opinions as well.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -46,6 +54,41 @@
...
+        InstanceDiscoveryResponse instanceDiscoveryResponse;
+        try {
+            instanceDiscoveryResponse = JsonHelper.convertJsonToObject(
+                    instanceDiscoveryJson,
+                    InstanceDiscoveryResponse.class);
+
+            validate(instanceDiscoveryResponse);
+
+        } catch(Exception ex){
+            throw new MsalClientException("Error parsing instance discovery response. Data must be" +
```
bgavrilMS: needs more spaces...

bgavrilMS:
@SomkaPe - `ValidateAuthority` is a good intention but a bad API. Users don't understand it and they shouldn't be asked to understand. AFAIK MSAL.Android even removed this API and we should strive to do the same, but bear in mind this is a breaking change and since we're all GA now, breaking changes have a much higher bar. 

In the end, what we want to protect against is the user calling an hacked IdP. But in 99% of the cases, the authority is hardcoded, I have yet to see applications that call an API, get a 401, extract the authority blindly and start talking to it.

As such, in my opinion, we should not throw if the user doesn't set ValidateAuthority properly, i.e. for B2C `ValidateAuthority` does not make sense so just ignore the param. For ADFS, `ValidateAuthority` makes sense, there is a [specific API](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/87ae86eeb561c2393ef2f2a57cd8f5e773417b90/src/client/Microsoft.Identity.Client/Instance/AdfsOpenIdConfigurationEndpointManager.cs#L28) for it. 


So the general algorithm is: 

- try the user metadata first (if it exists). If the authority is not found in the user json, throw (bad config). 
- try the instance metadata static cache second, do not throw
- call the instance discovery endpoint. If it fails with `invalid_instance` throw. If it fails with anything else (especially timeouts), DO NOT throw (this is to increase availability of AAD). It's true that in this case we might not understand aliases, but this is much more preferable to crashing people's apps.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
sangonzal: What needs more spaces? The error message?

bgavrilMS: Yes, "Data must be_..."

sangonzal:
I agree that we should get rid of `ValidateAuthority` whenever we next introduce breaking changes.

@SomkaPe Instead of throwing exception, the other would be to clearly document in the `AadInstanceDiscovery`, `ValidateAuthority`, and `B2cAuthority` API JavaDoc that validate authority will only be validated if it's an AAD/ADFS authority and instance discovery metadata is not passed in.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
sangonzal: I see, will update

src/main/java/com/microsoft/aad/msal4j/ClientApplicationBase.java
sangonzal: Ok will update

SomkaPe:
> I agree that we should get rid of `ValidateAuthority` whenever we next introduce breaking changes.
> 
> @SomkaPe Instead of throwing exception, the other would be to clearly document in the `AadInstanceDiscovery`, `ValidateAuthority`, and `B2cAuthority` API JavaDoc that validate authority will only be validated if it's an AAD/ADFS authority and instance discovery metadata is not passed in.

Java Doc might be good enough,  ... only be validated if it's an AAD/ADFS authority  - may be if it is AAD authority

SomkaPe:
:shipit:

Issue #140, opened 2019-12-07
B2C does not always issue access token but TokenResponse assumes there is an access token resulting in a JSON parse error


alex-mason-by:
Using MSAL4J library with B2C can result in a JSON parse error due to an assumption in TokenResponse that doesn't always hold with B2C and that assumption is that an access token is always returned. The refresh token has the same assumption although perhaps that is fine? In the case where MSAL4J is used for a server-side application login that needs no custom scope then no access token is returned and this code should not blow up on that scenario.

https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/main/java/com/microsoft/aad/msal4j/TokenResponse.java#L48
        final AccessToken accessToken = AccessToken.parse(jsonObject);
        final RefreshToken refreshToken = RefreshToken.parse(jsonObject);

sangonzal:
@alex-mason-jdas Acknowledged, I am able to reproduce this error. Discussing possible solutions with the team, will update this issue once we have a ETA for a fix.

sangonzal:
@alex-mason-jdas Can you please share details about the scenario where you are trying to acquire just an Id token and no access token or refresh token? Is there any reason why are not requesting the access/ refresh tokens?

alex-mason-jdas:
That is the default behavior of B2C if there is no custom scope requested, so it would be a scenario that is just a sign in scenario (monolithic app perhaps). You can see indication in the provided link. I myself don't really agree with the implementation of B2C and think it ought to always return an access token, but it does not if there is no custom scope meaning that this application does not need to call some other web API resource. As far as refresh token, well that really depends on whether the nature of the application needs a longer life or not, doesn't it? In any case,  I think you really need to discuss with the B2C team. I am just reporting the behavior since MSAL4J claims to support B2C and this is kind of a basic error. Perhaps the resolution is on the B2C side but it's one of the two and should not really be put back on the user of MSAL nor B2C.

https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oidc

sangonzal:
@alex-mason-jdas Thanks for expanding. 

I've had a chance to dig in a bit deeper and chat with the B2C team. They should be returning access tokens on all successful token responses. This is a bug on their end which they are now aware of. 

[...]

PR #158, opened 2020-01-03
Remove assumption that token response contains AT or RT

sangonzal:
Addresses #140 

In certain scenarios, developers can chose not to request access tokens or refresh tokens from B2C. In those cases, the response will contain an Id_token, but no access or refresh tokens. MSAL4J should not assume that a successful response to the /token endpoint will have those values.

Not ready to merge in, as there are several open questions around how the token cache should behave in relation to AuthenticationResults with no access/refresh tokens. Opening PR to address initial feedback


src/main/java/com/microsoft/aad/msal4j/TokenResponse.java
```diff
@@ -5,27 +5,29 @@
...
     private String foci;
 
-    TokenResponse(final AccessToken accessToken, final RefreshToken refreshToken, final String idToken,
+    TokenResponse(final String accessToken, final String refreshToken, final String idToken,
                   final String scope, String clientInfo, long expiresIn, long extExpiresIn, String foci) {
-        super(new OIDCTokens(idToken, accessToken, refreshToken));
+
+        this.accessToken = accessToken;
+        this.refreshToken = refreshToken;
+        this.idToken = idToken; //TODO: Do we assume that IdToken won't be null or empty
```
henrik-me: >TODO: Do we assume that IdToken won't be null or empty [](start = 34, length = 54)

I believe the cache currently depends on it. Not sure what ADFS and B2C does for all scenarios though?

SomkaPe: i do not think that we assume that, for client credential grant STS does not return idToken

src/test/java/com/microsoft/aad/msal4j/TokenRequestExecutorTest.java
```diff
@@ -27,10 +27,50 @@
...
+        EasyMock.expect(msalOAuthHttpRequest.send()).andReturn(httpResponse).times(1);
+        PowerMock.replay(request, msalOAuthHttpRequest);
+
+        AuthenticationResult result = request.executeTokenRequest();
+
+        Assert.assertNotNull(result.idToken());
+        Assert.assertNotNull(result.account());
+        Assert.assertNull(result.accessToken());
+        Assert.assertNull(result.refreshToken());
+        // Assert.assertEquals(); //TODO: what should be the expiry date in AuthenticationResult when response does not contain expires_on?
```
SomkaPe: good question.
B2C response contains "not_before" claims which is not part of Oauth2.0 or OIDC specs, but it does not contains 
expires_in (probably because it is used for access_token).

We use expires_in from token response to calculate  expires_on for AuthenticationResult. 
Which is targeting access token expiration time, not id token.

We need to clarify which tokens are targeted by expiration time related properties in token response

PR #166, opened 2020-01-18
adding sameSite cookie attr support to samples

SomkaPe:
samples updated to support "SameSite" attribute of session cookies:
- setting "secure" and "SameSite=None" attributes  for session cookies
- using https with local host


src/samples/msal-web-sample/src/main/resources/application.properties
```diff
@@ -1,5 +1,15 @@
 aad.authority=https://login.microsoftonline.com/common/
 aad.clientId=
 aad.secretKey=
-aad.redirectUri=http://localhost:8080/msal4jsample/secure/aad
-aad.oboApi=<OboApi>/access_as_user
\ No newline at end of file
+aad.redirectUri=https://localhost:8081/msal4jsample/secure/aad
+aad.oboApi=<OboAi>/access_as_user
+
+server.servlet.session.cookie.secure=true
```
sangonzal: Is secure cookie necessary as part of the SameSite changes? 

I realize that this is a necessary practice in a production app, but it had been omitted in this samples previously because it further complicates getting the sample running. For the Azure Samples and the portal quickstart, these changes will have to be explained, including instructions on how to set this up. IMO a section in the README and in a comment in the code could be sufficient.

navyasric:
> These changes should also be made in AzureSamples samples and to the portal quickstart. Also a wiki page explaining these changes which all the samples can point to would be useful.

Adding to Santiago's comment: Please prioritize adding these changes to the Azure Samples here:
https://github.com/Azure-Samples/ms-identity-java-webapp  
https://github.com/Azure-Samples/ms-identity-java-webapi 
Java B2C sample: https://github.com/AzureAD/microsoft-authentication-library-for-java/tree/dev/src/samples/msal-b2c-web-sample 

These are the customer facing samples we promote through docs etc.

src/samples/msal-web-sample/src/main/resources/application.properties
SomkaPe: In new versions of Chrome cookies with same site none will not be included in cs request if they not secure

SomkaPe:
@navyasric Corresponding PR were submitted, please review

src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -99,6 +99,7 @@ private HttpResponse readResponseFromConnection(final HttpsURLConnection conn) t
             if (responseCode != HttpURLConnection.HTTP_OK) {
                 is = conn.getErrorStream();
                 if (is != null) {
+                    httpResponse.headers(conn.getHeaderFields());
```
henrik-me: >ields()); [](start = 56, length = 9)

bug? Seems like there is a test missing?

src/samples/msal-web-sample/src/main/java/com/microsoft/azure/msalwebsample/CookieHelper.java
```diff
@@ -0,0 +1,95 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
henrik-me: why is this not shared code across samples?

SomkaPe: technically all samples are different projects,

src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
SomkaPe: yes, bug - headers were not returned in case of error, missed test class during commit , will add

PR #190, opened 2020-02-29
support for national cloud integ tesing + arlington tests

SomkaPe:
- B2C currently not tested in Arlington because it is not supported yet
- instance aware not supported 

All others flows looks good


src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
```diff
@@ -56,8 +60,8 @@ public void DeviceCodeFlowTest() throws Exception {
     }
 
     private void runAutomatedDeviceCodeFlow(DeviceCode deviceCode, User user){
-        boolean isRunningLocally = !Strings.isNullOrEmpty(
-                System.getenv(TestConstants.LOCAL_FLAG_ENV_VAR));
+        boolean isRunningLocally = true;//!Strings.isNullOrEmpty(
```
sangonzal: Please revert this - otherwise the AzureDevops tests will fail. You should be able to just set an environment variable and not have to comment this out.

henrik-me: perhaps the test should say this if this fails... please set this env variable if you are running these tests on your local box?

SomkaPe: ok, going to revert it, but it might be not required any more - tests are green

PR #201, opened 2020-03-28
support of STS throttling instructions + caching of InteractionRequiredException…

SomkaPe:
- support of STS throttling instructions - 429 Retry-After header + 429, 5xx response status codes
- caching of InteractionRequiredException responses for refresh token grant requests
- refactoring

issue https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/176

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
@@ -34,6 +34,9 @@ final class HttpHeaders {
     private  final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_NAME = "return-client-request-id";
     private final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_VALUE = "true";
 
+    private  final static String X_MS_LIB_CAPABILITY_NAME = "x-ms-lib-capability";
+    private final static String X_MS_LIB_CAPABILITY_VALUE = "retry-after, h429";
+
     private final String headerValues;
     private final Map<String, String> headerMap = new HashMap<>();
 
@@ -66,6 +69,8 @@ final class HttpHeaders {
         if(!StringHelper.isBlank(this.applicationVersionHeaderValue)){
             init.accept(APPLICATION_VERSION_HEADER_NAME, this.applicationVersionHeaderValue);
         }
+
+        init.accept(X_MS_LIB_CAPABILITY_NAME, this.X_MS_LIB_CAPABILITY_VALUE);
 
         return sb.toString();
     }
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
@@ -6,32 +6,55 @@ package com.microsoft.aad.msal4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 class HttpHelper {
 
     private final static Logger log = LoggerFactory.getLogger(HttpHelper.class);
+    public static final String RETRY_AFTER_HEADER = "Retry-After";
 
-    static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
-                                            final RequestContext requestContext,
-                                            final ServiceBundle serviceBundle) {
+    private static String getRequestThumbprint(RequestContext requestContext) {
+        String DELIMITER = ".";
+
+        StringBuilder sb = new StringBuilder();
+        sb.append(requestContext.clientId() + DELIMITER);
+        sb.append(requestContext.authority() + DELIMITER);
+
+        IApiParameters apiParameters = requestContext.apiParameters();
+
+        if (apiParameters instanceof SilentParameters) {
+            IAccount account = ((SilentParameters) apiParameters).account();
+            if (account != null) {
+                sb.append(account.homeAccountId() + DELIMITER);
+            }
+        }
+
+        Set<String> sortedScopes = new TreeSet<>(apiParameters.scopes());
+        sb.append(String.join(" ", sortedScopes));
+
+        return StringHelper.createSha256Hash(sb.toString());
+    }
+
+    static IHttpResponse executeHttpRequest(HttpRequest httpRequest,
+                                            RequestContext requestContext,
+                                            ServiceBundle serviceBundle) {
+        checkForThrottling(requestContext);
 
         HttpEvent httpEvent = new HttpEvent(); // for tracking http telemetry
         IHttpResponse httpResponse;
 
-        try(TelemetryHelper telemetryHelper = serviceBundle.getTelemetryManager().createTelemetryHelper(
+        try (TelemetryHelper telemetryHelper = serviceBundle.getTelemetryManager().createTelemetryHelper(
                 requestContext.telemetryRequestId(),
                 requestContext.clientId(),
                 httpEvent,
-                false)){
+                false)) {
 
             addRequestInfoToTelemetry(httpRequest, httpEvent);
 
-            try{
+            try {
                 IHttpClient httpClient = serviceBundle.getHttpClient();
                 httpResponse = httpClient.send(httpRequest);
-            } catch(Exception e){
+            } catch (Exception e) {
                 httpEvent.setOauthErrorCode(AuthenticationErrorCode.UNKNOWN);
                 throw new MsalClientException(e);
             }
@@ -42,17 +65,66 @@ class HttpHelper {
                 HttpHelper.verifyReturnedCorrelationId(httpRequest, httpResponse);
             }
         }
+        processThrottlingInstructions(httpResponse, requestContext);
+
         return httpResponse;
     }
 
-    private static void addRequestInfoToTelemetry(final HttpRequest httpRequest, HttpEvent httpEvent){
-        try{
+    private static void checkForThrottling(RequestContext requestContext) {
+        if (requestContext.clientApplication() instanceof PublicClientApplication &&
+                requestContext.apiParameters() != null) {
+            String requestThumbprint = getRequestThumbprint(requestContext);
+
+            long retryInMs = ThrottlingCache.retryInMs(requestThumbprint);
+
+            if (retryInMs > 0) {
+                throw new MsalThrottlingException(retryInMs);
+            }
+        }
+    }
+
+    private static void processThrottlingInstructions(IHttpResponse httpResponse, RequestContext requestContext) {
+        if (requestContext.clientApplication() instanceof PublicClientApplication) {
+            Long expirationTimestamp = null;
+
+            Integer retryAfterHeaderVal = getRetryAfterHeader(httpResponse);
+            if (retryAfterHeaderVal != null) {
+                expirationTimestamp = System.currentTimeMillis() + retryAfterHeaderVal * 1000;
+            } else if (httpResponse.statusCode() == 429 ||
+                    (httpResponse.statusCode() >= 500 && httpResponse.statusCode() < 600)) {
+
+                expirationTimestamp = System.currentTimeMillis() + ThrottlingCache.DEFAULT_THROTTLING_TIME_SEC * 1000;
+            }
+            if (expirationTimestamp != null) {
+                ThrottlingCache.set(getRequestThumbprint(requestContext), expirationTimestamp);
+            }
+        }
+    }
+
+    private static Integer getRetryAfterHeader(IHttpResponse httpResponse) {
+
+        if (httpResponse.headers() != null) {
+            TreeMap<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            headers.putAll(httpResponse.headers());
+
+            if (headers.containsKey(RETRY_AFTER_HEADER) && headers.get(RETRY_AFTER_HEADER).size() == 1) {
+                int headerValue = Integer.parseInt(headers.get(RETRY_AFTER_HEADER).get(0));
+                if (headerValue > 0 && headerValue <= ThrottlingCache.MAX_THROTTLING_TIME_SEC) {
+                    return headerValue;
+                }
+            }
+        }
+        return null;
+    }
+
+    private static void addRequestInfoToTelemetry(final HttpRequest httpRequest, HttpEvent httpEvent) {
+        try {
             httpEvent.setHttpPath(httpRequest.url().toURI());
             httpEvent.setHttpMethod(httpRequest.httpMethod().toString());
-            if(!StringHelper.isBlank(httpRequest.url().getQuery())){
+            if (!StringHelper.isBlank(httpRequest.url().getQuery())) {
                 httpEvent.setQueryParameters(httpRequest.url().getQuery());
             }
-        } catch(Exception ex){
+        } catch (Exception ex) {
             String correlationId = httpRequest.headerValue(
                     HttpHeaders.CORRELATION_ID_HEADER_NAME);
 
@@ -62,28 +134,28 @@ class HttpHelper {
         }
     }
 
-    private static void addResponseInfoToTelemetry(IHttpResponse httpResponse, HttpEvent httpEvent){
+    private static void addResponseInfoToTelemetry(IHttpResponse httpResponse, HttpEvent httpEvent) {
 
         httpEvent.setHttpResponseStatus(httpResponse.statusCode());
 
         Map<String, List<String>> headers = httpResponse.headers();
 
         String userAgent = HttpUtils.headerValue(headers, "User-Agent");
-        if(!StringHelper.isBlank(userAgent)){
+        if (!StringHelper.isBlank(userAgent)) {
             httpEvent.setUserAgent(userAgent);
         }
 
         String xMsRequestId = HttpUtils.headerValue(headers, "x-ms-request-id");
-        if(!StringHelper.isBlank(xMsRequestId)){
+        if (!StringHelper.isBlank(xMsRequestId)) {
             httpEvent.setRequestIdHeader(xMsRequestId);
         }
 
-        String xMsClientTelemetry = HttpUtils.headerValue(headers,"x-ms-clitelem");
-        if(xMsClientTelemetry != null){
+        String xMsClientTelemetry = HttpUtils.headerValue(headers, "x-ms-clitelem");
+        if (xMsClientTelemetry != null) {
             XmsClientTelemetryInfo xmsClientTelemetryInfo =
                     XmsClientTelemetryInfo.parseXmsTelemetryInfo(xMsClientTelemetry);
 
-            if(xmsClientTelemetryInfo != null){
+            if (xmsClientTelemetryInfo != null) {
                 httpEvent.setXmsClientTelemetryInfo(xmsClientTelemetryInfo);
             }
         }
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
@@ -37,8 +37,9 @@ class HttpHelper {
         return StringHelper.createSha256Hash(sb.toString());
     }
 
-    static private boolean isRetriable(IHttpResponse httpResponse) {
-        return httpResponse.statusCode() >= 500 && httpResponse.statusCode() < 600;
+    static private boolean isRetryable(IHttpResponse httpResponse) {
+        return httpResponse.statusCode() >= 500 && httpResponse.statusCode() < 600 &&
+                getRetryAfterHeader(httpResponse) == null;
     }
 
     static IHttpResponse executeHttpRequest(HttpRequest httpRequest,
@@ -83,7 +84,7 @@ class HttpHelper {
         IHttpResponse httpResponse = null;
         for (int i = 0; i < RETRY_NUM; i++) {
             httpResponse = httpClient.send(httpRequest);
-            if (!isRetriable(httpResponse)) {
+            if (!isRetryable(httpResponse)) {
                 break;
             }
             Thread.sleep(RETRY_DELAY_MS);
@@ -129,9 +130,15 @@ class HttpHelper {
             headers.putAll(httpResponse.headers());
 
             if (headers.containsKey(RETRY_AFTER_HEADER) && headers.get(RETRY_AFTER_HEADER).size() == 1) {
-                int headerValue = Integer.parseInt(headers.get(RETRY_AFTER_HEADER).get(0));
-                if (headerValue > 0 && headerValue <= ThrottlingCache.MAX_THROTTLING_TIME_SEC) {
-                    return headerValue;
+                try {
+                    int headerValue = Integer.parseInt(headers.get(RETRY_AFTER_HEADER).get(0));
+
+                    if (headerValue > 0 && headerValue <= ThrottlingCache.MAX_THROTTLING_TIME_SEC) {
+                        return headerValue;
+                    }
+                }
+                catch (NumberFormatException ex){
+                    log.warn("Failed to parse value of Retry-After header - NumberFormatException");
                 }
             }
         }
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
@@ -8,24 +8,26 @@ import org.slf4j.LoggerFactory;
 
 import java.util.*;
 
+import static com.microsoft.aad.msal4j.Constants.POINT_DELIMITER;
+
 class HttpHelper {
 
     private final static Logger log = LoggerFactory.getLogger(HttpHelper.class);
     public static final String RETRY_AFTER_HEADER = "Retry-After";
+    public static final int RETRY_NUM = 2;
+    public static final int RETRY_DELAY_MS = 1000;
 
     private static String getRequestThumbprint(RequestContext requestContext) {
-        String DELIMITER = ".";
-
         StringBuilder sb = new StringBuilder();
-        sb.append(requestContext.clientId() + DELIMITER);
-        sb.append(requestContext.authority() + DELIMITER);
+        sb.append(requestContext.clientId() + POINT_DELIMITER);
+        sb.append(requestContext.authority() + POINT_DELIMITER);
 
         IApiParameters apiParameters = requestContext.apiParameters();
 
         if (apiParameters instanceof SilentParameters) {
             IAccount account = ((SilentParameters) apiParameters).account();
             if (account != null) {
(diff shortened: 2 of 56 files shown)


src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -42,17 +65,66 @@ static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
...
+
+    private static void processThrottlingInstructions(IHttpResponse httpResponse, RequestContext requestContext) {
+        if (requestContext.clientApplication() instanceof PublicClientApplication) {
+            Long expirationTimestamp = null;
+
+            Integer retryAfterHeaderVal = getRetryAfterHeader(httpResponse);
+            if (retryAfterHeaderVal != null) {
+                expirationTimestamp = System.currentTimeMillis() + retryAfterHeaderVal * 1000;
+            } else if (httpResponse.statusCode() == 429 ||
+                    (httpResponse.statusCode() >= 500 && httpResponse.statusCode() < 600)) {
```
sangonzal, line 94: We want to throttle every time we get any 500 status code back?

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
```diff
@@ -30,25 +30,26 @@ static MsalServiceException fromHttpResponse(HTTPResponse httpResponse){
         errorResponse.statusCode(httpResponse.getStatusCode());
         errorResponse.statusMessage(httpResponse.getStatusMessage());
 
-        if(errorResponse.error() != null &&
+        if (isRefreshTokenGrantRequest &&
```
sangonzal, line 33: I don't know if it makes sense to just limit throwing InteractionRequiredError for just RefreshTokenGrant. What about in the case of username password and integrated windows authentication? interaction required is just as relevant in those cases as it is for refresh flow.

src/main/java/com/microsoft/aad/msal4j/RefreshTokenRequest.java
```diff
@@ -6,18 +6,54 @@
...
     private static AbstractMsalAuthorizationGrant createAuthenticationGrant(
-            RefreshTokenParameters parameters){
+            RefreshTokenParameters parameters) {
 
         RefreshTokenGrant refreshTokenGrant = new RefreshTokenGrant(new RefreshToken(parameters.refreshToken()));
         return new OAuthAuthorizationGrant(refreshTokenGrant, parameters.scopes());
     }
+
+    private String getRefreshTokenRequestFullThumbprint() {
+        String DELIMITER = ".";
```
sangonzal, line 40: This delimeter is used in multiple places. Might make sense to add to Contants file.

src/main/java/com/microsoft/aad/msal4j/SilentRequest.java
```diff
@@ -31,4 +33,20 @@
         application.getServiceBundle().getServerSideTelemetry().getCurrentRequest().forceRefresh(
                 parameters.forceRefresh());
     }
+
+    String getFullThumbprint(){
```
sangonzal, line 37: This is repeated - maybe could be put in MsalRequest and reused in RefreshTokenRequest.

src/main/java/com/microsoft/aad/msal4j/StringHelper.java
```diff
@@ -27,4 +27,19 @@ static String createBase64EncodedSha256Hash(String stringToHash){
         }
         return base64EncodedSha256Hash;
     }
+
+    static String createSha256Hash(String stringToHash){
```
sangonzal, line 31: Can `createBase64EncodedSha256Hash` be refactored to use this new method?

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: according to design doc yes, but only requests with same thumbprint.
I see your concern, generally speaking 5xx errors are retriable. 
It would be better to add some retry logic around this, and throttle only if all retries failed

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
SomkaPe: both cases are "interactive", if it fails It means that provided credentials are wrong, there is no  plan B - plan B exists only for silent flow

src/main/java/com/microsoft/aad/msal4j/SilentRequest.java
SomkaPe: Thumbprints of silent and refresh tokens requests are different

src/main/java/com/microsoft/aad/msal4j/RefreshTokenRequest.java
SomkaPe: ok

src/main/java/com/microsoft/aad/msal4j/StringHelper.java
SomkaPe: ok

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
sangonzal: Not sure I would consider IWA and U/P interactive. To me it seems that an interactive request is one where the IDP renders UI, which is not the case for either of those. 

Also IWA and U/P can fail for more than just providing wrong credentials - one example is MFA. In this case, someone doing U/P could theoretically fall back to doing an interactive flow to meet the MFA requirement. I agree that it probably does not make sense to do this in an application, but anyways the InteractionRequired error is still 100% valid and the most transparent error to expose to an user. In the case that we made the change you are recommending, and someone is using one of this flows with MFA enabled on the tenant, what kind of error would they see?

SomkaPe: interactive vs non-interactive/silent distinguishing does make sense only for flows where user perform authorization "interactively" - like auth code flow - user redirected to identity provider authorization endpoint and he interact with it - so identity provider can give him additional challenges and he can satisfy it. In case of U/P or WIA - if credentials are wrong or missing there is not much we can do, we do not have plan B (different api) to handle it, like we do with auth code flow

henrik-me:
@bgavrilMS @rayluo : please add comments and discuss offline as needed.

src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
```diff
@@ -34,6 +34,9 @@
     private  final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_NAME = "return-client-request-id";
     private final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_VALUE = "true";
 
+    private  final static String X_MS_LIB_CAPABILITY_NAME = "x-ms-lib-capability";
```
bgavrilMS, line 37: Consider adding a comment on what these 2 capabilities are.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -6,32 +6,63 @@
...
+import static com.microsoft.aad.msal4j.Constants.POINT_DELIMITER;
 
 class HttpHelper {
 
     private final static Logger log = LoggerFactory.getLogger(HttpHelper.class);
+    public static final String RETRY_AFTER_HEADER = "Retry-After";
+    public static final int RETRY_NUM = 2;
+    public static final int RETRY_DELAY_MS = 1000;
+
+    private static String getRequestThumbprint(RequestContext requestContext) {
```
bgavrilMS, line 20: Is this method supposed to work on other request types than SilentRequest? I think the thumbprints for non-silent requests are not valid, as they contain too little data. Consider renaming to `GetSilentRequestThumbprint`..

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -6,32 +6,63 @@
...
+        Set<String> sortedScopes = new TreeSet<>(apiParameters.scopes());
+        sb.append(String.join(" ", sortedScopes));
 
-    static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
-                                            final RequestContext requestContext,
-                                            final ServiceBundle serviceBundle) {
+        return StringHelper.createSha256Hash(sb.toString());
+    }
+
+    static private boolean isRetriable(IHttpResponse httpResponse) {
```
bgavrilMS, line 40: nit: typo retryable

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -42,17 +73,79 @@ static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
...
-        try{
+    private static void checkForThrottling(RequestContext requestContext) {
+        if (requestContext.clientApplication() instanceof PublicClientApplication &&
+                requestContext.apiParameters() != null) {
+            String requestThumbprint = getRequestThumbprint(requestContext);
+
+            long retryInMs = ThrottlingCache.retryInMs(requestThumbprint);
+
+            if (retryInMs > 0) {
+                throw new MsalThrottlingException(retryInMs);
```
bgavrilMS, line 102: I was under the impression that throttling concept is not exposed to client. We should just throw the old exception, as if the client had made the HTTP call and the server returned the same error. 

I think this makes sense because it would behave well with retry policies that customers have added on their own.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -42,17 +73,79 @@ static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
...
+
+    private static Integer getRetryAfterHeader(IHttpResponse httpResponse) {
+
+        if (httpResponse.headers() != null) {
+            TreeMap<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            headers.putAll(httpResponse.headers());
+
+            if (headers.containsKey(RETRY_AFTER_HEADER) && headers.get(RETRY_AFTER_HEADER).size() == 1) {
+                int headerValue = Integer.parseInt(headers.get(RETRY_AFTER_HEADER).get(0));
+                if (headerValue > 0 && headerValue <= ThrottlingCache.MAX_THROTTLING_TIME_SEC) {
```
bgavrilMS, line 133: if headerValue > MAX or cannot be parsed, perhaps we should behave as if no "RetryAfter" header was set, i.e. use DEFAULT_THROTTLING_TIME_SEC instead of NULL, at least if response is 429 or 5xx.

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
bgavrilMS: This is a great point @sangonzal,
I have seen many client applications implementing flow as:

try
  - AcquireTokenSilent
catch
   - IWA
catch 
    - Interactive

In fact, the spec states

`... for silent flow some errors of WS-Trust flow should go in UI required cache as well as errors from token endpoint. Applications have tendency to loop on WS-Trust flow, which would result in DDoS of UserRealm endpoint. There is no way such cases for the application to proceed without showing UI. Here is the list of these errors:

account is managed/unknown,
integration is not WS-Trust (SAML2.0)
missing MEX endpoint,
invalid MEX document,
MEX doesn’t have Windows Transport Endpoint,
failed to get SAML token from ADFS,
invalid_grant,
interaction_required.`

src/main/java/com/microsoft/aad/msal4j/MsalThrottlingException.java
```diff
@@ -0,0 +1,29 @@
...
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
+/**
+ * Exception type thrown when service returns throttling instruction:
+ * Retry-After header, 429 or 5xx statuses.
+ */
+@Accessors(fluent = true)
+@Getter
+public class MsalThrottlingException extends MsalServiceException {
```
bgavrilMS, line 12: As stated above, ideally we wouldn't expose a new type of exception, but just rethrow an old MsalServiceExcpetion.

src/test/java/com/microsoft/aad/msal4j/UIRequiredCacheTest.java
```diff
@@ -0,0 +1,127 @@
...
+import org.powermock.api.easymock.PowerMock;
+import org.testng.Assert;
+import org.testng.annotations.BeforeClass;
+import org.testng.annotations.Test;
+
+import java.util.*;
+
+import static org.easymock.EasyMock.anyObject;
+
+public class UIRequiredCacheTest extends AbstractMsalTests  {
```
bgavrilMS, line 16: Do these tests cover all tests from the spec?

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
SomkaPe: So, you are suggesting to cache all invalid_grant responses?
What about next IWA  scenario: 
1. user get invalid_grant  because kerboros ticket expired.
2. Interaction required exception is thrown.
3. user execute kinit command to get new kerboros ticket.
4. user run IWA again and get cached exception 
5. cached exception will be returned for 2 mins, even though you have 
valid just generated kerberos ticket.
6. worst part about it  - That user is not even informed that he got cached response.

This is just one example of consequences of having dishonest/non_transparent API.

That is why i think caching of responses should be avoided / limited to some well considered specific cases.

Abuse of endpoints for WS-Trust flow is handled by Throttling instructions like Retry_Header, 429, 5xx

SomkaPe: Regarding business scope of MsalInteractionRequiredException:
- I believe that idea behind this exception is to inform that refresh token can not be used, 
so you need to provide credentials.
- if you already sending credentials but they are wrong - What does this message deliver to user ? From my perspective  MSALServiceException with "invalid_grant" error code should be used in this case

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: Under requests we mean set of parameters passed to MSAL api - it is applicable to any request

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: Not sure about it, if we get instructions from STS to throttle some requests, we do it and do it transparently for app develpers.

src/test/java/com/microsoft/aad/msal4j/UIRequiredCacheTest.java
SomkaPe: it covers implemented cases

src/main/java/com/microsoft/aad/msal4j/MsalThrottlingException.java
SomkaPe: bad idea

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: "we should behave as if no "RetryAfter" header was set, i.e. use DEFAULT_THROTTLING_TIME_SEC, at least if response is 429 or 5xx." - that is exactly what happens

src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
SomkaPe: going to add

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
bgavrilMS: Yes, I must have read the code wrong.

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
bgavrilMS: I believe MSALs throw this exception for most (but not all) InvalidGrant and InteractionRequired errors. See this spec for details https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1148 (I think @sangonzal implemented this in Java lib). 

So this exception is not exclusive to `AcquireTokenSilent`, but to all silent flows: 

- AquireTokenSilent
- AcquireTokenInteractive when using a silent web ui
- IWA
- ROPC

As for the IWA scenario you described, it looks more like a corner case. Are you sure AAD throws UiRequiredException there?

bgavrilMS: I sent an email to Sasha for clarifications

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
bgavrilMS: I sent an email to Sasha for clarifications as I am also not sure about this.

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
SomkaPe: "One of common status codes returned from authentication libraries in silent mode today is InvalidGrant. This status code means that the application should call the authentication library again, but in interactive mode. " - sounds pretty straight forward -> it is about silent mode, only one type of requests can be sent in silent mode - refresh token grant

SomkaPe: throttling design also explicitly highlight that scope of responses caching is silent requests.

sangonzal: "it is about silent mode, only one type of requests can be sent in silent mode - refresh token grant" I don't agree with you here. The most meaningful difference IMO between "silent mode" and "non-silent mode" is showing UI. From that perspective the other flows are silent as well. 

InteractionRequired is a valid error to expose for the other flows. What would we expose to users in the case of MFA for example (which is relevant in all flows)? InteractionRequired (pulling up UI and having someone click on it) is the way to resolve that issue, and the most transparent exception to expose to users, whether it's actionable or not for them. 

From the perspective of throttling, wouldn't we also want to throttle applications stuck in a loop in ROPC or IWA flows if they are getting interaction required returned from the server?

sangonzal: One more thing to consider is that changing this behavior (only throwing InteractionRequired for silent requests) is a breaking change, and would require bumping up the major version.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -42,17 +73,79 @@ static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
...
+    }
+
+    private static Integer getRetryAfterHeader(IHttpResponse httpResponse) {
+
+        if (httpResponse.headers() != null) {
+            TreeMap<String, List<String>> headers = new TreeMap<>(String.CASE_INSENSITIVE_ORDER);
+            headers.putAll(httpResponse.headers());
+
+            if (headers.containsKey(RETRY_AFTER_HEADER) && headers.get(RETRY_AFTER_HEADER).size() == 1) {
+                int headerValue = Integer.parseInt(headers.get(RETRY_AFTER_HEADER).get(0));
```
bgavrilMS, line 132: I was looking at the spec for Retry-After and it looks like it can also be a timestamp https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After

Consider using a library call that can parse this.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -6,32 +6,63 @@
...
 
 class HttpHelper {
 
     private final static Logger log = LoggerFactory.getLogger(HttpHelper.class);
+    public static final String RETRY_AFTER_HEADER = "Retry-After";
+    public static final int RETRY_NUM = 2;
+    public static final int RETRY_DELAY_MS = 1000;
+
+    private static String getRequestThumbprint(RequestContext requestContext) {
+        StringBuilder sb = new StringBuilder();
```
bgavrilMS, line 21: [As per spec](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FService%20protection%2FIntial%20set%20of%20protection%20measures.md&version=GBdev&_a=preview&anchor=terminology), this should include the authority as well.

src/main/java/com/microsoft/aad/msal4j/RefreshTokenRequest.java
```diff
@@ -6,18 +6,54 @@
...
     private static AbstractMsalAuthorizationGrant createAuthenticationGrant(
-            RefreshTokenParameters parameters){
+            RefreshTokenParameters parameters) {
 
         RefreshTokenGrant refreshTokenGrant = new RefreshTokenGrant(new RefreshToken(parameters.refreshToken()));
         return new OAuthAuthorizationGrant(refreshTokenGrant, parameters.scopes());
     }
+
+    private String getRefreshTokenRequestFullThumbprint() {
+        StringBuilder sb = new StringBuilder();
```
bgavrilMS, line 42: As per spec, this should include all possible inputs except the refresh token. Might be better to do it based on the form body params...

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -6,32 +6,63 @@
...
+        sb.append(requestContext.clientId() + POINT_DELIMITER);
+        sb.append(requestContext.authority() + POINT_DELIMITER);
+
+        IApiParameters apiParameters = requestContext.apiParameters();
+
+        if (apiParameters instanceof SilentParameters) {
```
bgavrilMS, line 27: AcquireTokenSilent exists both on Public and Confidential clients, but we should only throttle Public client as per spec. Not sure if it's ok to throttle Confiential clients so aggressively. Maybe it is ...

I don't have all the context on how MSAL.java is organized, so maybe this comment isn't relevant.

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
SomkaPe: "InteractionRequired is a valid error to expose for the other flows." - even for non interactive  flows ?

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: it does include authority

src/main/java/com/microsoft/aad/msal4j/RefreshTokenRequest.java
SomkaPe: under request parameters I mean set of parameters passed to corresponding api - silent api or refresh token grant api

src/test/java/com/microsoft/aad/msal4j/RequestThrottlingTest.java
```diff
@@ -0,0 +1,168 @@
...
+        throttlingTest(TokenEndpointResponseType.RETRY_AFTER_HEADER);
+    }
+
+    @Test
+    public void STSResponseContainsStatusCode429_repeatedRequestsThrottled() throws Exception {
+        throttlingTest(TokenEndpointResponseType.STATUS_CODE_429);
+    }
+
+    @Test
+    public void STSResponseContainsStatusCode500_repeatedRequestsThrottled() throws Exception  {
```
bgavrilMS, line 165: Sasha wrote 21 acceptance tests. I was thinking it is too much, but there are subtle differences in each one. It's also a guarantee that libraries behave the same.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -6,32 +6,63 @@
...
+        sb.append(String.join(" ", sortedScopes));
 
-    static IHttpResponse executeHttpRequest(final HttpRequest httpRequest,
-                                            final RequestContext requestContext,
-                                            final ServiceBundle serviceBundle) {
+        return StringHelper.createSha256Hash(sb.toString());
+    }
+
+    static private boolean isRetriable(IHttpResponse httpResponse) {
+        return httpResponse.statusCode() >= 500 && httpResponse.statusCode() < 600;
```
bgavrilMS, line 41: I think we should also not retry if a Retry-After header is present.

SomkaPe: good point - going to change it.

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: good point, but according to design doc we server side are going to use only delta-seconds format

src/test/java/com/microsoft/aad/msal4j/RequestThrottlingTest.java
henrik-me: Please ensure the relevant acceptance tests are included here as that is the best way we have to ensure the code behaves as expected.

---
In reply to: [412276923](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/201#discussion_r412276923) [](ancestors = 412276923)

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: ok, will rename it, was not sure what it correct spelling

src/main/java/com/microsoft/aad/msal4j/MsalServiceExceptionFactory.java
SomkaPe: removing redefinition of InteractionRequired Exception from the scope of this PR

src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
SomkaPe: check for public client happens in checkForThrottling and processThrottlingInstructions functions

src/test/java/com/microsoft/aad/msal4j/RequestThrottlingTest.java
SomkaPe: Most of mentioned tests duplicated, and some not applicable for java.
Taking to account differences in api model of Msal and WAM we consider only refresh tokens grant requests as subject to UIRequired cache

Issue #180, opened 2020-02-19
Improve authority validation


sangonzal:
When creating a new application object, passing in "https://authority.com//tenant" fails with "java.lang.ArrayIndexOutOfBoundsException". 

We should either: 
- update authority validation logic should check that the value immediately after the first forward slash (the first value in the path) is valid, and if not throw a meaningful error
- Update authority creation logic to combine multiple forward slashes into a single forward slash

vikasbo:
# Steps to reproduce

> What steps can reproduce the defect?

```
IPartnerCredentials appCredentials = 
      PartnerCredentials.getInstance()
      .generateByApplicationCredentials(clientId, applicationSecret, aadApplicationDomain)
```
throws:

```
2020-02-12 16:33:00,191 155022 ERROR [com.microsoft.aad.msal4j.ConfidentialClientApplication] (ForkJoinPool.commonPool-worker-1:) [Correlation ID: d1562f78-c8b2-4f14-9937-350080770ada] Execution of class com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier failed.
    java.lang.ArrayIndexOutOfBoundsException: 0
    	at com.microsoft.aad.msal4j.Authority.getTenant(Authority.java:88)
    	at com.microsoft.aad.msal4j.AadInstanceDiscoveryProvider.sendInstanceDiscoveryRequest(AadInstanceDiscoveryProvider.java:110)
    	at com.microsoft.aad.msal4j.AadInstanceDiscoveryProvider.doInstanceDiscoveryAndCache(AadInstanceDiscoveryProvider.java:131)
    	at com.microsoft.aad.msal4j.AadInstanceDiscoveryProvider.getMetadataEntry(AadInstanceDiscoveryProvider.java:42)
    	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.getAuthorityWithPrefNetworkHost(AuthenticationResultSupplier.java:32)
    	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.execute(AcquireTokenByAuthorizationGrantSupplier.java:49)
    	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:59)
    	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:17)
[...]

PR #202, opened 2020-03-31
authority Path validation

SomkaPe:
Validation of Authority path segments + refactoring

issue https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/180

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -59,6 +59,8 @@ abstract class Authority {
             return new AADAuthority(authorityUrl);
         } else if (authorityType == AuthorityType.B2C) {
             return new B2CAuthority(authorityUrl);
+        } else if (authorityType == AuthorityType.ADFS) {
+            return new ADFSAuthority(authorityUrl);
         } else {
             throw new IllegalArgumentException("Unsupported Authority Type");
         }
@@ -76,7 +78,6 @@ abstract class Authority {
         }
 
         final String firstPath = path.substring(0, path.indexOf("/"));
-
 
         if (isB2CAuthority(firstPath)) {
             return AuthorityType.B2C;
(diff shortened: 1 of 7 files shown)


src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -76,35 +78,56 @@ static AuthorityType detectAuthorityType(URL authorityUrl) {
...
             return AuthorityType.AAD;
         }
     }
 
-    void validateAuthorityUrl() {
-        if (!this.canonicalAuthorityUrl.getProtocol().equalsIgnoreCase("https")) {
+    static void validateAuthority(URL authorityUrl) {
+        if (!authorityUrl.getProtocol().equalsIgnoreCase("https")) {
             throw new IllegalArgumentException(
                     "authority should use the 'https' scheme");
```
sangonzal, line 93: Might be a good opportunity to move these error messages to AuthenticationErrorMessage or a constants file.

src/test/java/com/microsoft/aad/msal4j/AuthorityTest.java
```diff
@@ -169,4 +152,39 @@ public void testDoStaticInstanceDiscovery_ValidateFalse_TrustedAuthority()
...
+                "https://login.microsoftonline.com/tenant//",
+                "https://login.microsoftonline.com////tenant//path1"};
+
+        for (String testAuthority : authoritiesWithInvalidPath) {
+            try {
+                URL authorityUrl = new URL(testAuthority);
+                Authority.validateAuthority(authorityUrl);
+            } catch (Exception ex) {
+                Assert.assertTrue(ex instanceof IllegalArgumentException);
+                Assert.assertEquals
```
sangonzal, line 171: Instead of having duplicated code below, can you add "https://login.microsoftonline.com" to this array and check if(testAuthority =="https://login.microsoftonline.com" ), then assert the specific error message.

src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -51,15 +51,17 @@ private void setCommonAuthorityProperties() {
...
-           return new AADAuthority(authorityUrl);
-       } else if(authorityType == AuthorityType.B2C) {
-           return new B2CAuthority(authorityUrl);
-       } else {
-           throw new IllegalArgumentException("Unsupported Authority Type");
-       }
+    static Authority createAuthority(URL authorityUrl) {
+        validateAuthority(authorityUrl);
+
+        AuthorityType authorityType = detectAuthorityType(authorityUrl);
```
sangonzal, line 57: Looks like `detectAuthorityType` can return AuthorityType.ADFS, but that would throw "Unsupported Authority Type" error, which is actually supported. Right?

src/main/java/com/microsoft/aad/msal4j/Authority.java
SomkaPe: it is not AuthenticationErrorMessage it is IllegalArgumentException message - going to add corresponding constants

src/test/java/com/microsoft/aad/msal4j/AuthorityTest.java
SomkaPe: generally speaking test data which leads to different outcomes should be separated, refactored test

PR #216, opened 2020-04-17
Update cache miss log level

sangonzal:
Update log level of AuthenticationErrorCode.CACHE_MISS to from ERROR to DEBUG

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java
@@ -85,11 +85,7 @@ abstract class AuthenticationResultSupplier implements Supplier<IAuthenticationR
                         msalRequest.requestContext().correlationId(),
                         error);
 
-                clientApplication.log.error(
-                        LogHelper.createMessage(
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java
```diff
@@ -132,6 +128,22 @@ private void logResult(AuthenticationResult result, HttpHeaders headers) {
         }
     }
 
+     private void logException(Exception ex) {
+
+         String logMessage = LogHelper.createMessage("Execution of " + this.getClass()
+                 + " failed.", msalRequest.headers().getHeaderCorrelationIdValue());
+
+         if (ex instanceof MsalClientException) {
```
SomkaPe, line 136: probably we need to log not MsalClientExceptions also, looks like clientApplication.log.error(logMessage, ex); line should be moved from if closure

sangonzal: Good catch, updated.

SomkaPe:
:shipit:

PR #221, opened 2020-04-25
Fix for private jwt client to rebuild when expired

Budlee:
The PrivateJwt assertion with a certificate is generated once
This means that when a request for a new OAuth Token is made that
the client assertion has expired and fails. This change
fixes that by reubuilding the assersion for a private jwt when
it has expired

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
+++ b/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
@@ -12,6 +12,7 @@ import com.nimbusds.oauth2.sdk.id.ClientID;
 import org.slf4j.LoggerFactory;
 
 import java.util.Collections;
+import java.util.Date;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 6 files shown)


src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -71,19 +76,37 @@ private void initClientAuthentication(IClientCredential clientCredential) {
...
         }
     }
 
+    @Override
+    protected ClientAuthentication clientAuthentication() {
+        if (clientCertAuthentication) {
+            final Date currentDateTime = new Date(System.currentTimeMillis());
+            final Date expirationTime = ((PrivateKeyJWT) clientAuthentication).getJWTAuthenticationClaimsSet().getExpirationTime();
+            if (expirationTime.before(currentDateTime)) {
+                //The asserted private jwt with the client certificate can expire so rebuild it when the
```
sangonzal, line 95: Comment is missing a couple of words.

src/test/java/com/microsoft/aad/msal4j/ConfidentialClientApplicationTest.java
```diff
@@ -153,4 +164,70 @@ public void testAcquireToken_KeyCred() throws Exception {
         PowerMock.verifyAll();
         PowerMock.resetAll(app);
     }
+
+    @Test
+    public void testClientCertificateRebuildsWhenExpired() throws Exception {
```
sangonzal, line 169: Expiration is misspelled in a couple of places.

Issue #197, opened 2020-03-18
[Feature Request] Client capabilities support


henrik-me:
Smart Session Management (SSM) sometimes referred to as the long lived tokens (LLT) scenario require MSAL to allow developers to resolve challenges (formatted as claims request parameters) to be resolved silently rather than interactively.  Current releases of MSAL assume that interaction is required whenever a claims request parameter is included in the request.  

This change in behavior requires the client to indicate that it's aware of a new calling pattern relative to the use of MSAL.

See additional details of requirements in:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1545

API/Design details:
https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FClientCapabilities%2Foverview.md&version=GBdev
**Is your feature request related to a problem? Please describe.**

Related internal work item:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/824806

henrik-me:
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1545

PR #226, opened 2020-04-30
Client Capabilities Support

Avery-Dunn:
Added support for client capabilities, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/197

Main functionality is to create a new parameter that can be set as part of an authorization request, and this new parameter is merged into the existing claims parameter. 

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1641 for a related implementation in the .NET library

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
@@ -44,10 +44,10 @@ public class AuthorizationCodeIT extends SeleniumTest {
         Map<String, Set<String>> claimsAndCapabilities = new HashMap<>();
 
         //Two separate claims used to test merging claim JSONs
-        claims.add("{\"userinfo\":{\"given_name\":{\"essential\":true},\"nickname\":null,\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true},\"picture\":null,\"http://example.info/claims/groups\":null}}");
-        claims.add("{\"id_token\":{\"auth_time\":{\"essential\":true}}}");
+        claims.add(TestConstants.CLAIMS_USERINFO);
+        claims.add(TestConstants.CLAIMS_IDTOKEN);
         //Client Capabilities with no values, as client capabilities were not yet supported when this test was written
-        clientCapabilities.add("{\"access_token\":{\"xms_cc\":{\"values\":[]}}}");
+        clientCapabilities.add(TestConstants.CLIENT_CAPABILITIES);
 
         claimsAndCapabilities.put("claims", claims);
         claimsAndCapabilities.put("clientCapabilities", clientCapabilities);
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
@@ -33,7 +33,8 @@ public class AuthorizationCodeIT extends SeleniumTest {
         assertAcquireTokenAAD(user, null);
     }
 
-    @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
+    //TODO: Re-enable test once list of claims/capabilities and their expected behavior is known
+    //@Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
     public void acquireTokenWithAuthorizationCode_ManagedUserWithClaimsAndCapabilities(String environment){
         cfg = new Config(environment);
 
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
@@ -39,18 +39,10 @@ public class AuthorizationCodeIT extends SeleniumTest {
 
         User user = labUserProvider.getDefaultUser(cfg.azureEnvironment);
 
-        Set<String> claims = new HashSet<>();
-        Set<String> clientCapabilities = new HashSet<>();
         Map<String, Set<String>> claimsAndCapabilities = new HashMap<>();
 
-        //Two separate claims used to test merging claim JSONs
-        claims.add(TestConstants.CLAIMS_USERINFO);
-        claims.add(TestConstants.CLAIMS_IDTOKEN);
-        //Client Capabilities with no values, as client capabilities were not yet supported when this test was written
-        clientCapabilities.add(TestConstants.CLIENT_CAPABILITIES);
-
-        claimsAndCapabilities.put("claims", claims);
-        claimsAndCapabilities.put("clientCapabilities", clientCapabilities);
+        claimsAndCapabilities.put("claims", Collections.singleton(TestConstants.CLAIMS));
+        claimsAndCapabilities.put("clientCapabilities", TestConstants.CLIENT_CAPABILITIES_LLT);
 
         assertAcquireTokenAAD(user, claimsAndCapabilities);
     }
@@ -333,7 +325,7 @@ public class AuthorizationCodeIT extends SeleniumTest {
                 AuthorizationRequestUrlParameters
                         .builder(TestConstants.LOCALHOST + httpListener.port(),
                                 Collections.singleton(scope))
-                        .claims(parameters.getOrDefault("claims", null))
+                        .claims(String.valueOf(parameters.getOrDefault("claims", null).toArray()[0]))
                         .clientCapabilities(parameters.getOrDefault("clientCapabilities", null))
                         .build();
 
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
@@ -152,10 +152,18 @@ public class AuthorizationCodeIT extends SeleniumTest {
 
         PublicClientApplication pca;
         try {
-            pca = PublicClientApplication.builder(
-                    user.getAppId()).
-                    authority(cfg.organizationsAuthority()).
-                    build();
+            if (parameters != null) {
+                pca = PublicClientApplication.builder(
+                        user.getAppId()).
+                        authority(cfg.organizationsAuthority()).
+                        clientCapabilities(parameters.getOrDefault("clientCapabilities", null)).
+                        build();
+            } else {
+                pca = PublicClientApplication.builder(
+                        user.getAppId()).
+                        authority(cfg.organizationsAuthority()).
+                        build();
+            }
         } catch(MalformedURLException ex){
             throw new RuntimeException(ex.getMessage());
         }
@@ -325,8 +333,7 @@ public class AuthorizationCodeIT extends SeleniumTest {
                 AuthorizationRequestUrlParameters
                         .builder(TestConstants.LOCALHOST + httpListener.port(),
                                 Collections.singleton(scope))
-                        .claims(String.valueOf(parameters.getOrDefault("claims", null).toArray()[0]))
-                        .clientCapabilities(parameters.getOrDefault("clientCapabilities", null))
+                        .claims(String.valueOf(parameters.getOrDefault("claims", Collections.singleton("")).toArray()[0]))
                         .build();
 
         return app.getAuthorizationRequestUrl(requestUrlParameters).toString();
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
@@ -315,7 +315,6 @@ public class AuthorizationCodeIT extends SeleniumTest {
 
     private String buildAuthenticationCodeURLWithParameters(ClientApplicationBase app, Map<String, Set<String>> parameters) {
         String scope;
-        Set<String> claims = new HashSet<>();
 
         AuthorityType authorityType= app.authenticationAuthority.authorityType;
         if(authorityType == AuthorityType.AAD){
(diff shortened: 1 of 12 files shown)


src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
```diff
@@ -155,6 +164,30 @@ private void assertAcquireTokenAAD(User user){
...
+        IAuthenticationResult result = acquireTokenAuthorizationCodeFlow(
+                pca,
+                authCode,
+                Collections.singleton(cfg.graphDefaultScope()));
+
+        Assert.assertNotNull(result);
+        Assert.assertNotNull(result.accessToken());
+        Assert.assertNotNull(result.idToken());
+        Assert.assertEquals(user.getUpn(), result.account().username());
+    }
```
sangonzal, line 189: I think you should be able to assert that the id token that gets returned has the claims that you requested.

src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
```diff
@@ -30,6 +31,14 @@ public void acquireTokenWithAuthorizationCode_ManagedUser(String environment){
         assertAcquireTokenAAD(user);
     }
 
+    @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
+    public void acquireTokenWithAuthorizationCode_ManagedUserWithClaims(String environment){
+        cfg = new Config(environment);
+
+        User user = labUserProvider.getDefaultUser(cfg.azureEnvironment);
+        assertAcquireTokenAADWithClaims(user);
```
sangonzal, line 39: I think you should be able to refactor some of the code that you added, as there is a bit of duplication. For example, consider using `assertAcquireTokenAAD`, and passing an extra parameter for claims. You should then be able to use of all of the same methods, and in `buildAuthenticationCodeURL` you could check if claims were passed in, and if so then add them to the client application.

src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -29,4 +31,73 @@
             throw new MsalClientException(e);
         }
     }
+
+    /**
+     * Merges given JSON strings into one Jackson JSONNode object, which is returned as a String
+     */
+    static String mergeJSONString(String mainJsonString, String addJsonString) {
+        JsonNode mainJson, addJson;
```
sangonzal, line 39: nit: split these into two lines. 

For the time being, I've been using the [Google Java style](https://google.github.io/styleguide/javaguide.html#s4.3-one-statement-per-line) for guidance. We should all agree on guidelines that we agree on and ideally add a linter to the build.

src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -29,4 +31,73 @@
...
+        try {
+            mainJson = mapper.readTree(mainJsonString);
+            addJson = mapper.readTree(addJsonString);
+        } catch (JsonProcessingException e) {
+            throw new MsalClientException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
+        }
+
+        mergeJSONNode(mainJson, addJson);
+
+        return mainJson != null ? mainJson.toString() : null;
```
sangonzal, line 49: I think we should throw if null here, instead of returning null - there is no scenario I can think of where you would call `mergeJSONString` and have null be a valid return value. This might save us from getting a null ref error somewhere else.

src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -29,4 +31,73 @@
...
+        while (jsons.hasNext()) {
+            try {
+                addJson = mapper.readTree(jsons.next());
+            } catch (JsonProcessingException e) {
+                throw new MsalClientException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
+            }
+            mergeJSONNode(mainJson, addJson);
+        }
+
+        return mainJson != null ? mainJson.toString() : null;
```
sangonzal, line 76: Same as above

src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
```diff
@@ -280,4 +313,80 @@ else if (authorityType == AuthorityType.ADFS){
...
+        }
+        return result.code();
+    }
+
+    private String buildAuthenticationCodeURLWithClaims(ClientApplicationBase app) {
+        String scope;
+        Set<String> claims = new HashSet<>();
+
+        //Two separate claims used to test merging claim JSONs
+        claims.add("{\"userinfo\":{\"given_name\":{\"essential\":true},\"nickname\":null,\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true},\"picture\":null,\"http://example.info/claims/groups\":null}}");
```
sangonzal, line 363: We briefly discussed this the other day, and after seeing it I think it would make sense to add these to the test constants file

src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
Avery-Dunn: I refactored it to remove most of the duplication by adding a new parameter to `assertAcquireTokenAAD` to hold the auth request URL parameters, which is just passed along until `acquireAuthorizationCodeAutomated`. There, it's used to either build the URL using an existing method (with no optional params), or build the URL with a new method that can handle optional params. For now it just looks for claims+capabilities, but it can be extended to handle tests of other optional params as needed.

src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -72,10 +72,20 @@ private AuthorizationRequestUrlParameters(Builder builder){
...
             requestParameters.put("claims", Collections.singletonList(claimsParam));
         }
 
+        if(builder.clientCapabilities != null){
+            String capabilitiesParam = JsonHelper.mergeJSONString(builder.clientCapabilities);
+            if (requestParameters.containsKey("claims")) {
+                String mergedClaimsAndCapabilities = JsonHelper.mergeJSONString(String.valueOf(requestParameters.get("claims").get(0)), capabilitiesParam);
+                requestParameters.replace("claims", Collections.singletonList(mergedClaimsAndCapabilities));
+            } else {
+                requestParameters.put("claims", Collections.singletonList(capabilitiesParam));
```
bgavrilMS, line 85: Please note that claim and client capabilities should be sent to /token endpoint as well.

bgavrilMS: image

bgavrilMS:
Although slightly off-topic, I encountered the following issue while implementing this on .NET - it looks like there is an expectation that if the user configures `claims`, then the token cache is bypassed (AcquireTokenSilent will behave as if ForceRefres=true). But if they configure only `client capabilities`, then the token cache is not bypassed. I believe this is how Android team also implemented this.

SomkaPe:
Not sure about sending claims to token endpoint - according to spec https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter
it is an optional parameter for Authorization Request

src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -187,11 +199,20 @@ public Builder scopes(Set<String> val){
          * policy has not been met,{@link MsalServiceException} will contain claims that need be
          * consented to.
          */
-        public Builder claims(Set<String> val){
+        public Builder claims(String val){
```
sangonzal, line 202: This is a breaking change, we would have to bump up major version.

src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -187,11 +199,20 @@ public Builder scopes(Set<String> val){
...
             this.claims = val;
             return self();
         }
 
+        /**
+         * Allows the client to indicate to the token service what policies or features it is
+         * capable of handling
+         */
+        public Builder clientCapabilities(Set<String> val){
```
sangonzal, line 211: Might have missed the latest discussion around this, but I thought client capabilities should be added to the client application object?

src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
```diff
@@ -121,5 +123,6 @@ public void testBuilder_optionalParameters() throws UnsupportedEncodingException
         Assert.assertEquals(queryParameters.get("correlation_id"), "corr_id");
         Assert.assertEquals(queryParameters.get("login_hint"), "hint");
         Assert.assertEquals(queryParameters.get("domain_hint"), "domain_hint");
+        Assert.assertEquals(queryParameters.get("claims"), "{\"userinfo\":{\"given_name\":{\"essential\":true},\"nickname\":null,\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true},\"picture\":null,\"http://example.info/claims/groups\":null},\"access_token\":{\"xms_cc\":{\"values\":[]}}}");
```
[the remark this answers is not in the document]
Avery-Dunn: These were just copied over from the integration test, where having values in that field would cause the server to return an error (since we don't know the valid capabilities values, or even if there are any valid ones). This test now takes in a set of two values, to properly testing merging of the capabilities set and the merging of capabilities into claims.

src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
```diff
@@ -92,6 +92,8 @@ public void testBuilder_optionalParameters() throws UnsupportedEncodingException
                         .correlationId("corr_id")
                         .loginHint("hint")
                         .domainHint("domain_hint")
+                        .claims("{\"userinfo\":{\"given_name\":{\"essential\":true},\"nickname\":null,\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true},\"picture\":null,\"http://example.info/claims/groups\":null}}")
+                        .clientCapabilities(Collections.emptySet())
```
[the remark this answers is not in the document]
Avery-Dunn: This test now covers the case when claims and capabilities both have values in 'access_token', and the integration test covers the case when only claims are sent.

src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
```diff
@@ -92,6 +92,8 @@ public void testBuilder_optionalParameters() throws UnsupportedEncodingException
                         .correlationId("corr_id")
                         .loginHint("hint")
                         .domainHint("domain_hint")
+                        .claims("{\"userinfo\":{\"given_name\":{\"essential\":true},\"nickname\":null,\"email\":{\"essential\":true},\"email_verified\":{\"essential\":true},\"picture\":null,\"http://example.info/claims/groups\":null}}")
```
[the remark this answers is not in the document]
Avery-Dunn: This test is just for making sure the formatting is correct after merging claims with capabilities, so the actual values don't matter (the values originally came from an example in the OIDC claims spec). Regardless, they've now been updated with values that lead to a more robust test of the merging, though I'm sure it's still not a 'real' situation.

src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -171,6 +175,14 @@ public URL getAuthorizationRequestUrl(AuthorizationRequestUrlParameters paramete
 
         parameters.requestParameters.put("client_id", Collections.singletonList(this.clientId));
 
+        if (this.clientCapabilities != null) {
```
SomkaPe, line 178: Please add comment here

src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
```diff
@@ -121,5 +127,6 @@ public void testBuilder_optionalParameters() throws UnsupportedEncodingException
         Assert.assertEquals(queryParameters.get("correlation_id"), "corr_id");
         Assert.assertEquals(queryParameters.get("login_hint"), "hint");
         Assert.assertEquals(queryParameters.get("domain_hint"), "domain_hint");
+        Assert.assertEquals(queryParameters.get("claims"), "{\"id_token\":{\"auth_time\":{\"essential\":true}},\"access_token\":{\"auth_time\":{\"essential\":true},\"xms_cc\":{\"values\":[\"llt\",\"ssm\"]}}}");
     }
 }
```
SomkaPe, line 132: client capabilities should be sent to token endpoint as well, here it is send only to authorization endpoint

src/integrationtest/java/com.microsoft.aad.msal4j/AuthorizationCodeIT.java
```diff
@@ -131,19 +149,27 @@ private void assertAcquireTokenADFS2019(User user){
...
+                        user.getAppId()).
+                        authority(cfg.organizationsAuthority()).
+                        clientCapabilities(parameters.getOrDefault("clientCapabilities", null)).
+                        build();
+            } else {
+                pca = PublicClientApplication.builder(
+                        user.getAppId()).
+                        authority(cfg.organizationsAuthority()).
+                        build();
+            }
```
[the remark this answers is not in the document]
Avery-Dunn: Got rid of most of the code duplication, variables are now set at start of method rather than having two different builders.

src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
Avery-Dunn: In the latest commit, I changed the original 'claims' parameter back to what it was and marked it as deprecated, and added a new 'claimsChallenge' parameter to handle all of the new behavior without being a breaking change.

src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
Avery-Dunn: In the latest commit capabilities are added to the parameters of requests to the token endpoint.

Issue #237, opened 2020-05-12
Enable device code flow for ADFS 2019


sangonzal:
When ADFS 2019 support was added, it did not support device code flow. That support has been added since, and it can be enabled in the library.

PR #241, opened 2020-05-29
Enable device code flow for ADFS 2019

Avery-Dunn:
Allows device code flow to for ADFS users, per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/237

- Adds device code endpoint parameter to Authority objects
- Changes device code's HTTP request from GET to POST, moves query parameters to body
- Adds integration test for device code with ADFS user, removes unit tests looking for exception in that situation

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
@@ -35,7 +35,7 @@ public class DeviceCodeIT {
     }
 
     @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
-    public void DeviceCodeFlowTest(String environment) throws Exception {
+    public void DeviceCodeFlowADTest(String environment) throws Exception {
         cfg = new Config(environment);
 
         User user = labUserProvider.getDefaultUser(cfg.azureEnvironment);
@@ -59,22 +59,57 @@ public class DeviceCodeIT {
         Assert.assertTrue(!Strings.isNullOrEmpty(result.accessToken()));
     }
 
+    @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
+    public void DeviceCodeFlowADFSv2019Test(String environment) throws Exception {
+        cfg = new Config(environment);
+
+        User user = labUserProvider.getOnPremAdfsUser(FederationProvider.ADFS_2019);
+
+        PublicClientApplication pca = PublicClientApplication.builder(
+                TestConstants.ADFS_APP_ID).
+                authority(TestConstants.ADFS_AUTHORITY).validateAuthority(false).
+                build();
+
+        Consumer<DeviceCode> deviceCodeConsumer = (DeviceCode deviceCode) -> {
+            runAutomatedDeviceCodeFlow(deviceCode, user);
+        };
+
+        IAuthenticationResult result = pca.acquireToken(DeviceCodeFlowParameters
+                .builder(Collections.singleton(TestConstants.ADFS_SCOPE),
+                        deviceCodeConsumer)
+                .build())
+                .get();
+
+        Assert.assertNotNull(result);
+        Assert.assertFalse(Strings.isNullOrEmpty(result.accessToken()));
+    }
+
     private void runAutomatedDeviceCodeFlow(DeviceCode deviceCode, User user){
         boolean isRunningLocally = true;//!Strings.isNullOrEmpty(
                 //System.getenv(TestConstants.LOCAL_FLAG_ENV_VAR));
+
+        boolean isADFS2019 = user.getFederationProvider().equals("adfsv2019");
 
         LOG.info("Device code running locally: " + isRunningLocally);
         try{
             String deviceCodeFormId;
             String continueButtonId;
             if(isRunningLocally){
-                deviceCodeFormId = "otc";
-                continueButtonId = "idSIButton9";
+                if (isADFS2019) {
+                    deviceCodeFormId = "userCodeInput";
+                    continueButtonId = "confirmationButton";
+                } else {
+                    deviceCodeFormId = "otc";
+                    continueButtonId = "idSIButton9";
+                }
             } else {
                 deviceCodeFormId = "code";
                 continueButtonId = "continueBtn";
             }
             LOG.info("Loggin in ... Entering device code");
+            if (isADFS2019) {
+                seleniumDriver.manage().deleteAllCookies();
+            }
             seleniumDriver.navigate().to(deviceCode.verificationUri());
             seleniumDriver.findElement(new By.ById(deviceCodeFormId)).sendKeys(deviceCode.userCode());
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
```diff
@@ -55,11 +55,6 @@
     @Override
     public CompletableFuture<IAuthenticationResult> acquireToken(DeviceCodeFlowParameters parameters) {
 
-        if (!AuthorityType.AAD.equals(authenticationAuthority.authorityType())) {
-            throw new IllegalArgumentException(
-                    "Invalid authority type. Device Flow is only supported by AAD authority");
-        }
-
```
SomkaPe, line 62: Is device code flow supported by B2C Authority ?

src/main/java/com/microsoft/aad/msal4j/DeviceCodeFlowRequest.java
```diff
@@ -41,10 +41,17 @@ DeviceCode acquireDeviceCode(String url,
                                  Map<String, String> clientDataHeaders,
                                  ServiceBundle serviceBundle) throws Exception {
 
-        String urlWithQueryParams = createQueryParamsAndAppendToURL(url, clientId);
```
SomkaPe, line 44: this function is not used any more? should it be deleted ?
it might make sense to keep the code which creates query parameters incapsulated in this function

src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -36,6 +36,17 @@
     String authorizationEndpoint;
     String tokenEndpoint;
 
+    final static String DEVICE_CODE_ENDPOINT = "oauth2/devicecode";
+
+    private final static String ADFS_AUTHORITY_FORMAT = "https://%s/%s/";
+    private final static String DEVICE_CODE_ENDPOINT_FORMAT = ADFS_AUTHORITY_FORMAT + DEVICE_CODE_ENDPOINT;
```
SomkaPe, line 42: why general authority contains fields from child class ?

src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
```diff
@@ -59,22 +59,57 @@ public void DeviceCodeFlowTest(String environment) throws Exception {
...
+                    continueButtonId = "idSIButton9";
+                }
             } else {
                 deviceCodeFormId = "code";
                 continueButtonId = "continueBtn";
             }
             LOG.info("Loggin in ... Entering device code");
+            if (isADFS2019) {
+                seleniumDriver.manage().deleteAllCookies();
+            }
```
sangonzal, line 112: I think you can just extend the SeleniumTest class, and that takes care of cleaning up for you.

src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
sangonzal: B2C does not support device code flow. We should check if authority is B2C, and then throw error if so

src/test/java/com/microsoft/aad/msal4j/DeviceCodeFlowTest.java
```diff
@@ -134,54 +134,21 @@ public void deviceCodeFlowTest() throws Exception {
...
-
-        app.acquireToken
-                (DeviceCodeFlowParameters
-                        .builder(Collections.singleton(AAD_RESOURCE_ID), (DeviceCode deviceCode) -> {})
-                        .build());
-    }
-
-    @Test(expectedExceptions = IllegalArgumentException.class,
-            expectedExceptionsMessageRegExp = "Invalid authority type. Device Flow is only supported by AAD authority")
-    public void executeAcquireDeviceCode_B2CAuthorityUsed_IllegalArgumentExceptionThrown()
```
sangonzal, line 171: We should keep this test, as B2C is still not supported for device code.

src/main/java/com/microsoft/aad/msal4j/DeviceCodeFlowRequest.java
Avery-Dunn: In the latest commit I refactored the method from "createQueryParamsAndAppendToURL" to just "createQueryParams", and moved all of the logic for setting up the query parameters (which go into the request body now instead of URL) back into that method.

src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
Avery-Dunn: For some reason I originally read that as "if ADFS" rather than "if not AD", so I incorrectly got rid of the logic for blocking B2C. I re-added the authority type check and exception throw, so that it now throws the exception if the given authority is not AD or ADFS, and also re-add the related unit test.

src/test/java/com/microsoft/aad/msal4j/DeviceCodeFlowTest.java
Avery-Dunn: I just re-added that test in the latest commit. I originally misinterpreted the exception this test covers when I first looked at the flow, but I re-added the logic for that exception as well.

src/main/java/com/microsoft/aad/msal4j/Authority.java
Avery-Dunn: It was leftover from when I initially started working on it and thought all authorities would now need to the device code endpoint (forgot about B2C authority). In the latest commit I cleaned it up to follow better code practices.

src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -36,6 +36,10 @@
     String authorizationEndpoint;
     String tokenEndpoint;
 
+    private final static String ADFS_AUTHORITY_FORMAT = "https://%s/%s/";
```
SomkaPe, line 39: ADFS authority format should be inside ADFS Authority, Authority class represents common authority

src/main/java/com/microsoft/aad/msal4j/ADFSAuthority.java
```diff
@@ -9,14 +9,23 @@
 
     final static String AUTHORIZATION_ENDPOINT = "oauth2/authorize";
     final static String TOKEN_ENDPOINT = "oauth2/token";
+    final static String DEVICE_CODE_ENDPOINT = "oauth2/devicecode";
 
     private final static String ADFS_AUTHORITY_FORMAT = "https://%s/%s/";
+    private final static String DEVICE_CODE_ENDPOINT_FORMAT = ADFS_AUTHORITY_FORMAT + DEVICE_CODE_ENDPOINT;
+
+    String deviceCodeEndpoint;
```
sangonzal, line 17: I think you should be able to move need this and the getter to 'Authority'. You can also get rid of 'deviceCodeEnpoint' in AADAuthority as well, and just set it on the parent class based on the child classes format.

Avery-Dunn: Fixed this and your other comments in the latest commit.

src/main/java/com/microsoft/aad/msal4j/ADFSAuthority.java
```diff
@@ -9,14 +9,19 @@
...
     ADFSAuthority(final URL authorityUrl) {
         super(authorityUrl, AuthorityType.ADFS);
         this.authority = String.format(ADFS_AUTHORITY_FORMAT, host, tenant);
         this.authorizationEndpoint = authority + AUTHORIZATION_ENDPOINT;
         this.tokenEndpoint = authority + TOKEN_ENDPOINT;
         this.selfSignedJwtAudience = this.tokenEndpoint;
+        this.deviceCodeEndpoint = String.format(DEVICE_CODE_ENDPOINT_FORMAT, host, tenant);
     }
+
+
```
sangonzal, line 26: nit: extra line

Issue #240, opened 2020-05-29
Unable to acquire token


RajatNair:
Hi,

The following snippet of code was working fine and generating an access token couple of days back -

```Java
private static IAuthenticationResult acquireToken() throws Exception {

     // Load token cache from file and initialize token cache aspect. The token cache will have
     // dummy data, so the acquireTokenSilently call will fail.
     TokenCacheAspect tokenCacheAspect = new TokenCacheAspect("sample_cache.json");

     IClientCredential credential = ClientCredentialFactory.createFromSecret(CLIENT_SECRET);
     ConfidentialClientApplication cca =
             ConfidentialClientApplication
                     .builder(CLIENT_ID, credential)
                     .authority(AUTHORITY)
                     .setTokenCacheAccessAspect(tokenCacheAspect)
                     .build();

     IAuthenticationResult result;
     try {
         SilentParameters silentParameters =
                 SilentParameters
                         .builder(SCOPE)
                         .build();

         // try to acquire token silently. This call will fail since the token cache does not
         // have a token for the application you are requesting an access token for
         result = cca.acquireTokenSilently(silentParameters).join();
     } catch (Exception ex) {
         if (ex.getCause() instanceof MsalException) {

             ClientCredentialParameters parameters =
                     ClientCredentialParameters
                             .builder(SCOPE)
                             .build();

             // Try to acquire a token. If successful, you should see
             // the token information printed out to console
             result = cca.acquireToken(parameters).join();
         } else {
             // Handle other exceptions accordingly
             throw ex;
         }
     }
     return result;
 }
```
[Source](https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-daemon-acquire-token?tabs=java#acquiretokenforclient-api)

but today this exception is throw

> java.util.concurrent.CompletionException: java.lang.NoSuchMethodError: net.minidev.json.JSONObject.getAsNumber(Ljava/lang/String;)Ljava/lang/Number;

Current setup - 
- Apache Maven 3.6.3
- Java 1.8.0_252
- Spring Boot 2.3.0.RELEASE
- msal4j 1.5.0

Was using ms-graph-sdk 1.7.1 and ms-graph-core 1.0.1 but got same error even after removing them.

Pulling my hair out trying to figure out root cause. Not sure which dependency is causing this error. I see that msal4j uses net.minidev's json-smart 1.3.1
[...]

Issue #213, opened 2020-04-15
Instance discovery fails in Azure Databricks 6.4


lyytinen:
Instance discovery fails in Azure Databricks 6.4 when using msal4j 1.4.0:

```
Caused by: java.util.concurrent.CompletionException: com.microsoft.aad.msal4j.MsalClientException: com.fasterxml.jackson.databind.JsonMappingException: No suitable constructor found for type [simple type, class com.microsoft.aad.msal4j.InstanceDiscoveryMetadataEntry]: can not instantiate from JSON object (missing default constructor or creator, or perhaps need to add/enable type information?)
 at [Source: {"tenant_discovery_endpoint":"https://login.microsoftonline.com/<tenant-id-removed>/v2.0/.well-known/openid-configuration","api-version":"1.1","metadata":[{"preferred_network":"login.microsoftonline.com","preferred_cache":"login.windows.net","aliases":["login.microsoftonline.com","login.windows.net","login.microsoft.com","sts.windows.net"]},{"preferred_network":"login.partner.microsoftonline.cn","preferred_cache":"login.partner.microsoftonline.cn","aliases":["login.partner.microsoftonline.cn","login.chinacloudapi.cn"]},{"preferred_network":"login.microsoftonline.de","preferred_cache":"login.microsoftonline.de","aliases":["login.microsoftonline.de"]},{"preferred_network":"login.microsoftonline.us","preferred_cache":"login.microsoftonline.us","aliases":["login.microsoftonline.us","login.usgovcloudapi.net"]},{"preferred_network":"login-us.microsoftonline.com","preferred_cache":"login-us.microsoftonline.com","aliases":["login-us.microsoftonline.com"]}]}; line: 1, column: 174] (through reference chain: com.microsoft.aad.msal4j.AadInstanceDiscoveryResponse["metadata"]->Object[][0])
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:93)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:17)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1596)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
[...]

PR #242, opened 2020-06-02
compatibility with jackson 2.6.7 and json-smart 1.3.1

SomkaPe:
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/240
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/213

### DIFF
--- a/src/integrationtest/java/labapi/LabService.java
+++ b/src/integrationtest/java/labapi/LabService.java
@@ -3,11 +3,11 @@
 
 package labapi;
 
-import com.fasterxml.jackson.core.JsonProcessingException;
 import com.fasterxml.jackson.databind.DeserializationFeature;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.microsoft.aad.msal4j.*;
 
+import java.io.IOException;
 import java.net.MalformedURLException;
 import java.util.Collections;
 import java.util.HashMap;
@@ -24,7 +24,7 @@ public class LabService {
     static <T> T convertJsonToObject(final String json, final Class<T> clazz) {
         try {
             return mapper.readValue(json, clazz);
-        } catch (JsonProcessingException e) {
+        } catch (IOException e) {
             throw new RuntimeException("JsonProcessingException: " + e.getMessage(), e);
         }
     }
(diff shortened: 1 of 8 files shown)


src/integrationtest/java/labapi/LabService.java
```diff
@@ -24,7 +24,7 @@
     static <T> T convertJsonToObject(final String json, final Class<T> clazz) {
         try {
             return mapper.readValue(json, clazz);
-        } catch (JsonProcessingException e) {
+        } catch (IOException e) {
             throw new RuntimeException("JsonProcessingException: " + e.getMessage(), e);
```
henrik-me, line 28: >JsonProcessingException [](start = 40, length = 23)

nit: consider changing this?

src/main/java/com/microsoft/aad/msal4j/TokenResponse.java
```diff
@@ -52,28 +52,28 @@ static TokenResponse parseJsonObject(final JSONObject jsonObject)
...
         String clientInfo = null;
         if (jsonObject.containsKey("client_info")) {
-            clientInfo = jsonObject.getAsString("client_info");
+            clientInfo = JSONObjectUtils.getString(jsonObject, "client_info");
         }
 
         long expiresIn = 0;
         if (jsonObject.containsKey("expires_in")) {
-            expiresIn = jsonObject.getAsNumber("expires_in").longValue();
+            expiresIn = Long.parseLong(jsonObject.getAsString("expires_in"));
```
henrik-me, line 71: >Long.parseLong( [](start = 24, length = 15)

should this have been a mthod in the JSONObjectUtils instead? to align behavior and parsing logic in one place?

SomkaPe: yes, but that method expects json value as JSON number, so fails if it is string. AAD returns those values as JSON number, B2C as JSON String,  so can not rely on JSONObjectUtils

src/integrationtest/java/labapi/LabService.java
SomkaPe: good point - going to change message

Issue #247, opened 2020-06-09
Tenant profile support in MSAL Java


navyasric:

jan-tosovsky-cz:
When creating daemon app which should send emails by my name (graphClient.me().sendMail()), I assume the only possibility to authenticate as a specific user is to get AuthToken using UserNamePassword approach. However, this "flow" can only be used in PublicClientApplication. When using this way:
```java
public static IAuthenticationResult getUserPasswordAccessToken() throws Exception {

    PublicClientApplication app = PublicClientApplication.builder(CLIENT_ID).build();

    UserNamePasswordParameters userNamePasswordParams = UserNamePasswordParameters.builder(
        Collections.singleton("Send.Mail"), "email", "password".toCharArray())
        .build();

    CompletableFuture<IAuthenticationResult> future = app.acquireToken(userNamePasswordParams);

    return future.get();
}
```
But I am getting this error the Tenant has to be specified:
```
com.microsoft.aad.msal4j.MsalServiceException: AADSTS9001023: The grant type is not supported 
over the /common or /consumers endpoints. Please use the /organizations or tenant-specific endpoint.
```
I can't see any way how to specify Tenant, but looking into C# sample app I can see `PublicClientApplicationBuilder.CreateWithApplicationOptions(appConfig)` method which can be used for this.
https://github.com/Azure-Samples/active-directory-dotnetcore-console-up-v2/blob/master/up-console/Program.cs

Is this enhancement about the same thing or should I create separate feature request?

PR #263, opened 2020-07-21
Tenant Profiles

Avery-Dunn:
Adds support for tenant profiles, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/247 and this spec: https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/760?_a=files&path=/AccountDomainModel/Overview.md

Overall, it implements the tenant profile account type. Tenant profiles are representations of an account in tenants other than the account's home tenant, and are essentially accounts where the local UID is different from the home UID. Knowledge of tenant profiles and their functionality is not needed for clients which only ever access a single tenant, so tenant profiles are not exposed unless a client is dealing with accounts represented in multiple tenants.

Main changes:
- New interfaces/classes that extend IAccount, for tenant profiles and accounts that have tenant profiles
- IAccount interface now has a map for claims taken from the ID token
- getAccounts() can now return several IAccount implementations: Account objects (no change from before), MultiTenantAccount objects (Accounts with tenant profiles attached), and TenantProfile objects with no corresponding home account in the cache
- Unit tests for the above


src/main/java/com/microsoft/aad/msal4j/TenantProfile.java
```diff
@@ -0,0 +1,20 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import lombok.experimental.Accessors;
+
+import java.util.Map;
+
+@Accessors(fluent = true)
+@Getter
+@Setter
+public class TenantProfile extends Account implements ITenantProfile{
```
SomkaPe, line 12: Why tenant Profile extends Account ?

src/main/java/com/microsoft/aad/msal4j/MultiTenantAccount.java
```diff
@@ -0,0 +1,22 @@
...
+import lombok.Getter;
+import lombok.Setter;
+
+import java.util.Map;
+
+@Getter
+@Setter
+public class MultiTenantAccount extends Account implements IMultiTenantAccount {
+
+    private Map<String, ITenantProfile> tenantProfiles;
```
SomkaPe, line 12: as far As I see tenantProfiles are never set - so always null

Avery-Dunn: They are set in TokenCache (line 345), though it uses the setter generated by Lombok. I had an explicit getTenantProfiles() just because it was in the spec, but looking at it now I think it'd be better to have either an explicit getter/setter and not use Lombok here, or to just let both be handled by Lombok. Do you have any preference?

src/main/java/com/microsoft/aad/msal4j/TenantProfile.java
Avery-Dunn: A tenant profile is pretty much the same as an account, except that we need to know the local UID in addition to the home UID, and I based the extend/implement structure of the new classes off of the  implementation in Android.

src/main/java/com/microsoft/aad/msal4j/Account.java
```diff
@@ -23,4 +24,22 @@
...
+
+    public String getTenantId() {
+        return localAccountId;
+    }
+
+    public Map<String, ?> getClaims() {
+        return idTokenClaims;
+    }
+
+    public Map<String, IAccount> getTenantProfiles() {
```
SomkaPe, line 42: TenantProfile is not Account, relation between tenantProfile and account is many to one

src/main/java/com/microsoft/aad/msal4j/AccountCacheEntity.java
```diff
@@ -77,6 +80,7 @@ static AccountCacheEntity create(String clientInfoStr, Authority requestAuthorit
             account.localAccountId(localAccountId);
             account.username(idToken.preferredUsername);
             account.name(idToken.name);
+            account.idTokenClaims(idToken.tokenClaims());
```
SomkaPe, line 83: token claims should not be part of the key

src/main/java/com/microsoft/aad/msal4j/AccountCacheEntity.java
```diff
@@ -49,6 +50,8 @@ ClientInfo clientInfo() {
     @JsonProperty("authority_type")
     protected String authorityType;
 
+    protected Map<String, ?> idTokenClaims;
```
SomkaPe, line 53: how this will be populated ?

src/main/java/com/microsoft/aad/msal4j/Account.java
Avery-Dunn: The implementation in the Android library seems to suggest that they are accounts, or at least can be treated like accounts without issue.

If you look at Android's [AccountAdapter.java](https://github.com/AzureAD/microsoft-authentication-library-for-android/blob/dev/msal/src/main/java/com/microsoft/identity/client/AccountAdapter.java), you can see the only place (as far as I can tell) tenant profiles are directly created. There is an adapt() method with is given a list of cache records, for example in MultipleAccountPublicClientApplications.getAccount(), which appears to split a list of ICacheRecord into accounts and tenant profiles solely by comparing the local and home ID values from ICacheRecord.getAccount() , which returns AccountRecord objects. If the AccountRecord's home ID contains the local ID it's either an Account or MultiTenantAccount, and if not it's a tenant profile.

So, if the data in AccountRecords (which seem to be Android's version of Java's AccountCacheEntity) is the only thing that determines what is an Account and what is a TenantProfile, and since we decided not to add all of the extra interfaces/classes that Android and iOS implemented, then isn't a TenantProfile essentially an Account?

src/main/java/com/microsoft/aad/msal4j/AccountCacheEntity.java
Avery-Dunn: It is populated in the create() method, and the Account objects get it through the toAccount() method. Is there a situation where an AccountCacheEntity isn't created for an account, or a situation where we return an Account object that doesn't come from the AccountCacheEntity?

However, you also say that token claims "shouldn't be part of the key", but if the Account class now has token claims as a field, shouldn't the cache entity match? If not, where would the Account/IdToken relation come from? As far as I can tell, the IdToken object is only set in TokenRequestExecutor.createAuthenticationResultFromOauthHttpResponse(), which is also where AccountCacheEntities are created.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -300,16 +302,92 @@ static private AppMetadataCacheEntity createAppMetadataCacheEntity(TokenRequestE
...
-                        collect(Collectors.mapping(AccountCacheEntity::toAccount, Collectors.toSet()));
+                        filter(acc -> acc.homeAccountId().contains(acc.localAccountId())).
+                        collect(Collectors.toSet());
+                Set<AccountCacheEntity> tenantAccounts = accounts.values().stream().
+                        filter(acc -> environmentAliases.contains(acc.environment())).
+                        filter(acc -> !acc.homeAccountId().contains(acc.localAccountId())).
+                        collect(Collectors.toSet());
+
+                for (AccountCacheEntity homeAccCached : homeAccounts) {
+                    IdTokenCacheEntity homeAcctIdToken = idTokens.get(getIdTokenKey(
```
SomkaPe, line 319: should not normally happen , but it is good to make sure it exist and not null

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -300,16 +302,92 @@ static private AppMetadataCacheEntity createAppMetadataCacheEntity(TokenRequestE
...
+
+                    if (tenantProfiles.size() > 0) {
+                        homeAcc.tenantProfiles(tenantProfiles);
+                    }
+
+                    rootAccounts.add(homeAcc);
+                }
+                //Create Accounts from any tenant profile which did not have a corresponding home account,
+                // and add it to the list of root accounts
+                for (AccountCacheEntity tenantProfileNoHome : tenantAccounts) {
```
SomkaPe, line 358: why do you call accounts tenantProfile?

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -300,16 +302,92 @@ static private AppMetadataCacheEntity createAppMetadataCacheEntity(TokenRequestE
...
+                            tenantProfileNoHome.homeAccountId(),
+                            tenantProfileNoHome.environment(),
+                            clientId,
+                            tenantProfileNoHome.realm()));
+                    Account account = (Account) tenantProfileNoHome.toAccount();
+                    account.idTokenClaims(JWTParser.parse(token.secret).getJWTClaimsSet().getClaims());
+                    rootAccounts.add(account);
+                }
+
+                return rootAccounts;
```
SomkaPe, line 369: the logic of aggregation seem a bit complicated, why would not you just fill 
map <homeAccountId, Set<AccountCacheEntity> > than iterate it and populate tenantProfiles ?
that would be easier to read + complexity will be O(N) instead of O(N pow 2)

src/main/java/com/microsoft/aad/msal4j/Account.java
```diff
@@ -23,4 +24,18 @@
...
 
     String username;
+
+    String localAccountId;
+
+    Map<String, ?> idTokenClaims;
+
+    Map<String, ITenantProfile> tenantProfiles;
+
+    public Map<String, ?> getClaims() {
```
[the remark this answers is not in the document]
Avery-Dunn: This map is essentially the same claims map in a JWT, basically the same stuff defined in the IdToken class (things like issuer, subject, audience, etc.). I'll add some Java docs, but it definitely can't just be a set.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -300,16 +302,92 @@ static private AppMetadataCacheEntity createAppMetadataCacheEntity(TokenRequestE
...
+
+                for (AccountCacheEntity homeAccCached : homeAccounts) {
+                    IdTokenCacheEntity homeAcctIdToken = idTokens.get(getIdTokenKey(
+                            homeAccCached.homeAccountId(),
+                            homeAccCached.environment(),
+                            clientId,
+                            homeAccCached.realm()));
+
+                    Account homeAcc = (Account) homeAccCached.toAccount();
+                    homeAcc.idTokenClaims = JWTParser.parse(homeAcctIdToken.secret).getJWTClaimsSet().getClaims();
```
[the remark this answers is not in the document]
Avery-Dunn: According to the docs for JWTParser, the [getClaims() method](javadoc.io/doc/com.nimbusds/nimbus-jose-jwt/5.4/com/nimbusds/jwt/JWTClaimsSet.html#getClaims--) returns both standard and custom claims, so this should add every claim in the token to the idTokenClaims map.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -300,16 +302,92 @@ static private AppMetadataCacheEntity createAppMetadataCacheEntity(TokenRequestE
...
                 lock.readLock().lock();
+                Set<IAccount> rootAccounts = new HashSet<>();
 
-                return accounts.values().stream().
+                //Filter accounts map into two sets: a set of home accounts (local OID = home OID),
+                // and a set of tenant profiles (local OID != home OID)
+                Set<AccountCacheEntity> homeAccounts = accounts.values().stream().
                         filter(acc -> environmentAliases.contains(acc.environment())).
-                        collect(Collectors.mapping(AccountCacheEntity::toAccount, Collectors.toSet()));
+                        filter(acc -> acc.homeAccountId().contains(acc.localAccountId())).
```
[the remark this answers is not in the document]
Avery-Dunn: homeAccountID is in the format of objectID.tenantID, but the localAccountID only corresponds to the objectID,: for an account the local object ID matches the one in the homeAccountID, and for a tenant profile does not. I could've done something like `localAccountId.equals(homeAccountId.split("\\.")[0])`, but I figured this would be more readable.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
Avery-Dunn: We need the 'realm' value of an AccountCacheEntity to know which ID tokens are relevant to an Account, and this value is not added to the Account class. Would it be alright to just add a new `realm` field to the Account class? If so, then I could turn all of the home AccountCacheEntities to Accounts, and just iterate over the tenant profiles like you suggest.

src/main/java/com/microsoft/aad/msal4j/IAccount.java
```diff
@@ -26,4 +26,14 @@
      * @return account username
      */
     String username();
+
+    /**
+     * @return claims in id token
```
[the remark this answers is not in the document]
Avery-Dunn: Would a single account have multiple cached ID tokens? I'd assume either the relationship is 1:1 or the ID tokens don't have conflicting information. If they do have conflicting information, how do we determine which is the correct values?

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
```diff
@@ -0,0 +1,30 @@
...
+ */
+public interface ITenantProfile {
+
+    /**
+     * @return local OID
+     */
+    String getId();
+
+    /**
+     * @return local tenant ID
```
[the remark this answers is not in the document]
Avery-Dunn: It is the same as the `localAccountId` field in AccountCacheEntity, and is taken from the `oid` key of an ID token. It's used to determine if an AccountCacheEntity is a home account or a tenant profile (if the localAccountId/oid doesn't match the uid value in the homeAccountId, then it's a tenant profile).

I'll add some more info to that doc.

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
```diff
@@ -0,0 +1,30 @@
...
+ * Interface representing a single tenant profile. ITenantProfiles are made available through the
+ * {@link IAccount#getTenantProfiles()} method of an Account
+ *
+ */
+public interface ITenantProfile {
+
+    /**
+     * @return local OID
+     */
+    String getId();
```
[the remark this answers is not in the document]
Avery-Dunn: Both getId() and getClaims() are declared as public in the design doc, and the Android implementation makes them public as well, so I think it's meant to be exposed. Since this value is available from getClaims() anyway we could probably keep it private without changing the actual functionality, but leaving it exposed keeps it consisted with other SDKs/the design.

src/main/java/com/microsoft/aad/msal4j/TenantProfile.java
```diff
@@ -0,0 +1,37 @@
...
+@AllArgsConstructor
+class TenantProfile implements ITenantProfile {
+    String objectID;
+
+    String tenantID;
+
+    Map<String, ?> idTokenClaims;
+
+    public String getId() {
+        return objectID;
```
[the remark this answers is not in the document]
Avery-Dunn: I originally named the field to be consistent with the name from the ID token (oid -> ObjectIdentifier -> objectId), and later made the getter to match how it's described in the design doc and how it was implemented in the Android/iOS libraries, and didn't think about that inconsistency.

I'll changed 'objectId' to just 'Id' in the next commit with other fixes.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
Avery-Dunn: Good call, I'll add this check in with the next commit with your other recommendations.

src/main/java/com/microsoft/aad/msal4j/TokenCache.java
Avery-Dunn: If an AccountCacheEntity is a tenant profile, it's provided to the user by getting attached to a home Account. However, if there is not home account to attach to, then getAccounts() will not be returned them unless they're returned as their own objects. Turning them into Accounts seemed like the correct way to do this, because:
- These AccountCacheEntities would've been returned them as Accounts before this change anyway
- We agreed that TenantProfile should not extend Account, so if I returned them as TenantProfiles I would've had to change the return type of getAccounts(), which would've definitely been a breaking change
- The design doc doesn't really cover this, but this is the behavior done in the Android/iOS libraries, so worst-case at least we're consistent.

src/main/java/com/microsoft/aad/msal4j/Account.java
```diff
@@ -23,4 +24,18 @@
     String environment;
 
     String username;
+
+    String localAccountId;
```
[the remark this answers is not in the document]
Avery-Dunn: I just overlooked the automatic getter/setter annotations, I'll make this private in the commit of fixes.

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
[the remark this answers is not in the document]
SomkaPe: yes, it is available from claims - why this particular claim should be duplicated?

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
```diff
@@ -0,0 +1,34 @@
...
+     * This value corresponds to the 'oid' key of an ID token
+     *
+     * @return String local OID
+     */
+    String getId();
+
+    /**
+     * This value corresponds to the 'realm' key of an ID token
+     *
+     * @return String local tenant ID
```
SomkaPe, line 25: not sure we should use "local" here

src/main/java/com/microsoft/aad/msal4j/IdToken.java
```diff
@@ -59,6 +61,29 @@
     @JsonProperty("unique_name")
     protected String uniqueName;
 
+    /**
+     * Used to attach all claims in an ID token to an account object
+     *
+     * @return map of key/value pairs of claims in ID token
+     */
+    protected Map<String, Object> tokenClaims() {
```
SomkaPe, line 69: is it never used now ?

src/main/java/com/microsoft/aad/msal4j/Account.java
```diff
@@ -23,4 +24,12 @@
     String environment;
 
     String username;
+
+    private String localAccountId;
```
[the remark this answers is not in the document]
Avery-Dunn: After the last redesign it's no longer needed, will remove in the next commit.

src/main/java/com/microsoft/aad/msal4j/IdToken.java
Avery-Dunn: Nope, will remove.

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
Avery-Dunn: Good catch, now that we don't have the home/non-home tenants distinction there's no need to refer to these sorts of things as 'local'. Will remove in the next commit.

src/main/java/com/microsoft/aad/msal4j/IAccount.java
```diff
@@ -26,4 +26,9 @@
      * @return account username
      */
     String username();
+
+    /**
+     * @return tenant profiles
```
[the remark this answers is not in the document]
Avery-Dunn: In the latest commit I just add some info explaining it follows the key/value structure of a JWT.

src/main/java/com/microsoft/aad/msal4j/IAccount.java
```diff
@@ -26,4 +26,12 @@
      * @return account username
      */
     String username();
+
+    /**
+     * Map of {@link ITenantProfile} objects related to this account. The keys of the map are tenant ID values, which
+     * are taken from {@link AccountCacheEntity#realm}
```
SomkaPe, line 32: AccountCacheEntity is not public  class - should not be referenced in java doc

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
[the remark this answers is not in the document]
SomkaPe: please be more specific - is "oid" claim value from id token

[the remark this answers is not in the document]
SomkaPe: please use consistent naming - here getter , name of field , description  and description in "return" parts are all different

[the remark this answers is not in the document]
SomkaPe: I am still not sure why we need to expose it here - why int can not be read from claims map direcly

src/main/java/com/microsoft/aad/msal4j/IAccount.java
Avery-Dunn: Fixed in the latest commit

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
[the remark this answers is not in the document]
Avery-Dunn: Per our messages on teams, I've removed the getID() and getTenantID() related stuff from TenantProfile, and now it only has getClaims().

Issue #257, opened 2020-06-30
[Enhancement] Set meaningful defaults for MSAL http client


navyasric:
This addresses issues in Azure SDK's use of MSAL

henrik-me:
This was discussed to ensure we provide a set of meaningful defaults for services. The settings is overridable by any application. Today we leave the defaults of the system but are exploring setting values ensuring time-outs will not be indefinitely etc.

PR #264, opened 2020-07-31
HTTPClient default timeouts

Avery-Dunn:
Adds some default timeouts for the library's default HTTPClient, per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/257

Also fixes an issue discovered in a DeviceCodeIT test: there is now a 'Stay signed in?' confirmation page in the device code flow, which caused the test to hang until the device code became invalid. An extra Selenium step was added to handle this confirmation page.

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
@@ -46,7 +46,7 @@ public class DeviceCodeIT {
         build();
 
         Consumer<DeviceCode> deviceCodeConsumer = (DeviceCode deviceCode) -> {
-            runAutomatedDeviceCodeFlow(deviceCode, user);
+            runAutomatedDeviceCodeFlow(deviceCode, user, environment);
         };
 
         IAuthenticationResult result = pca.acquireToken(DeviceCodeFlowParameters
@@ -70,7 +70,7 @@ public class DeviceCodeIT {
                 build();
 
         Consumer<DeviceCode> deviceCodeConsumer = (DeviceCode deviceCode) -> {
-            runAutomatedDeviceCodeFlow(deviceCode, user);
+            runAutomatedDeviceCodeFlow(deviceCode, user, environment);
         };
 
         IAuthenticationResult result = pca.acquireToken(DeviceCodeFlowParameters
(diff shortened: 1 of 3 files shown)


src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
```diff
@@ -123,6 +123,15 @@ private void runAutomatedDeviceCodeFlow(DeviceCode deviceCode, User user){
             } else {
                 SeleniumExtensions.performADLogin(seleniumDriver, user);
             }
+
+            if (environment.equals("azurecloud") && !isADFS2019) {
```
sangonzal, line 127: Consider adding to constants file

src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -15,6 +15,8 @@
 
     private final Proxy proxy;
     private final SSLSocketFactory sslSocketFactory;
+    public int defaultConnectTimeout = 3000;
```
SomkaPe, line 18: Upper case should be used for Constants

src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -46,6 +50,8 @@ private HttpResponse executeHttpPost(HttpRequest httpRequest) throws Exception {
         configureAdditionalHeaders(conn, httpRequest);
         conn.setRequestMethod("POST");
         conn.setDoOutput(true);
+        conn.setConnectTimeout(defaultConnectTimeout);
+        conn.setReadTimeout(defaultReadTimeout);
```
SomkaPe, line 54: might be better to set timeouts in openConnection function

src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
Avery-Dunn: In the latest commit, this value now comes from the constants in AzureEnvironment.java

src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
Avery-Dunn: It's now in the openConnection() function, and the two constants are all upper-case.

Issue #219, opened 2020-04-22
[Feature] SubjectName/Issuer (SNI) authentication. WithSendx5c()


henrik-me:
**Supported in:**
[Python](https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/60)
[.NET](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1490)

Background of SNI: How does it work?
-------------------------------------------

The following description is derived from Matt Bearup, a software developer from Microsoft.

1. User update AAD tenant to accept approved issuers.
   * The AAD Tenant only accepts certain issuers (AME) which have strict association of subject name and requestor
2. User associates an issuer/subject name with their service principal.
   * Once I as the requestor register contoso.com with AME CA, no one else can get it.
   * This is the foundation of SN/Issuer trust. Rather than trusting the thumbprint, you’re trusting that the CA vouches for me as the only one who can generate certs for that subject name.
3. User configures app to use SNIssuer auth instead of thumbprint-based.
   * We don’t need to deprecate thumbprint auth or switch back and forth. The client would simply use a different login option going forward, e.g. `az login --newoption /path/tocert`. MSAL needs to help enable this.
   * As the requester/user it’s still my responsibility to get the latest pfx from keyvault/dsms/etc. An expired cert would still be rejected by AAD.
   * What this saves us is the trouble of registering a new thumbprint with AAD, which would create a more complex autorotation scenario (use old cert to register new cert, then use new cert to deregister old cert).
   * Without SN/Issuer auth there is also a gap in authentication when key material is rotated. When KeyVault issues a new cert, its thumbprint is not in AAD. My app would have to recognize this, rotate thumbprints in AAD, and try again.
   * Note that the key material does get rotated, but that’s not what is being trusted. Instead we’re trusting that an approved CA issued a cert with an approved subject name and that we have the matching private key for that cert.
   * With SN/Issuer auth the *only* thing I have to worry about is getting the latest pfx for my app.

How does it need to be implemented in MSAL?
----------------------------------------------------
* What the “x5c” JWT header value would be? It appears only public certificate is sufficient. MSAL do NOT need to, and MUST NOT send private cerficate.
Sagar: Correct, the x5c is the public cert.

* With “x5c” available, why “x5t” is still required, since it can be generated from “x5c” through sha1 hash easily?
[...]

PR #265, opened 2020-08-04
Allow support for a chained certificate / pkcs12 file off the bat

thebenwaters:
Trying to tackle #219 

It seems via the code that SNI is "supported by default"

https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java#L52

This goes ahead and sets multiple certificates to the x5c header and it's also already using x5t

The issue seems that when supporting a chained certificate, it's currently hardcoded to only accept one certificate and not a chain.

I'm not a Java guy and this is not pretty code and a decent amount of stackoverflow/googling

Please help critique the code so it's up to standards, but the goal of the PR is to support a list of x509 certificates instead of a single cert

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/ClientCertificate.java
+++ b/src/main/java/com/microsoft/aad/msal4j/ClientCertificate.java
@@ -104,6 +104,7 @@ final class ClientCertificate implements IClientCertificate {
             String alias = aliases.nextElement();
             if (keystore.isKeyEntry(alias)) {
                 key = (PrivateKey) keystore.getKey(alias, password.toCharArray());
+                //dig down the certificate chain and put the public keys in the keystore
                 Certificate[] chain = keystore.getCertificateChain(alias);
                 for (Certificate c: chain) {
                     publicCertificates.add((X509Certificate)c);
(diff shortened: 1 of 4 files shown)


src/main/java/com/microsoft/aad/msal4j/ClientCertificate.java
```diff
@@ -86,16 +98,25 @@ static ClientCertificate create(final InputStream pkcs12Certificate, final Strin
...
-        final X509Certificate publicCertificate = (X509Certificate) keystore
-                .getCertificate(alias);
-        return create(key, publicCertificate);
+        final ArrayList<X509Certificate> publicCertificates = new ArrayList<X509Certificate>();
+        PrivateKey key = null;
+        while (aliases.hasMoreElements()) {
+            String alias = aliases.nextElement();
+            if (keystore.isKeyEntry(alias)) {
+                key = (PrivateKey) keystore.getKey(alias, password.toCharArray());
+                Certificate[] chain = keystore.getCertificateChain(alias);
```
thebenwaters, line 107: Go through the chain and add the public certs to the keystore

SomkaPe:
@thebenwaters was you able to E2E test this ?

thebenwaters:
@SomkaPe working on resolving backend issue. on paper it should work

thebenwaters:
@SomkaPe I asked the owner of the app 55e7e5af-ca53-482d-9aa3-5cb1cc8eecb5 for access to configure SNI so we can have full integration testing

thebenwaters:
This has been accepted and merged. Closing

PR #266, opened 2020-08-04
Pesomka/extra scopes

SomkaPe:
Extra scope to consent during authorization request


SomkaPe:
good point about order of scopes, but it should not matter because it is used just for authorization request, during token request (with auth code) scope for access token is specified

Issue #273, opened 2020-08-28
V 1.6.2 failed with proxy configured


ravindra12jan:
Hi,

We were using 1.4.0 msal4j version and now we have upgraded it to latest version 1.6.2. But now we get this exception while fetching the toke. Proxy is configured in our environment.

`com.microsoft.aad.msal4j.MsalClientException: java.net.SocketTimeoutException: Read timed outcom.microsoft.aad.msal4j.MsalClientException: java.net.SocketTimeoutException: Read timed out
	at com.microsoft.aad.msal4j.HttpHelper.executeHttpRequest(HttpHelper.java:68)
	at com.microsoft.aad.msal4j.OAuthHttpRequest.send(OAuthHttpRequest.java:34)
	at com.microsoft.aad.msal4j.TokenRequestExecutor.executeTokenRequest(TokenRequestExecutor.java:37)
	at com.microsoft.aad.msal4j.AbstractClientApplicationBase.acquireTokenCommon(AbstractClientApplicationBase.java:117)
	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.execute(AcquireTokenByAuthorizationGrantSupplier.java:63)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:59)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:17)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1604)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624)
	at java.lang.Thread.run(Thread.java:748)
Caused by: java.net.SocketTimeoutException: Read timed out
	at java.net.SocketInputStream.socketRead0(Native Method)
	at java.net.SocketInputStream.socketRead(SocketInputStream.java:116)
	at java.net.SocketInputStream.read(SocketInputStream.java:171)
`

We are using Client Credential flow to get the token. Exactly same as - https://github.com/Azure-Samples/ms-identity-java-daemon/tree/master/msal-client-credential-secret.

Additional thing what we did is whitelisted below urls on the proxy. We are using Squid proxy.

` .login.microsoftonline.com,
 .dataservice.protection.outlook.com,
[...]

PR #275, opened 2020-08-31
Improve HTTP client timeouts

Avery-Dunn:
Increases the default timeouts for connections made through DefaultHttpClient, and allows the user to change those timeout values when creating a PublicClientApplication

Should fix the sort of issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/273 in most cases, and allow users to have more control of timeouts if they need.


src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -446,6 +456,34 @@ public T sslSocketFactory(SSLSocketFactory val) {
...
         }
 
+        /**
+         * Sets the connect timeout value used in HttpsURLConnection connections made by {@link DefaultHttpClient},
+         * and is not needed if using a custom HTTP client
+         *
+         * @param val timeout value in milliseconds
+         * @return instance of the Builder on which method was called
+         */
+        public T httpClientConnectTimeout(Integer val) {
```
SomkaPe, line 466: name should highlight that this settings used only for default http client

Avery-Dunn: I originally called it it 'defaultHttpClientConnectTimeout' (same for the read timeout), but thought naming a settable property 'default' would be confusing. However, I now realize that a user will just think these timeouts work with the httpClient() part of the builder.

connectTimeoutForDefaultHttpClient maybe?

Issue #281, opened 2020-09-09
Incorrect parsing of ClientInfo when trying to authorize on some old accounts


RomanNosachev:
While testing our application using the MSAL4j (authorization by code), we identified the following problem: some clients could not authorize, because when receiving a response from the server, they received a response parsing error:

`java.lang.IllegalArgumentException: Illegal base64 character 2d`

After debugging, I managed to find a problem: in the ClientInfo class, when parsing a response from JSON, Base64 encoded characters are expected, but in practice, responses arrive that may contain Base64url encoded characters, such as '_' and '-'.
More about encoding differences: [RFC 4648 ](https://tools.ietf.org/html/rfc4648#page-7)
I corrected the code in the fork by replacing `Base64.getDecoder` with `Base64.getUrlDecoder` and our clients were able to log in 
and user Microsoft Graph API.
We hope that you will correct this error so that we can use your version of the library again.

Link to pull request: #282

Best regards, Roman Nosachev, middle Kotlin developer at Getapy

SomkaPe:
@RomanNosachev Thank you for reporting this issue and providing PR with fix. Could you please provide more details about accounts for which you got this problem ? Was it federated account ?

navyasric:
PR is merged.

PR #282, opened 2020-09-09
Fixed parsing ClientInfo: added support for Base64URL encoding characters

RomanNosachev:
Fixed parsing ClientInfo: on some accounts, the server response contained characters that are incorrect for Base64 encoding 
('_', '-'), but acceptable for Base64URL

This case was tested in our project Getapy and after the patch everything worked correctly as expected.
Otherwise our clients were getting errors like this: `java.lang.IllegalArgumentException: Illegal base64 character 2d`

Link to issue: #281


SomkaPe:
@RomanNosachev Thank you for contribution, fix will be included into next release

PR #293, opened 2020-09-29
Add integration tests for acquire token silent with B2C and ADFS 2019 accounts

Avery-Dunn:
Added test coverage for acquireTokenSilent flow for B2C and ADFS2019, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/272. Additionally, refactored AcquireTokenSilentIT by extracting duplicate code into some common methods.


sangonzal (approved):
Thanks for adding these!

Issue #292, opened 2020-09-28
Serialization failed for IAuthenticationResult


abhijeetsaxena7:
I am trying to use spring jdbc session and I want to store the IAuthenticationResult object returned on calling ClientConfidentialParamter.acquireToken() method. However, I keep getting the error that IAuthenticationResult cannot be serialized. I looked into it and realized that IAuthenticationResult is an interface. AuthenticationResult is the subclass but I cannot use it as its visibility is default in the package and I dont have access to it. Is there anyother way to serialize/deserialize IAuthenticationResult data?

Avery-Dunn:
Hello @abhijeetsaxena7 : Unfortunately, I don't think there's a quick way to serialize IAuthenticationResult/AuthenticationResult objects themselves. We plan on discussing this issue, and make serialization of these objects more straightforward in a future update.

For now though, the best workaround I can suggest is to extract the data you need from the IAuthenticationResult into a custom object, and serialize that.

idg-sam:
Maybe we should add a demonstration in our samples for this.

abhijeetsaxena7:
> Hello @abhijeetsaxena7 : Unfortunately, I don't think there's a quick way to serialize IAuthenticationResult/AuthenticationResult objects themselves. We plan on discussing this issue, and make serialization of these objects more straightforward in a future update.
> 
> For now though, the best workaround I can suggest is to extract the data you need from the IAuthenticationResult into a custom object, and serialize that.

I create my own class for and stored tokenCache and account details and then serialized that class.

Avery-Dunn:
Hello @abhijeetsaxena7 : Glad you were able to work around this issue. We plan on making IAccount serializable in the next update, to help with things like setting up .acquireTokenSilently() calls and searching through the results of .getAccounts(). 

However, we don't plan on making IAuthenticationResult serializable, as we don't think there is a good use-case for that: all of the token info they contain can be found in the token cache (which is serializable), and once IAccount is serializable there'll be no reason to store the IAuthenticationResult instead of the IAccount and the token cache.

Could you share a bit of your use case here? Was there some specific need/convenience to store IAuthenticationResult objects in session rather than the token cache, and would the ability to serialize/store the IAccount from an auth result along with the token cache solve that issue?

abhijeetsaxena7:
[...]

PR #297, opened 2020-10-07
Make IAccount serializable

Avery-Dunn:
As per discussions regarding https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/292, this change makes IAccount serializable, and removes the potential implicit serialization of IAuthenticationResult by making AuthenticationResult not serializable.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
+++ b/src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
@@ -3,6 +3,7 @@
 
 package com.microsoft.aad.msal4j;
 
+import java.io.Serializable;
 import java.util.Map;
 
 /**
@@ -10,7 +11,7 @@ import java.util.Map;
  * {@link IAccount#getTenantProfiles()} method of an Account
  *
  */
-public interface ITenantProfile {
+public interface ITenantProfile extends Serializable {
 
     /**
      * A map of claims taken from an ID token. Keys and values will follow the structure of a JSON Web Token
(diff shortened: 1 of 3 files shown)


sangonzal (approved):

src/main/java/com/microsoft/aad/msal4j/ITenantProfile.java
```diff
@@ -3,14 +3,15 @@
...
+import java.io.Serializable;
 import java.util.Map;
 
 /**
  * Interface representing a single tenant profile. ITenantProfiles are made available through the
  * {@link IAccount#getTenantProfiles()} method of an Account
  *
  */
-public interface ITenantProfile {
+public interface ITenantProfile extends Serializable {
```
SomkaPe, line 14: I have noticed that ? is used as value of claim, we probably should switch to String

Avery-Dunn: The JWT parser we use (from the Nimbus package) returns the claims as a `Map<String, Object>` , and I don't think there's an easy way to convert that to a  `Map<String, String>`  (other than going through each item and getting the String value). 

Because of that, I could make claims either a `Map<String, Object>` or a `Map<String, ?>` . A `Map<String, Object>` can only passed into methods and whatnot that explicitly have a parameter of `Map<String, Object>`, whereas I believe `Map<String, ?>` could be passed as a parameter of type `Map<String, String>` (or another String/object pair), which I figured would be more useful for users.

burhanelgun:
@Avery-Dunn why do we need serializing the IAccount? We can reach the account from the TokenCache(which is also serializable)?
For example like below:
```
IConfidentialClientApplication confidentialClientApplication = ConfidentialClientApplication.builder("sampleClientId", ClientCredentialFactory.createFromSecret("sampleClientSecret")).
                    authority("sampleAuthority").build();
 Set<IAccount> accounts = app.getAccounts().get();
```

Issue #258, opened 2020-07-01
Add public convenience string constants for endpoints of each cloud (sovereign and public)


henrik-me:
**Context**
Customers have had difficulty finding the right documentation pages:
https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-developer-guide#endpoint-mapping
https://docs.microsoft.com/en-us/azure/china/resources-developer-guide#check-endpoints-in-azure

**See also other libraries exposing these:**
- [obj-c](https://azuread.github.io/microsoft-authentication-library-for-objc/Enums/MSALAzureCloudInstance.html) 
- [.net](https://docs.microsoft.com/en-us/dotnet/api/microsoft.identity.client.azurecloudinstance?view=azure-dotnet)

**Reference:**
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1063014

**Other:**
https://github.com/AzureAD/microsoft-authentication-library-for-python/issues/221

Avery-Dunn:
Relevant PR is merged into dev branch: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/298

Avery-Dunn:
Released as part of version 1.8.0 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/314

PR #298, opened 2020-10-07
Add public constants for cloud endpoints

Avery-Dunn:
Adds public constants for national cloud endpoints (plus the public cloud), as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/258, for convenience when setting authorities.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
@@ -1,0 +1,26 @@
+package com.microsoft.aad.msal4j;
+
+public enum AzureCloudEndpoint {
+    /**
+     * Microsoft Azure public cloud, https://login.microsoftonline.com
+     */
+    AzurePublic("https://login.microsoftonline.com/"),
+    /**
+     * Microsoft Chinese national cloud, https://login.chinacloudapi.cn
+     */
+    AzureChina("https://login.chinacloudapi.cn/"),
+    /**
+     * Microsoft German national cloud ("Black Forest"), https://login.microsoftonline.de
+     */
+    AzureGermany("https://login.microsoftonline.de/"),
+    /**
+     * US Government cloud, https://login.microsoftonline.us
+     */
+    AzureUsGovernment("https://login.microsoftonline.us/");
+
+    public final String endpoint;
+
+    AzureCloudEndpoint(String endpoint) {
+        this.endpoint = endpoint;
+    }
+}
--- a/src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
@@ -3,6 +3,12 @@
 
 package com.microsoft.aad.msal4j;
 
+/**
+ * All the national clouds authenticate users separately in each environment and have separate authentication endpoints.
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
```diff
@@ -0,0 +1,26 @@
+package com.microsoft.aad.msal4j;
```
sangonzal, line 1: Please add license header

sangonzal (approved):
Would be good to update some of the samples, the wiki, and perhaps the Java doc for authority to reference this so people can find it.

src/main/java/com/microsoft/aad/msal4j/AzureCloudEndpoint.java
Avery-Dunn: Added the license header, some javadocs with a link to [our wiki page on the national cloud stuff](https://github.com/AzureAD/microsoft-authentication-library-for-java/wiki/National-Clouds), and add some info about the enum on that wiki page.

src/test/java/com/microsoft/aad/msal4j/AuthorityTest.java
```diff
@@ -174,4 +174,27 @@ public void testValidateAuthorityEmptyPathSegments(String authority) throws Malf
...
+        pca = PublicClientApplication.builder("client_id")
+                .authority(AzureCloudEndpoint.AzureGermany.endpoint + "common")
+                .build();
+        Assert.assertEquals(pca.authority(), TestConfiguration.AAD_COMMON_AUTHORITY_DE);
+
+        pca = PublicClientApplication.builder("client_id")
+                .authority(AzureCloudEndpoint.AzureUsGovernment.endpoint + "common")
+                .build();
+        Assert.assertEquals(pca.authority(), TestConfiguration.AAD_COMMON_AUTHORITY_US);
+    }
```
SomkaPe, line 199: I do not think this test bring any value.

src/test/java/com/microsoft/aad/msal4j/TestConfiguration.java
```diff
@@ -19,6 +19,9 @@
     public final static String AAD_TOKEN_SUCCESS_FILE = "/token.xml";
     public final static String AAD_DEFAULT_REDIRECT_URI = "https://non_existing_uri.windows.com/";
     public final static String AAD_COMMON_AUTHORITY = "https://login.microsoftonline.com/common/";
+    public final static String AAD_COMMON_AUTHORITY_CN = "https://login.chinacloudapi.cn/common/";
+    public final static String AAD_COMMON_AUTHORITY_DE = "https://login.microsoftonline.de/common/";
+    public final static String AAD_COMMON_AUTHORITY_US = "https://login.microsoftonline.us/common/";
```
SomkaPe, line 24: duplicating constants does not make sense

src/test/java/com/microsoft/aad/msal4j/AuthorityTest.java
Avery-Dunn: I based it off a test of setting default authority, but I guess for these endpoints just setting the authority is not a meaningful test. A good test would be acquiring a token at each endpoint, but that would mainly be a test of making sure the endpoint still works rather than if the library does, so I just removed the test and related config constants altogether.

Issue #296, opened 2020-10-06
Tenant Profile is not set in IAuthenticationResult


SomkaPe:

navyasric:
This is to address the issue https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/294

Avery-Dunn:
Fixed in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/300, and part of the 1.8 release https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/314

Issue #294, opened 2020-10-02
Expose ID Token Claims on successful Authentication


idg-sam:
Some other implementations of MSAL (e.g. Python) make available the decoded ID Token claims for the user immediately upon receiving the token. 
Currently, the Java implementation requires the user to import a JWT parsing library, such as this one from nimbusds:

```pom.xml
<dependency>
  <groupId>com.nimbusds</groupId>
  <artifactId>nimbus-jose-jwt</artifactId>
  <version>9.0.1</version>
</dependency>
```

And extract them as follows:

```Java
import com.nimbusds.jwt.JWTClaimsSet;
import com.nimbusds.jwt.SignedJWT;
...

Future<IAuthenticationResult> future = client.acquireToken(authParams);
IAuthenticationResult result = future.get();

SignedJWT idToken = SignedJWT.parse(result.idToken());
JWTClaimsSet jcs = idToken.getJWTClaimsSet();
Map<String,Object> claimsObjectMap = jcs.getClaimsSet();
Map<String,String> claimsStringMap = new HashMap<>();

claimsObjectMap.forEach((String claim, Object value) -> {
   String val = value.toString();
   claimsStringMap.put(claim, val);
});
```

We could implement this and expose the claims to the user as a non-breaking change in the library. MSAL4J already includes nimbusds dependency.

navyasric:
Now tracked in #296

PR #300, opened 2020-10-13
Add tenant profile/id claims to auth result

Avery-Dunn:
Adds the parsed claims from an ID token directly to the authentication result via the TenantProfile class, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/294 and https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/296


sangonzal (approved):
:shipit:

Issue #198, opened 2020-03-18
[Feature Request] refresh_in


henrik-me:
Refresh_In (token response) support. 
This feature allows the service to control when MSAL should attempt to refresh the access token (ahead of its expiration)

[API Review](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FRefreshAtExpirationPercentage%2Foverview.md&version=GBdev&_a=contents)

Avery-Dunn:
Implemented in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/305, and released as part of version 1.8 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/314

PR #305, opened 2020-10-23
Support for refresh_in

Avery-Dunn:
Adds support for the optional refresh_in field of a token response, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/198

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -231,7 +231,6 @@ public class AcquireTokenSilentIT {
                 build();
 
         IAuthenticationResult resultOriginal = acquireTokenUsernamePassword(user, pca, cfg.graphDefaultScope());
-
         assertResultNotNull(resultOriginal);
 
         IAuthenticationResult resultSilent = acquireTokenSilently(pca, resultOriginal.account(), cfg.graphDefaultScope(), false);
@@ -253,6 +252,7 @@ public class AcquireTokenSilentIT {
         IAuthenticationResult resultSilentWithRefreshOn = acquireTokenSilently(pca, resultOriginal.account(), cfg.graphDefaultScope(), false);
         //Current time is before refreshOn, so token should not have been refreshed
         Assert.assertNotNull(resultSilentWithRefreshOn);
+        Assert.assertEquals(pca.tokenCache.accessTokens.get(key).refreshOn(), Long.toString(currTimestampSec + 60));
         assertTokensAreEqual(resultSilent, resultSilentWithRefreshOn);
 
         token = pca.tokenCache.accessTokens.get(key);
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -10,6 +10,7 @@ import org.testng.annotations.Test;
 
 import java.net.MalformedURLException;
 import java.util.Collections;
+import java.util.Date;
 import java.util.Set;
 import java.util.concurrent.ExecutionException;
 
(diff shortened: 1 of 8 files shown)


src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -39,7 +41,12 @@ AuthenticationResult execute() throws Exception {
                 clientApplication.getServiceBundle().getServerSideTelemetry().incrementSilentSuccessfulCount();
             }
 
-            if (silentRequest.parameters().forceRefresh() || StringHelper.isBlank(res.accessToken())) {
+            //Determine if the current token needs to be refreshed according to the refresh_in value
+            long currTimeStampSec = new Date().getTime() / 1000;
+            boolean afterRefreshOn = res.refreshOn() != null && res.refreshOn() > 0 &&
```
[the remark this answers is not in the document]
Avery-Dunn: If `res` was null, then a null pointer would happen a few lines before when it checks `res.accessToken`(), and that check has been there since May. The method that makes the auth result (`TokenCache.getCachedAuthenticationResult()`) returns an auth result made from the builder, so even if nothing in the result is set it still doesn't return a null. Just to be safe though (in case some exception somewhere could lead to a null), I've copied the 'cache miss' null check/error message that's used before returning `res` to before `res` is first referenced.

src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -31,6 +31,8 @@
 
     private final String refreshToken;
 
+    private final Long refreshOn;
```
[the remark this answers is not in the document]
Avery-Dunn: I believe this field should be kept hidden. The design docs say  that 'no change to public API is required` and it's a feature that's ultimately controlled server-side (i.e., they determine if it appears in token responses), and talking with people who were part of the design discussions or worked on implementing it in .NET it seems like this is meant as an 'under the hood' feature that a user won't need to worry about.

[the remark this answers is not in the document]
SomkaPe: should not be exposed

src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -35,11 +37,20 @@ AuthenticationResult execute() throws Exception {
...
+
             if (!StringHelper.isBlank(res.accessToken())) {
                 clientApplication.getServiceBundle().getServerSideTelemetry().incrementSilentSuccessfulCount();
             }
 
-            if (silentRequest.parameters().forceRefresh() || StringHelper.isBlank(res.accessToken())) {
+            //Determine if the current token needs to be refreshed according to the refresh_in value
+            long currTimeStampSec = new Date().getTime() / 1000;
+            boolean afterRefreshOn = res.refreshOn() != null && res.refreshOn() > 0 &&
+                    res.refreshOn() < currTimeStampSec && res.expiresOn() >= currTimeStampSec;
```
SomkaPe, line 51: res.expiresOn() >= currTimeStampSec means you will not do refresh if AT is expired, why ?

Avery-Dunn: I didn't want to change any behavior if the token was expired, so if the token was after the `expiresOn` time then `refreshOn` time is irrelevant and we shouldn't do any of the `refreshOn` behavior. If the token was after the `refreshOn` time, then if any service needed to refresh it was down then the current token should be returned as though it nothing happened, which is not the (existing) behavior for an expired token.

src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
```diff
@@ -218,6 +219,52 @@ public void acquireTokenSilent_ConfidentialClient_acquireTokenSilentDifferentSco
                 .get();
     }
 
+    @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
```
SomkaPe, line 222: I think it would be better to have use unit test to test this scenario, current test does not cover end to end scenario , also it is hybrid of unit test + integration testing , i would keep them separated

Avery-Dunn: Without a way to get `refresh_in` in an actual token response, I'm not sure if there's a way to have meaningful tests without a hybrid like this. A unit test of setting the `refreshOn` field would just be 'set `refreshOn` manually, then assert that it was set'. The only time it's used is halfway through the acquireTokenSilent flow, and a unit test of that would need a lot of hardcoding that the actual token request process takes care of anyway.

sangonzal (approved):

PR #312, opened 2020-10-28
refresh on is optional field

SomkaPe:


sangonzal (approved):

Avery-Dunn (approved):
Now that refresh on is optional, the test `AcquireTokenSilentIT.acquireTokenSilent_WithRefreshOn()` will fail due to an assert that `token.refreshOn()` is expected to be the old default of 0

PR #313, opened 2020-10-29
1.8.0 Release

Avery-Dunn:


sangonzal (approved):
Sample MSAL versions should be updated as well.

Issue #307, opened 2020-10-23
Support for claims request parameter 


navyasric:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/824816

Avery-Dunn:
Supported as of the 1.8.1 release, https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/327

PR #315, opened 2020-11-10
Support for claims request parameter

Avery-Dunn:
Adds support for setting the OIDC claims request parameter in auth and token requests, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/307

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/ClaimsRequest.java
+++ b/src/main/java/com/microsoft/aad/msal4j/ClaimsRequest.java
@@ -15,22 +15,14 @@ import java.util.List;
  *
  * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
  */
-@Getter
-@Setter
 public class ClaimsRequest {
-    List<RequestedClaim> userInfoRequestedClaims = new ArrayList<>();
-    List<RequestedClaim> idTokenRequestedClaims = new ArrayList<>();
-    List<RequestedClaim> accessTokenRequestedClaims = new ArrayList<>();
 
-    /**
-     * Inserts a claim into the list of claims to be added to the "userinfo" section of an OIDC claims request
-     *
-     * @param claim the name of the claim to be requested
-     * @param requestedClaimAdditionalInfo additional information about the claim being requested
-     */
-    public void requestClaimInUserInfo(String claim, RequestedClaimAdditionalInfo requestedClaimAdditionalInfo) {
-        userInfoRequestedClaims.add(new RequestedClaim(claim, requestedClaimAdditionalInfo));
-    }
+    @Getter
+    @Setter
+    List<RequestedClaim> idTokenRequestedClaims = new ArrayList<>();
+
+    List<RequestedClaim> userInfoRequestedClaims = new ArrayList<>();
+    List<RequestedClaim> accessTokenRequestedClaims = new ArrayList<>();
 
     /**
      * Inserts a claim into the list of claims to be added to the "id_token" section of an OIDC claims request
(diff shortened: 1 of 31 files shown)


src/main/java/com/microsoft/aad/msal4j/SilentParameters.java
```diff
@@ -28,6 +28,11 @@
 
     private IAccount account;
 
+    /**
+     * Claims to be requested through the OIDC claims request parameter, allowing requests for standard and custom claims
+     */
+    private ClaimsRequest withClaims;
```
SomkaPe, line 34: for silent API this value will be ignored if we just return token form cache - might be miss-leading

SomkaPe (changes requested):

src/main/java/com/microsoft/aad/msal4j/IApiParameters.java
```diff
@@ -7,4 +7,6 @@
 
 interface IApiParameters {
     Set<String> scopes();
+
+    ClaimsRequest withClaims();
```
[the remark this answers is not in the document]
Avery-Dunn: 'withClaims' is the naming used by the .NET, Android, and (if I understand the syntax) iOS libraries, and it's what the design doc suggests so I'm sure the other libraries will follow it too once they implement it. Plus, having it named a bit different from the existing claims/claimsChallenge fields will probably help inform devs that this has a different purpose from the claims they've already been handling.

It can definitely be used for claims challenge (since they all get merged together before sending), but it is meant to be used in the more general 'request optional/custom claims' scenario. I believe there were some conversations about the design and whether to have separate claims challenge and claims request options or just a single object for all claims scenarios, and the conclusion was to keep them separate: claimsRequest for when a dev has claims they know they want to request, and claims challenge when the token response has some claims challenge and the dev can just copy/paste it into the field.

src/main/java/com/microsoft/aad/msal4j/RequestedClaimAdditionalInfo.java
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Represents the additional information that can be sent to an authorization server for a request claim in the claim request parameter
+ *
+ * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
+ */
+@Getter
+@Setter
+@AllArgsConstructor
+public class RequestedClaimAdditionalInfo {
```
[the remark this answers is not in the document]
Avery-Dunn: All of the new classes now use Jackson annotations and methods for making the JSON's.

src/main/java/com/microsoft/aad/msal4j/IApiParameters.java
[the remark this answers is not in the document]
SomkaPe: there few options to name setters like setField1(), withField1(), field1() - my point that we use 3rd option. All setters named this way, .Net uses second option.
using different approaches for setter naming in the same SDK is confusing

src/main/java/com/microsoft/aad/msal4j/ClaimsRequest.java
```diff
@@ -0,0 +1,88 @@
...
+import lombok.Setter;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Represents the claims request parameter as an object
+ *
+ * @see <a href="https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter">https://openid.net/specs/openid-connect-core-1_0-final.html#ClaimsParameter</a>
+ */
+public class ClaimsRequest {
```
SomkaPe, line 18: Did not we decide to go with just claims for accessToken ?

src/main/java/com/microsoft/aad/msal4j/IApiParameters.java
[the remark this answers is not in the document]
SomkaPe: so this is not supposed to be used for claimsChallenge ?

src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
```diff
@@ -0,0 +1,53 @@
...
+        cr.requestClaimInUserInfo("given_name", new RequestedClaimAdditionalInfo(true, null, null));
+        cr.requestClaimInUserInfo("email", null);
+        cr.requestClaimInIdToken("acr", new RequestedClaimAdditionalInfo(false, null, values));
+        cr.requestClaimInIdToken("sub", new RequestedClaimAdditionalInfo(true, "248289761001", null));
+
+        Assert.assertEquals(cr.formatAsJSONString(), TestConfiguration.CLAIMS_REQUEST);
+    }
+
+    @Test
+    public void testClaimsRequest_MergeWithClientCapabilities() {
```
SomkaPe, line 32: we need to test claims in actual requests send to STS,
Also we need to test merge logic with claim challenge

SomkaPe (changes requested):

src/main/java/com/microsoft/aad/msal4j/ClaimsRequest.java
Avery-Dunn: I believe we decided to just do ID tokens. The OIDC spec seems to support custom claims in for the id_token and userinfo sections, however we talked about how the library doesn't really do anything with userinfo.

However, claims in the access_token section are needed for client capabilities, so I decided to keep the userinfo and access_token methods but make them non-public, so that our library can still use them for setting up whatever we may need to (like client capabilities), and only the id_token claims are public and configurable by users.

src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
Avery-Dunn: Added tests merge behavior between claims requests/client capabilities and claims requests/claims challenge, as well as all three together.

src/main/java/com/microsoft/aad/msal4j/ClaimsRequest.java
Avery-Dunn: As discussed, this class lets users and the library set claims in the id_token field, only the library can set claims in the accessToken field, and I removed the ability to set claims in userinfo to avoid confusion (as these claims would not really ever be used).

SomkaPe (approved):

Issue #306, opened 2020-10-23
Support Azure AD region sepration


navyasric:
https://identitydivision.visualstudio.com/Engineering/_workitems/edit/954488

navyasric:
@Avery-Dunn Can this issue be closed since the feature is done?

PR #343, opened 2021-01-26
Region discovery support

Avery-Dunn:
Implements support for discovering Azure regions and fetching regional instance discovery metadata, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/306 and [this spec](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/2363?_a=files&path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md)

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -3,11 +3,15 @@
 
 package com.microsoft.aad.msal4j;
 
+import com.nimbusds.oauth2.sdk.http.HTTPResponse;
+
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.Set;
 import java.util.TreeSet;
+import java.util.Map;
+import java.util.HashMap;
 import java.util.concurrent.ConcurrentHashMap;
 
 class AadInstanceDiscoveryProvider {
@@ -15,7 +19,12 @@ class AadInstanceDiscoveryProvider {
     private final static String DEFAULT_TRUSTED_HOST = "login.microsoftonline.com";
     private final static String AUTHORIZE_ENDPOINT_TEMPLATE = "https://{host}/{tenant}/oauth2/v2.0/authorize";
     private final static String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE = "https://{host}/common/discovery/instance";
+    private final static String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE_WITH_REGION = "https://{region}.{host}/common/discovery/instance";
     private final static String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
+    private final static String REGION_NAME = "REGION_NAME";
+    // For information of the current api-version refer: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#versioning
+    private final static String DEFAULT_API_VERSION = "2020-06-01";
+    private final static String IMDS_ENDPOINT = "https://169.254.169.254/metadata/instance/compute/location?" + DEFAULT_API_VERSION + "&format=text";
 
     final static TreeSet<String> TRUSTED_HOSTS_SET = new TreeSet<>(String.CASE_INSENSITIVE_ORDER);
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -102,26 +111,85 @@ private static String getInstanceDiscoveryEndpoint(String host) {
...
+
     private static AadInstanceDiscoveryResponse sendInstanceDiscoveryRequest(URL authorityUrl,
                                                                              MsalRequest msalRequest,
                                                                              ServiceBundle serviceBundle) {
 
-        String instanceDiscoveryRequestUrl = getInstanceDiscoveryEndpoint(authorityUrl.getAuthority()) +
-                INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE.replace("{authorizeEndpoint}",
-                        getAuthorizeEndpoint(authorityUrl.getAuthority(),
-                                Authority.getTenant(authorityUrl, Authority.detectAuthorityType(authorityUrl))));
+        String region = "";
```
sangonzal, line 127: Can use `StringHelper.EmtpyString` here

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -102,26 +111,85 @@ private static String getInstanceDiscoveryEndpoint(String host) {
...
+        }
+
+        //If the region is unknown or the instance discovery failed at the region endpoint, try the global endpoint
+        if (region.isEmpty() || httpResponse == null || httpResponse.statusCode() != HTTPResponse.SC_OK) {
+            String instanceDiscoveryRequestUrl = getInstanceDiscoveryEndpoint(authorityUrl.getAuthority()) +
+                    INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE.replace("{authorizeEndpoint}",
+                            getAuthorizeEndpoint(authorityUrl.getAuthority(),
+                                    Authority.getTenant(authorityUrl, Authority.detectAuthorityType(authorityUrl))));
+
+            httpResponse = httpRequest(instanceDiscoveryRequestUrl, msalRequest.headers().getReadonlyHeaderMap(), msalRequest, serviceBundle);
```
sangonzal, line 152: There is some code duplication with 137-142. Possible to refactor into it's own method?

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -102,26 +111,85 @@ private static String getInstanceDiscoveryEndpoint(String host) {
...
+            //If call to IMDS endpoint was successful, return region from response body
+            if (httpResponse.statusCode() == HTTPResponse.SC_OK && !httpResponse.body().isEmpty()) {
+                return httpResponse.body();
+            }
+        } catch (Exception e) {
+            //IMDS call failed, cannot find region
+            return "";
+        }
+
+        return "";
```
sangonzal, line 192: Can use StringHelper.EMPTY_STRING. Also it would be helpful to add some logging to this file, especially to the case when things fail.

src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -135,4 +135,55 @@ public void aadInstanceDiscoveryTest_responseSetByDeveloper_invalidJson() throws
                 .aadInstanceDiscoveryResponse(instanceDiscoveryResponse)
                 .build();
     }
+
```
sangonzal, line 138: Should there also be a test for the happy case, where region is detected, to validate URL is correctly formatted?

Avery-Dunn: I was working on something for the happy case, but couldn't think of anything useful to test since all of the endpoints are only reachable in a VM.

There's two main things that'd be good to test, but would require too much hardcoding to be useful tests:

- The IMDS call in `discoverRegion()`: The [IMDS endpoint](https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service?tabs=windows) is only accessible in an Azure VM, so we'd need to just mock the response to that endpoint ourselves. However, all we do is pull the region name from the body of the response, so all the test would do is confirm that `httpResponse.body()` returns the body that we mocked.
- The instance discovery call in `sendInstanceDiscoveryRequest()`: This endpoint is only available in certain regions and for certain clients, so we'd need to mock the response here too. And if we mock the response, then we're only testing whether the response we mocked can be parsed just like a response from the global endpoint, and I don't believe that the metadata format is different if it comes from a regional endpoint.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: Done in the latest commit, along with your other empty string/code duplication comments

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: Added logging for success and failure cases in the most recent commit

sangonzal (approved):
Looks good to me!

src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
sangonzal: Makes sense.

PR #344, opened 2021-01-28
Add String-to-ClaimsRequest helper method

Avery-Dunn:
Adds a `formatAsClaimsRequest` method to the ClaimsRequest class to help create ClaimsRequest objects from JSON-formatted Strings of [OIDC claims](https://openid.net/specs/openid-connect-core-1_0.html#ClaimsParameter), such as those found in a claims challenge.


sangonzal (approved):
Looks good to me.

PR #385, opened 2021-05-04
fix foe issue with common scopes override

SomkaPe:


src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -14,42 +14,49 @@
...
-        this.grant = grant;
-
+    String addCommonScopes(String scopes){
+        String allScopes = COMMON_SCOPES_PARAM;
 
         if (!StringHelper.isBlank(scopes)) {
-            this.scopes = scopes;
-            params.put(SCOPE_PARAM_NAME,
-                    Collections.singletonList(String.join(" ",params.get(SCOPE_PARAM_NAME)) + SCOPES_DELIMITER + scopes));
+            allScopes = allScopes + SCOPES_DELIMITER + scopes;
```
sangonzal, line 27: Should we be merging these scopes instead? What if `scopes` contains some of the default scopes?

src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -14,42 +14,49 @@
...
     }
 
-    OAuthAuthorizationGrant(final AuthorizationGrant grant,
-                                final Map<String, List<String>> params) {
-        this();
+    OAuthAuthorizationGrant(AuthorizationGrant grant, String scopes, Map<String, List<String>> extraParams) {
         this.grant = grant;
+
+        String allScopes = addCommonScopes(scopes);
+        this.scopes = allScopes;
```
sangonzal, line 49:
```suggestion
        this.scopes = addCommonScopes(scopes);
```

src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java
```diff
@@ -37,12 +37,11 @@ private static OAuthAuthorizationGrant createAuthenticationGrant(OnBehalfOfParam
...
 
         Map<String, List<String>> params = new HashMap<>();
-        params.put("scope", Collections.singletonList(String.join(" ", parameters.scopes())));
         params.put("requested_token_use", Collections.singletonList("on_behalf_of"));
         if (parameters.claims() != null) {
             params.put("claims", Collections.singletonList(parameters.claims().formatAsJSONString()));
         }
 
-        return new OAuthAuthorizationGrant(jWTBearerGrant, params);
+        return new OAuthAuthorizationGrant(jWTBearerGrant, String.join(" ", parameters.scopes()), params);
```
sangonzal, line 45: nit: instead of using " ", can use SCOPES_DELIMITER.

src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
sangonzal: Requires an update on line 50 as well.

sangonzal (commented):
Overall looks good to me.

src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
SomkaPe: what do you mean ?

src/integrationtest/java/com.microsoft.aad.msal4j/OAuthRequestValidationUnitT.java
```diff
@@ -7,8 +7,7 @@
 import org.testng.Assert;
 import org.testng.annotations.Test;
 
-import java.util.Collections;
-import java.util.Map;
+import java.util.*;
```
sangonzal, line 10: I realize that intellij might do this by default, but most Java style guides that I've seen recommend against this. For example: https://google.github.io/styleguide/javaguide.html#s3.3.1-wildcard-imports

src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
SomkaPe: STS will ignore duplicates but, probably we should do it as well

src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
sangonzal: Just meant that instead of setting `allScopes` here, you can get rid of it and just use `this.scopes`.

src/integrationtest/java/com.microsoft.aad.msal4j/OAuthRequestValidationUnitT.java
SomkaPe: there are different opinions about it, some do not like scrolling through import statements while browsing git hub, that is why it is default in Intellij, but if you like it better I will change settings

Avery-Dunn (approved):

src/integrationtest/java/com.microsoft.aad.msal4j/OAuthRequestValidationUnitT.java
sangonzal: I'd prefer not to include them. It's better to be explicit. 95% of the time I'm reading code in an IDE, not on GitHub.

PR #387, opened 2021-05-11
Add CCS routing information to headers

sangonzal:
- Adding CCS Routing information to headers
- Rename `IApiParameters` to `IAcquireTokenParameters`
- Adding extraHttpHeaders to `IAcquireTokenParameters` 

[CCS MSAL routing header one pager - information available in epic](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/1331220)

[Design document](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/3304?path=%2FCCS%20Routing%20Headers%2FCCSRoutingHeaders.md&discussionId=31233&_a=files)

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
@@ -39,8 +39,8 @@ final class HttpHeaders {
 
     // Used for CCS routing
     static final String X_ANCHOR_MAILBOX = "X-AnchorMailbox";
-    private static final String X_ANCHOR_MAILBOX_OID_FORMAT = "oid:%s";
-    private static final String X_ANCHOR_MAILBOX_UPN_FORMAT = "upn:%s";
+    static final String X_ANCHOR_MAILBOX_OID_FORMAT = "oid:%s";
+    static final String X_ANCHOR_MAILBOX_UPN_FORMAT = "upn:%s";
     private String anchorMailboxHeaderValue = null;
 
     private String headerValues;
--- a/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
+++ b/src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
@@ -10,45 +10,64 @@ import java.util.function.BiConsumer;
 
 final class HttpHeaders {
 
-    final static String PRODUCT_HEADER_NAME = "x-client-SKU";
-    final static String PRODUCT_HEADER_VALUE = "MSAL.Java";
+    static final String PRODUCT_HEADER_NAME = "x-client-SKU";
+    static final String PRODUCT_HEADER_VALUE = "MSAL.Java";
 
-    final static String PRODUCT_VERSION_HEADER_NAME = "x-client-VER";
-    final static String PRODUCT_VERSION_HEADER_VALUE = getProductVersion();
+    static final String PRODUCT_VERSION_HEADER_NAME = "x-client-VER";
+    static final String PRODUCT_VERSION_HEADER_VALUE = getProductVersion();
 
-    final static String CPU_HEADER_NAME = "x-client-CPU";
-    final static String CPU_HEADER_VALUE = System.getProperty("os.arch");
+    static final String CPU_HEADER_NAME = "x-client-CPU";
+    static final String CPU_HEADER_VALUE = System.getProperty("os.arch");
 
-    final static String OS_HEADER_NAME = "x-client-OS";
-    final static String OS_HEADER_VALUE = System.getProperty("os.name");
+    static final String OS_HEADER_NAME = "x-client-OS";
+    static final String OS_HEADER_VALUE = System.getProperty("os.name");
 
-    final static String APPLICATION_NAME_HEADER_NAME = "x-app-name";
+    static final String APPLICATION_NAME_HEADER_NAME = "x-app-name";
     private final String applicationNameHeaderValue;
 
-    final static String APPLICATION_VERSION_HEADER_NAME = "x-app-ver";
+    static final String APPLICATION_VERSION_HEADER_NAME = "x-app-ver";
     private final String applicationVersionHeaderValue;
 
-    final static String CORRELATION_ID_HEADER_NAME = "client-request-id";
+    static final String CORRELATION_ID_HEADER_NAME = "client-request-id";
     private final String correlationIdHeaderValue;
 
-    private  final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_NAME = "return-client-request-id";
-    private final static String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_VALUE = "true";
+    private static final String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_NAME = "return-client-request-id";
+    private static final String REQUEST_CORRELATION_ID_IN_RESPONSE_HEADER_VALUE = "true";
 
-    private  final static String X_MS_LIB_CAPABILITY_NAME = "x-ms-lib-capability";
-    private final static String X_MS_LIB_CAPABILITY_VALUE = "retry-after, h429";
+    private static final String X_MS_LIB_CAPABILITY_NAME = "x-ms-lib-capability";
+    private static final String X_MS_LIB_CAPABILITY_VALUE = "retry-after, h429";
 
-    private final String headerValues;
-    private final Map<String, String> headerMap = new HashMap<>();
+    // Used for CCS routing
+    static final String X_ANCHOR_MAILBOX = "X-AnchorMailbox";
+    private static final String X_ANCHOR_MAILBOX_OID_FORMAT = "oid:%s";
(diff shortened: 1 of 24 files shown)


src/main/java/com/microsoft/aad/msal4j/IntegratedWindowsAuthenticationParameters.java
```diff
@@ -21,24 +22,30 @@
 @Accessors(fluent = true)
 @Getter
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
-public class IntegratedWindowsAuthenticationParameters implements IApiParameters {
+public class IntegratedWindowsAuthenticationParameters implements IAcquireTokenParameters {
```
SomkaPe, line 25: is extraHttpHeaders applicable for any implementation of IAcquireTokenParameters ?
if yes Abstract class should be introduced to avoid code duplication

src/main/java/com/microsoft/aad/msal4j/RequestContext.java
```diff
@@ -44,6 +45,14 @@ public RequestContext(AbstractClientApplicationBase clientApplication,
         this.apiParameters = apiParameters;
     }
 
+    public RequestContext(AbstractClientApplicationBase clientApplication,
+                          PublicApi publicApi,
+                          IAcquireTokenParameters apiParameters,
```
SomkaPe, line 50: userIdentifier should be part of IAcquireTokenParameters, so should be available through it, so no need to add this constructor ?

src/main/java/com/microsoft/aad/msal4j/UserIdentifier.java
```diff
@@ -0,0 +1,48 @@
...
+
+    public static UserIdentifier fromHomeAccountId(String homeAccountId) {
+        UserIdentifier userIdentifier = new UserIdentifier();
+
+        // HomeAccountId is userObjectId.userTenantId
+        String[] homeAccountIdParts = homeAccountId.split("\\.");
+        if (homeAccountIdParts.length < 2
+                || StringHelper.isBlank(homeAccountIdParts[0])
+                || StringHelper.isBlank(homeAccountIdParts[1])) {
+            userIdentifier.oid = StringHelper.EMPTY_STRING;
```
SomkaPe, line 41: why not null ? in general  empty string  is valid value, if we do not know value it is null

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
```diff
@@ -21,8 +23,12 @@ public void testHttpHeaderConstructor(){
                 .applicationVersion("app-version")
                 .build();
 
+        IAcquireTokenParameters parameters = UserNamePasswordParameters
+                .builder(Collections.singleton("scopes"), "username", "password".toCharArray())
+                .build();
+
         RequestContext requestContext = new RequestContext(
-                app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, null);
+                app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, parameters);
```
SomkaPe, line 31: what is the point of adding this, I see no assert where this is checked

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
```diff
@@ -38,14 +44,18 @@ public void testHttpHeaderConstructor(){
...
                 .builder("client-id")
                 .build();
 
+        IAcquireTokenParameters parameters = UserNamePasswordParameters
+                .builder(Collections.singleton("scopes"), "username", "password".toCharArray())
+                .build();
+
         RequestContext requestContext = new RequestContext(
-                app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, null);
+                app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, parameters);
```
SomkaPe, line 58: nothing is tested here, except that this call will not throw exception

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
```diff
@@ -59,4 +69,94 @@ public void testHttpHeaderConstructor_valuesNotSet(){
...
+                app,
+                PublicApi.ACQUIRE_TOKEN_BY_USERNAME_PASSWORD,
+                parameters,
+                UserIdentifier.fromUpn(upn));
+
+        HttpHeaders httpHeaders = new HttpHeaders(requestContext);
+
+        Map<String, String> httpHeaderMap = httpHeaders.getReadonlyHeaderMap();
+
+        String expectedValue = String.format("%s:%s", "upn", upn);
```
SomkaPe, line 95: probably better to reuse constant for format from corresponding class

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
```diff
@@ -59,4 +69,94 @@ public void testHttpHeaderConstructor_valuesNotSet(){
...
+                .build();
+
+        RequestContext requestContext = new RequestContext(
+                app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, parameters);
+
+        HttpHeaders httpHeaders = new HttpHeaders(requestContext);
+
+        Map<String, String> httpHeaderMap = httpHeaders.getReadonlyHeaderMap();
+
+        Assert.assertEquals(httpHeaderMap.get(HttpHeaders.PRODUCT_HEADER_NAME), HttpHeaders.PRODUCT_HEADER_VALUE);
```
SomkaPe, line 153: would be good to have comments to distinguish standard headers, overwritten standard headers and extra header

src/main/java/com/microsoft/aad/msal4j/IntegratedWindowsAuthenticationParameters.java
sangonzal: I took a look at this and since the parameter classes are Lombok builders,  IAcquireTokenParameters would have to be an abstract builder. If I remember correctly we couldn't get the Lombok annotations to play nicely with abstract builders in the ClientApplication builders, so they had to be written out. 

Updating this here would require the same change, and would actually result in writing a lot more code (writing out all of the Parameter builders) in exchange for avoiding some code duplication.

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
sangonzal: Agreed, updated.

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
sangonzal: Added some comments.

src/main/java/com/microsoft/aad/msal4j/RequestContext.java
sangonzal: `UserIdentifier` is not part of `IAcquireTokenParameters`, because not all acquire token calls will have an `UserIdentifier` (for example, client credentials).

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
sangonzal: Due to the changes in `HttpHeaders`, unless parameters were passed in this would lead to null ref.

src/test/java/com/microsoft/aad/msal4j/HttpHeaderTest.java
sangonzal: Due to the changes in HttpHeaders, unless parameters were passed in this would lead to null ref.

src/main/java/com/microsoft/aad/msal4j/UserIdentifier.java
sangonzal: Updated to null.

src/main/java/com/microsoft/aad/msal4j/HttpHeaders.java
```diff
@@ -10,45 +10,64 @@
...
+            String upn = requestContext.userIdentifier().upn();
+            String oid = requestContext.userIdentifier().oid();
+            if (!StringHelper.isBlank(upn)) {
+                anchorMailboxHeaderValue = String.format(X_ANCHOR_MAILBOX_UPN_FORMAT, upn);
+            } else if (!StringHelper.isBlank(oid)) {
+                anchorMailboxHeaderValue = String.format(X_ANCHOR_MAILBOX_OID_FORMAT, oid);
+            }
+        }
+
+        Map<String, String> extraHttpHeaders = requestContext.apiParameters() == null ?
```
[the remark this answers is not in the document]
sangonzal: Can be null in `getAccounts` or `removeAccounts`.

Avery-Dunn (approved):

SomkaPe (approved):

PR #388, opened 2021-05-14
Improve region support and add region telemetry

Avery-Dunn:
Improves support for Azure regions and updates telemetry to follow the most recent schema, as per [this region design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FPinAuthToRegion%2FAAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&_a=preview) and [this telemetry design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=%2FTelemetry%2FMSALServerSideTelemetry.md&version=GBdev&_a=preview).

Additionally, updates the oauth2-oidc-sdk dependency to fix an issue with the transitive json-smart dependency mentioned in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/394

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -129,33 +129,58 @@ class AadInstanceDiscoveryProvider {
                                                                              MsalRequest msalRequest,
                                                                              ServiceBundle serviceBundle) {
 
-        String region = StringHelper.EMPTY_STRING;
         IHttpResponse httpResponse = null;
+        String providedRegion = msalRequest.application().azureRegion();
+        String detectedRegion = null;
+        int regionOutcome = 0;
+        String regionToUse = null;
 
-        //If the autoDetectRegion parameter in the request is set, attempt to discover the region
-        if (msalRequest.application().autoDetectRegion()) {
-            region = discoverRegion(msalRequest, serviceBundle);
+        //If a region was provided by a developer or they set the autoDetectRegion parameter,
+        // attempt to discover the region and set telemetry info based on the outcome
+        if (providedRegion != null) {
+            detectedRegion = discoverRegion(msalRequest, serviceBundle);
+            regionToUse = providedRegion;
+
+            if (detectedRegion == null) {
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_DEVELOPER_AUTODETECT_FAILED.telemetryValue;
+            } else if (providedRegion.equals(detectedRegion)) {
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_DEVELOPER_AUTODETECT_MATCH.telemetryValue;
+            } else {
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_DEVELOPER_AUTODETECT_MISMATCH.telemetryValue;
+            }
+        } else if (msalRequest.application().autoDetectRegion()) {
+            detectedRegion = discoverRegion(msalRequest, serviceBundle);
+
+            if (detectedRegion != null) {
+                regionToUse = detectedRegion;
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_AUTODETECT_SUCCESS.telemetryValue;
+            } else {
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_AUTODETECT_FAILED.telemetryValue;
+            }
         }
 
         //If the region is known, attempt to make instance discovery request with region endpoint
-        if (!region.isEmpty()) {
-            String instanceDiscoveryRequestUrl = getInstanceDiscoveryEndpointWithRegion(authorityUrl.getAuthority(), region) +
+        if (regionToUse != null) {
+            String instanceDiscoveryRequestUrl = getInstanceDiscoveryEndpointWithRegion(authorityUrl.getAuthority(), regionToUse) +
                     formInstanceDiscoveryParameters(authorityUrl);
 
-            httpResponse = httpRequest(instanceDiscoveryRequestUrl, msalRequest.headers().getReadonlyHeaderMap(), msalRequest, serviceBundle);
+            try {
+                httpResponse = httpRequest(instanceDiscoveryRequestUrl, msalRequest.headers().getReadonlyHeaderMap(), msalRequest, serviceBundle);
+            } catch (MsalClientException ex) {
+                log.warn("Could not retrieve regional instance discovery metadata, falling back to global endpoint");
+            }
         }
 
         //If the region is unknown or the instance discovery failed at the region endpoint, try the global endpoint
-        if (region.isEmpty() || httpResponse == null || httpResponse.statusCode() != HTTPResponse.SC_OK) {
-            if (!region.isEmpty()) {
-                log.warn("Could not retrieve regional instance discovery metadata, falling back to global endpoint");
-            }
+        if ((detectedRegion == null && providedRegion == null) || httpResponse == null || httpResponse.statusCode() != HTTPResponse.SC_OK) {
 
             String instanceDiscoveryRequestUrl = getInstanceDiscoveryEndpoint(authorityUrl.getAuthority()) +
                     formInstanceDiscoveryParameters(authorityUrl);
 
             httpResponse = httpRequest(instanceDiscoveryRequestUrl, msalRequest.headers().getReadonlyHeaderMap(), msalRequest, serviceBundle);
         }
+
+        serviceBundle.getServerSideTelemetry().getCurrentRequest().regionOutcome(regionOutcome);
 
         return JsonHelper.convertJsonToObject(httpResponse.body(), AadInstanceDiscoveryResponse.class);
     }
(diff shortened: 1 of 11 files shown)


src/main/java/com/microsoft/aad/msal4j/ServerSideTelemetry.java
```diff
@@ -17,7 +16,7 @@
 
     private final static Logger log = LoggerFactory.getLogger(ServerSideTelemetry.class);
 
-    private final static String SCHEMA_VERSION = "2";
+    private final static String SCHEMA_VERSION = "5";
```
sangonzal, line 19: Should LastRequest also be bumped up to version 5? I think keeping current request and last request in sync could be important from a kusto query perspective.

sangonzal (commented):
Just a question about keeping CurrentRequest and LastRequest in sync. Otherwise looks good to me.

src/main/java/com/microsoft/aad/msal4j/ServerSideTelemetry.java
Avery-Dunn: It turns out that LastRequest hasn't changed between schema versions 2 and 5 so no worries there, and in the latest commit I made all of the non-regional telemetry changes to more thoroughly follow schema version 5.

src/main/java/com/microsoft/aad/msal4j/RegionTelemetry.java
```diff
@@ -0,0 +1,36 @@
...
+    REGION_OUTCOME_DEVELOPER_AUTODETECT_MISMATCH(3),
+    REGION_OUTCOME_AUTODETECT_SUCCESS(4),
+    REGION_OUTCOME_AUTODETECT_FAILED(5);
+
+    final int telemetryValue;
+
+    RegionTelemetry(int telemetryValue){
+        this.telemetryValue = telemetryValue;
+    }
+
```
sangonzal, line 35: nit: extra space

sangonzal (approved):
Looks good to me

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -129,34 +129,59 @@ private static AadInstanceDiscoveryResponse sendInstanceDiscoveryRequest(URL aut
...
-            region = discoverRegion(msalRequest, serviceBundle);
+        String providedRegion = msalRequest.application().azureRegion();
+        String detectedRegion = null;
+        int regionOutcome = 0;
+        String regionToUse = null;
+
+        //If a region was provided by a developer or they set the autoDetectRegion parameter,
+        // attempt to discover the region and set telemetry info based on the outcome
+        if (providedRegion != null) {
+            detectedRegion = discoverRegion(msalRequest, serviceBundle);
```
SomkaPe, line 141: why do we discover region if it was provided and auto detect might not be enabled?

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -129,34 +129,59 @@ private static AadInstanceDiscoveryResponse sendInstanceDiscoveryRequest(URL aut
...
 
-        String region = StringHelper.EMPTY_STRING;
         IHttpResponse httpResponse = null;
-
-        //If the autoDetectRegion parameter in the request is set, attempt to discover the region
-        if (msalRequest.application().autoDetectRegion()) {
-            region = discoverRegion(msalRequest, serviceBundle);
+        String providedRegion = msalRequest.application().azureRegion();
+        String detectedRegion = null;
+        int regionOutcome = 0;
```
SomkaPe, line 135: should be renamed to reflect usage of variable as telemetry datapoint

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -180,9 +205,12 @@ private static IHttpResponse httpRequest(String requestUrl, Map<String, String>
 
     private static String discoverRegion(MsalRequest msalRequest, ServiceBundle serviceBundle) {
```
SomkaPe, line 206: msalRequest does not seem to used

src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -603,6 +608,21 @@ public T autoDetectRegion(boolean val) {
...
+         *  the default behavior if this value is not set.
+         *
+         * @param val boolean (default is false)
+         * @return instance of the Builder on which method was called
+         */
+        public T azureRegion(String val) {
+            azureRegion = val;
+            return self();
+        }
+
```
SomkaPe, line 625: what if  region is provided and autodetect is enabled ?

src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -52,6 +52,22 @@ AuthenticationResult execute() throws Exception {
                     res.refreshOn() < currTimeStampSec && res.expiresOn() >= currTimeStampSec;
 
             if (silentRequest.parameters().forceRefresh() || afterRefreshOn || StringHelper.isBlank(res.accessToken())) {
+
+                //Set telemetry info for reason behind the refresh
```
SomkaPe, line 56: what does it mean?

src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -52,6 +52,22 @@ AuthenticationResult execute() throws Exception {
...
+                    clientApplication.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
+                            CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue);
+                } else if (res.expiresOn() < currTimeStampSec) {
+                    clientApplication.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
+                            CacheTelemetry.REFRESH_ACCESS_TOKEN_EXPIRED.telemetryValue);
+                } else if (StringHelper.isBlank(res.accessToken())) {
+                    clientApplication.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
+                            CacheTelemetry.REFRESH_NO_ACCESS_TOKEN.telemetryValue);
+                }
+
```
SomkaPe, line 70: why it is else if sequence ? this conditions does not seem to be related to each other - why you want to use just one value ? why this particular sequence ?

src/main/java/com/microsoft/aad/msal4j/RegionTelemetry.java
Avery-Dunn: Fixed in the latest commit.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -129,34 +129,59 @@ private static AadInstanceDiscoveryResponse sendInstanceDiscoveryRequest(URL aut
...
+            }
+        } else if (msalRequest.application().autoDetectRegion()) {
+            detectedRegion = discoverRegion(msalRequest, serviceBundle);
+
+            if (detectedRegion != null) {
+                regionToUse = detectedRegion;
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_AUTODETECT_SUCCESS.telemetryValue;
+            } else {
+                regionOutcome = RegionTelemetry.REGION_OUTCOME_AUTODETECT_FAILED.telemetryValue;
+            }
```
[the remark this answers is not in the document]
Avery-Dunn: I've extracted the telemetry determination to a separate method in the latest commit. There's still quite a few if/else's, but it should be easier to read now.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: The region design doc specifies that if the region API is used at all, then the library should attempt to discover the region. From the spec: "Region feature is opt-in ... if developer opts in, region auto-detection should always happen and telemetry MUST be used to send back region data,"

This is entirely for telemetry: half of the 'region outcome' telemetry values deal with the dev setting the region, and autodetect succeeding/failing/having a mismatch with the provided region. So, even if it's not needed when the region is provided, I guess it's considered an important piece of telemetry info.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: Good point, renamed to `regionOutcomeTelemetryValue` in the latest commit.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: It's referenced in the try/catch block of that method where it makes the call to the IMDS endpoint. It's needed in the `httpRequest` method that makes the actual call.

src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
Avery-Dunn: I don't think there's an actual use case for using both, however the design doc doesn't say they should be mutually exclusive (i.e., we shouldn't throw an error or anything if the dev tries to use both). 

Autodetection happens anyway if a region is provided so the autodetect flag is pretty much irrelevant in that case, and the dev provided region takes precedence over the autodetected one.

src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Newer versions of the telemetry schema have a field for collecting data about why a token was refreshed (i.e., because of force refresh, it expired, there was no access token, etc.), and this if/else block sets the telemetry value based on the reason.

I've clarified that comment in the latest commit.

SomkaPe (approved):

Issue #371, opened 2021-04-14
PublicClientApplication auth fails when username contains hash symbol (#)


Tim-Ford-Imprivata:
### Summary
I'm doing password authentication using the PublicClientApplication. I am finding that, if the UPN includes a hash symbol, authentication fails. 

Here is a example of pseudo code showing the issue:
```
  String username = "upn#with#hashes@example.microsoftonline.com";
  UserNamePasswordParameters parameters = 
	UserNamePasswordParameters
      .builder(SCOPE, username, password)
	  .build();
  PublicClientApplication pca = 
	PublicClientApplication.builder(clientId)
      .authority(AUTHORITY_URL)
      .build();
  pca.acquireToken(parameters);
```

Here is the resulting stack trace:
 ```
 java.lang.IllegalArgumentException: argument "content" is null
	at com.fasterxml.jackson.databind.ObjectMapper._assertNotNull(ObjectMapper.java:4737)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:3435)
	at com.microsoft.aad.msal4j.JsonHelper.convertJsonToObject(JsonHelper.java:30)
	at com.microsoft.aad.msal4j.UserDiscoveryRequest.execute(UserDiscoveryRequest.java:32)
	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.processPasswordGrant(AcquireTokenByAuthorizationGrantSupplier.java:91)
	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.execute(AcquireTokenByAuthorizationGrantSupplier.java:43)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:59)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:17)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1764)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1756)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:290)
[...]

Issue #363, opened 2021-03-24
Misleading error: Device authentication expired when should fail


ohadbitt:
Using msal4j 1.9.1 :
```scala
 val deviceCodeFlowParams: DeviceCodeFlowParameters = DeviceCodeFlowParameters.builder(scopes, deviceCodeConsumer).build
    clientApplication.acquireToken(deviceCodeFlowParams).join()
    ```

scopes here was given bad input (some name instead of URL) and the exception thrown is : 

java.util.concurrent.CompletionException: com.microsoft.aad.msal4j.MsalClientException: Expired Device code

	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:89)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:17)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1604)
	at java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java)
	at java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1596)
	at java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:289)
	at java.util.concurrent.ForkJoinPool$WorkQueue.runTask(ForkJoinPool.java:1067)
	at java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1703)
	at java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:172)
Caused by: com.microsoft.aad.msal4j.MsalClientException: Expired Device code
	at com.microsoft.aad.msal4j.AcquireTokenByDeviceCodeFlowSupplier.acquireTokenWithDeviceCode(AcquireTokenByDeviceCodeFlowSupplier.java:68)
	at com.microsoft.aad.msal4j.AcquireTokenByDeviceCodeFlowSupplier.execute(AcquireTokenByDeviceCodeFlowSupplier.java:27)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:59)
	... 8 more
	
	
I guess this is because expiresIn has a zero value but its missleading

Avery-Dunn:
[...]

PR #397, opened 2021-05-27
Improve error messages

sangonzal:
- For HttpRequests outside of /token endpoint, MSAL was not checking status code before attempting to parse response. When getting non 200 responses from service, MSAL would error out on the parsing of the response, and would emit either an incorrect message/ unhelpful error message. For example, see #371 and #363
- Certain places where throwing generic `Exception` . Updates to throw either `MsalClientException` or `MsalServiceException`

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
+++ b/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
@@ -25,9 +25,12 @@ class JsonHelper {
         mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
     }
 
-    static <T> T convertJsonToObject(final String json, final Class<T> clazz) {
+    private JsonHelper(){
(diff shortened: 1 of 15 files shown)


src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -86,36 +89,6 @@ static String mergeJSONString(String mainJsonString, String addJsonString) {
         return mainJson.toString();
     }
 
-    /**
-     * Merges set of given JSON strings into one Jackson JsonNode object, which is returned as a String
-     */
-    static String mergeJSONString(Set<String> jsonStrings) {
```
sangonzal, line 92: Not being used anywhere

Avery-Dunn (approved):

SomkaPe (approved):

PR #409, opened 2021-06-25
Use host parameter instead of authority

MSSedusch:
When using this library on SAP NetWeaver Java, I always got the following exception:
java.net.MalformedURLException: Invalid host: [login.microsoftonline.com:443]

authorityUrl was pointing to "https://login.microsoftonline.com:443/<tenant>". getAuthority() would return login.microsoftonline.com:443, however, using that string in
AuthenticationResultSupplier.java:38 would cause this exception.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -50,13 +50,13 @@ class AadInstanceDiscoveryProvider {
                                                            MsalRequest msalRequest,
                                                            ServiceBundle serviceBundle) {
 
-        InstanceDiscoveryMetadataEntry result = cache.get(authorityUrl.getAuthority());
+        InstanceDiscoveryMetadataEntry result = cache.get(authorityUrl.getHost());
 
         if (result == null) {
             doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
         }
 
-        return cache.get(authorityUrl.getAuthority());
+        return cache.get(authorityUrl.getHost());
     }
 
     static Set<String> getAliases(String host){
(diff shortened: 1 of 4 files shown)


sangonzal:
Hi @MSSedusch, is there a reason why you are passing in the authority name with a port number on it? Would your scenario work if you just used https://login.microsoftonline.com instead?

MSSedusch:
@sangonzal the code I own does not set the login URL, it just uses com.azure.core.management.AzureEnvironment.AZURE. The code is sideloaded into a Java application (SAP NetWeaver). Not sure where exactly the port is coming from. My point is that the authority is valid (login.microsoftonline.com:443), it could theoretically be even something like user@login.microsoftonline.com:443 which is not a valid host but msal passes it as the host parameter to the URL.

sangonzal:
@MSSedusch Thanks for the explanation. This is indeed an issue with how MSAL handles ports. Your fix looks good, I made one more update and added a test.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -205,7 +205,7 @@ else if (autoDetect) {
 
     private static String formInstanceDiscoveryParameters(URL authorityUrl) {
         return INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE.replace("{authorizeEndpoint}",
-                getAuthorizeEndpoint(authorityUrl.getAuthority(),
+                getAuthorizeEndpoint(authorityUrl.getHost(),
```
SomkaPe, line 208: if port specified for authority, it probably should be used with derived endpoints like authorize and instance discovery, is not it ?

MSSedusch: Can you give me more details on what you mean by that? Do you think the fix is not complete and I need to make changes in other areas as well?

sangonzal: SomkaPe and I discussed these changes offline. This change is not actually changing the authorization endpoint, instead its setting the authorization endpoint query parameter in the instance discovery URL. No changes needed on your end, I'll take care of finishing touches before merging.

Avery-Dunn (approved):

SomkaPe (approved):

Issue #415, opened 2021-07-09
AcquireTokenSilently with empty scope causes AT cache miss


sangonzal:
Acquiring a token via any flow with an empty scope ("") will populate the token cache with an access token with scopes "openid offline_access profile", as those are the default scopes that are always included in every request. 

When acquiring token silently afterwards with the same account and scope as the initial request, it is expected that the cached AT is returned as longs as it has not expired. Instead, when doin the AT lookup, there is a scope mismatch and the RT is used to refresh the AT.

Steps to reproduce: 
1) Run `AcquireTokenByXXX` to acquire token for an empty scope (scope= "")
2) Use same client app to acquire token silently with same parameters.

PR #422, opened 2021-07-19
Allow empty set and remove empty string if present in scopes

sangonzal:
Fix for #415

- Allow empty set for scopes. 
- Remove empty string if included in scope for silent request

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/SilentParameters.java
+++ b/src/main/java/com/microsoft/aad/msal4j/SilentParameters.java
@@ -100,7 +100,7 @@ public class SilentParameters implements IAcquireTokenParameters {
         Set<String> updatedScopes = new HashSet<>();
         for(String scope: scopes){
             if(!scope.equalsIgnoreCase(StringHelper.EMPTY_STRING)){
-                updatedScopes.add(scope);
+                updatedScopes.add(scope.trim());
             }
         }
         return updatedScopes;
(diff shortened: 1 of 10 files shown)


Avery-Dunn (approved):

src/main/java/com/microsoft/aad/msal4j/SilentParameters.java
```diff
@@ -88,8 +88,21 @@ public static SilentParametersBuilder builder(Set<String> scopes, IAccount accou
...
+        return builder().scopes(removeEmptyScope(scopes));
+    }
+
+    private static Set<String> removeEmptyScope(Set<String> scopes){
+        // empty string is not a valid scope, but we currently accept it and can't remove support
+        // for it yet as its a breaking change. This will be removed eventually (throwing
+        // exception if empty scope is passed in).
+        Set<String> updatedScopes = new HashSet<>();
+        for(String scope: scopes){
+            if(!scope.equalsIgnoreCase(StringHelper.EMPTY_STRING)){
```
SomkaPe, line 102: trim might be also used to remove trailing and leading whitespaces

SomkaPe (approved):

src/main/java/com/microsoft/aad/msal4j/SilentParameters.java
sangonzal: Good point, I will add trim here.

PR #427, opened 2021-08-04
Silent calls now use cached environment info

Avery-Dunn:
In some situations (described [here](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview/pullrequest/1868?_a=files&path=%2F%5Bxplat%5D%20Authority%20Mapping%2FSolve%20discrepancy%20between%20request%20authority%20and%20cache%20authority.md)), the authority set in the client app may differ from the cached authority info, causing issues during silent calls such as cache misses or refresh requests sent to the wrong authority. These changes plug some of those gaps by retrieving the cached environment info during silent calls.

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
@@ -44,7 +44,7 @@ public class DeviceCodeIT {
                 build();
 
         Consumer<DeviceCode> deviceCodeConsumer = (DeviceCode deviceCode) -> {
-            runAutomatedDeviceCodeFlow(deviceCode, user, environment);
+            runAutomatedDeviceCodeFlow(deviceCode, user);
         };
 
         IAuthenticationResult result = pca.acquireToken(DeviceCodeFlowParameters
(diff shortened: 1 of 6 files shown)


src/integrationtest/java/com.microsoft.aad.msal4j/DeviceCodeIT.java
```diff
@@ -57,8 +57,8 @@ public void DeviceCodeFlowADTest(String environment) throws Exception {
         Assert.assertFalse(Strings.isNullOrEmpty(result.accessToken()));
     }
 
-    @Test(dataProvider = "environments", dataProviderClass = EnvironmentsProvider.class)
-    public void DeviceCodeFlowADFSv2019Test(String environment) throws Exception {
+    @Test()
```
sangonzal, line 60: Why are you getting rid of environments for some of the tests?

Avery-Dunn: It didn't seem like environment was actually being used in those those tests. All of the config (like user or authority) referenced TestConstants directly, so defining the environments just made the exact same test run twice.

src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -69,6 +70,14 @@ AuthenticationResult execute() throws Exception {
                 }
 
                 if (!StringHelper.isBlank(res.refreshToken())) {
+                    //There are certain scenarios where the cached authority may differ from the client app's authority,
+                    // such as when a request is instance aware. Unless overridden by SilentParameters.authorityUrl, the
+                    // cached authority should be used in the token refresh request
```
SomkaPe, line 75: this statement does not sound correct, cache can contain tokens issued by different authorities, for silent call, requested authority should be used

siddhijain (approved):

Issue #429, opened 2021-08-10
Authority validation differ between app building and parameter


danilowitsch:
If you create a publc application and add the authority url: https://login.microsoftonline.com/<my-tenant-id> without a trailing slash the api will add the missing slash.

If you create an SilentParamater instance for a SilentRequest you are also able to add the authorityUrl. In this case the url will not check for a missing trailing slash and the parsing within the SilentRequest constructor fails with an IndexOutOfBound Exception.

> 
    PublicClientApplication pca = PublicClientApplication.builder(config.getClientId())
        .authority(config.getAuthority())
        .build();

    SilentParameters.builder(Collections.singleton("User.Read"))
        .authorityUrl(config.getAuthority())
        .build();

Avery-Dunn:
@danilowitsch : Thanks for bringing this to our attention. I'm able to reproduce it and it definitely seems like a bug, so we should have a fix for that trailing slash problem in our next release.

danilowitsch:
Nice.

danilowitsch:
Sorry, my intention was only to comment this.

Avery-Dunn:
Sorry for the long delay on this, but for anyone having this issue it was addressed as part of https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/431, and should be fixed as of the recent [1.12.0 release](https://search.maven.org/artifact/com.microsoft.azure/msal4j/1.12.0/jar).

PR #431, opened 2021-08-11
Add trailing slash to authorities in silent calls

Avery-Dunn:
Copies a method used in `AbstractClientApplicationBase` to enforce a trailing slash in a `SilentRequest`

Fixes an issue mentioned in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/429, where authorities passed to silent calls would cause an exception if they didn't have a trailing slash.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/SilentRequest.java
+++ b/src/main/java/com/microsoft/aad/msal4j/SilentRequest.java
@@ -28,11 +28,20 @@ class SilentRequest extends MsalRequest {
         this.assertion = assertion;
         this.requestAuthority = StringHelper.isBlank(parameters.authorityUrl()) ?
                 application.authenticationAuthority :
-                Authority.createAuthority(new URL(parameters.authorityUrl()));
+                Authority.createAuthority(new URL(enforceTrailingSlash(parameters.authorityUrl())));
 
         if (parameters.forceRefresh()) {
             application.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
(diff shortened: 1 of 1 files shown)


src/main/java/com/microsoft/aad/msal4j/SilentRequest.java
```diff
@@ -28,11 +28,20 @@
...
-                Authority.createAuthority(new URL(parameters.authorityUrl()));
+                Authority.createAuthority(new URL(enforceTrailingSlash(parameters.authorityUrl())));
 
         if (parameters.forceRefresh()) {
             application.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
                     CacheTelemetry.REFRESH_FORCE_REFRESH.telemetryValue);
         }
     }
+
+    protected static String enforceTrailingSlash(String authority) {
```
sangonzal, line 39: Can you just use the method defined in `AbstractClientApplicationBase` instead of duplicating this code? It might also make sense to just move the helper method into `Authority`.

Avery-Dunn: I originally just duplicated it because it felt weird for one of the token request classes to be using methods from `AbstractClientApplicationBase`, however moving it all to `Authority` makes a lot of sense and I just made that change in the latest commit.

sangonzal (approved):

SomkaPe (approved):

siddhijain (approved):

PR #440, opened 2021-09-21
Remove info flagged by automated build step

Avery-Dunn:
Seems like a step in the build pipeline has started flagging some test data that's been there since 2019, causing builds to fail. It wasn't directly referenced by any test, so this PR just removes it to get the builds working again.

Additionally removes references to old Azure Germany Cloud, since those endpoints are deprecated


siddhijain (approved):
Looks good

PR #457, opened 2022-01-19
Remove transitive dependency on retrofit.

siddhijain:
Update msal-java library to use the latest version of azure-keyvault, thereby also removing the transitive dependency on retrofit with security vulnerabilities.


Avery-Dunn (approved):
Looks good

PR #458, opened 2022-01-25
Sjain/client assertion exception

siddhijain:
Code to catch exception when validation fails for subject and issuer not matching while using ConfidentialClientApplicationBuilder using ClientAssertion.
Issue # 437


src/main/java/com/microsoft/aad/msal4j/CustomJWTAuthentication.java
```diff
@@ -0,0 +1,18 @@
+package com.microsoft.aad.msal4j;
```
Avery-Dunn, line 1: The top of each class should have a comment about copyright and the open-source license we use. You can find it at the top of every file, it's just:
`// Copyright (c) Microsoft Corporation. All rights reserved.`
`// Licensed under the MIT License.`

This should be added to `ConfidentialClientApplicationTest.java` too

siddhijain: Sure updated the file. Also, moved tests from ConfidentialClientApplicationTest to an existing file under Integration tests.

Avery-Dunn (approved):

Issue #437, opened 2021-09-07
client assertion does not work when its issuer and subject don't match


chewong:
```java
public Mono<AccessToken> getToken(TokenRequestContext request)  {
    Map<String, String> env = System.getenv();
    String clientAssertion = "<jwt>";

    IClientCredential credential = ClientCredentialFactory.createFromClientAssertion(clientAssertion);
    String authority = "...";
    try {
        ConfidentialClientApplication app = ConfidentialClientApplication.builder(env.get("AZURE_CLIENT_ID"), credential).authority(authority).build();

        Set<String> scopes = new HashSet<>();
        for (String scope : request.getScopes()) scopes.add(scope);

        ClientCredentialParameters parameters = ClientCredentialParameters.builder(scopes).build();
        IAuthenticationResult result = app.acquireToken(parameters).join();
        return Mono.just(new AccessToken(result.accessToken(), result.expiresOnDate().toInstant().atOffset(ZoneOffset.UTC)));
    } catch (Exception e) {
        System.out.printf("Error creating client application: %s", e.getMessage());
        System.exit(1);
    }

    return Mono.empty();
}
```

I got the following error:

```
Error creating client application: com.nimbusds.oauth2.sdk.ParseException: Issuer and subject in client JWT assertion must designate the same client identifier
```

I have used msal-python, msal-go, msal-net, msal-node, and they all do not require client assertion's issuer and subject to be the same.

Avery-Dunn:
Hello @chewong : Sorry for the delayed response. That error message is coming from a package we use, [com.nimbusds.oauth2.sdk](https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/latest/src-html/com/nimbusds/oauth2/sdk/auth/JWTAuthentication.html), which implements the OAuth2 spec and we use it for most of this kind of JWT parsing. 

I took a quick look through the JWT client spec and it doesn't say that they need to match so I'm not sure why the dependency is checking it, and confirmed with at least the MSAL .NET folks that they library allows it like you said. We'll work on a way to get around that check, unfortunately it's a thing that some other part of the dependency calls so we may need to replace more than just this. I'll update this thread once we have the fix.

siddhijain:
@chewong Latest version of MSAL(1.11.1) is now released that contains fix of this issue.

vdzhuvinov:
[...]

PR #465, opened 2022-02-07
Rework CustomJWTAuthentication to request the oauth token correctly

nclaeys:
Usecase:
I am using oidc on k8s with workload-identity for spark applications and noticed that requesting a token did not work. 

Problem:
The fix for #437 is not sufficient to work with workload identity as the new CustomJwtAuthentication class does not construct the correct body for requesting a token (the applyTo method does nothing).

I use this library as follows:

```
IClientCredential credential = ClientCredentialFactory.createFromClientAssertion(tokenFileString);
ConfidentialClientApplication app = ConfidentialClientApplication.builder(clientId, credential).authority(authority).build();

Set<String> scopes = new HashSet<>();
scopes.add("https://storage.azure.com/.default");

ClientCredentialParameters parameters = ClientCredentialParameters.builder(scopes).tenant(tenantId).build();
IAuthenticationResult token = app.acquireToken(parameters).get(5, TimeUnit.SECONDS);
```

Changes that I made are:
- I updated the customJwtAuthentication to work similarly to how the JwtAuthentication class from com.nimbusds.oauth2.sdk.auth. I think this is useful for other users as well, such that they can also use workload identity with java applications.
- The clientId used, should be the azure ad application id instead of the subject in the jwt token as this is the serviceAccount and thus does not exist in azure ad. 

Note on running all tests:
I had issues running all tests as some depended on libraries I do not have on Linux. To test the relevant classes I created a custom keystore with a self signed certificate in it with a password (minimum was 6 characters). Can you describe the setup required to run all the tests locally?

### DIFF
--- a/src/integrationtest/java/com.microsoft.aad.msal4j/CertificateHelper.java
+++ b/src/integrationtest/java/com.microsoft.aad.msal4j/CertificateHelper.java
@@ -16,6 +16,8 @@ public class CertificateHelper {
         String os = SystemUtils.OS_NAME;
         if (os.contains("Mac")) {
             return KeyStore.getInstance("KeychainStore");
+        } else if (os.contains("Linux")) {
+            return KeyStore.getInstance(KeyStore.getDefaultType());
         } else {
             return KeyStore.getInstance("Windows-MY", "SunMSCAPI");
         }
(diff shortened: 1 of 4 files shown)


siddhijain:
@nclaeys Thanks for putting your time and effort in fixing the code. We highly appreciate it. I have tested the changes and they look good to me. I am not sure of the setup needed to run these tests in Linux since we use a Windows machine. I will update you once I find something. 
I will discuss with my team and wait for them to review the PR before merging this in.
Thanks again.

siddhijain (approved):

PR #479, opened 2022-02-28
Drop jackson-databind dependency

siddhijain:


pom.xml
```diff
@@ -50,9 +60,9 @@
             <scope>provided</scope>
         </dependency>
         <dependency>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-databind</artifactId>
-            <version>2.12.1</version>
+            <groupId>org.json</groupId>
+            <artifactId>json</artifactId>
+            <version>20211205</version>
```
JonathanGiles, line 65: This dependency shouldn't be introduced. We should try to make do without replacing one JSON library with another.

src/integrationtest/java/labapi/App.java
```diff
@@ -29,4 +36,61 @@
...
     @JsonProperty("clientSecret")
     String clientSecret;
+
+    public static App convertJsonToObject(String json) throws IOException {
+
+        if (json != null) {
+            JsonFactory jsonFactory = new JsonFactory();
+            JsonParser jsonParser = jsonFactory.createParser(json);
+
+            if (jsonParser.nextToken().equals(JsonToken.START_ARRAY)) {
```
alzimmermsft, line 46: Should this, an other usages of `JsonToken.START_ARRAY`, in integration tests be checking for `JsonToken.START_OBJECT` instead?

Issue #466, opened 2022-02-07
Add the ability to consume Client Assertion as a Supplier / Callback.


g2vinay:
MSAL .NET uses a callback to consume client assertion.
Align the logic in MSAL4J's Confidential client with .NET's.

This will allow us to control the assertion used per request like MSAL .NET and not have it locked in at Confidential client construction time.

The equivalent API on MSAL .NET that allows to configure the Client Assertion as a callback: 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L213

siddhijain:
@g2vinay Can you elaborate on this? Do you have links to any documentation or MSAL .NET implementation?

g2vinay:
@siddhijain 

I am referring to this API on MSAL .NET: 
The equivalent API on MSAL .NET that allows to configure the Client Assertion as a callback: 
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationBuilder.cs#L213

We need to provide something equivalent to this in msal4j.

Avery-Dunn:
Feature was added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/482 and released as part of [1.11.3](https://github.com/AzureAD/microsoft-authentication-library-for-java/releases/tag/v1.11.3)

Closing this issue, feel free to reopen or leave a comment if anything else is needed.

PR #482, opened 2022-03-24
Allow client assertion to be a callback and a per-request parameter

Avery-Dunn:
Adds a new helper method to the ClientCredentialFactory to create client assertions in a callback, and allows client credentials to be set as a per-request parameter in the client credential flow

As requested in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/466


siddhijain (approved):
Looks good

PR #483, opened 2022-03-24
fix: Correct scope for test dependency in POM

bs-rscs:
The `azure-security-keyvault-secrets` test dependency was include twice and without the correct `test` scope.

Due to this, if used this library, Maven automatically included a lot of dependencies that are actually unneccasary, and in our case, this lead to problems as some of these dependencies clashed with existing ones in our project.


siddhijain (approved):
Thanks for this! It was in our list. I have updated the keyvault-secrets version to reflect the most recent one.

Issue #472, opened 2022-02-11
client_id is missing from request body


yongheyan:
I have the following code:
```java
public class AccessTokenManager {
    private PrivateKey privateKey;
    private X509Certificate publicCert;

    private static final Log log = LogFactory.getLog(AccessTokenManager.class);

    public AccessTokenManager(PrivateKey privateKey, X509Certificate publicCert) {
        this.privateKey = privateKey;
        this.publicCert = publicCert;
    }

    public String retrieveAccessToken(String adfsEndpoint, String clientId, String resourceId) throws Exception {
        log.info("Retrieving Authentication token");

        PublicClientApplication a = PublicClientApplication.builder(clientId)
                .authority(adfsEndpoint)
                .build();

        IClientCredential credential = ClientCredentialFactory.createFromCertificate(privateKey, publicCert);
        ConfidentialClientApplication app = ConfidentialClientApplication
                .builder(clientId, credential)
                .authority(adfsEndpoint)
                .sendX5c(false)
                .build();
        ClientCredentialParameters parameters = ClientCredentialParameters
                .builder(Set.of(String.format("%s/openid", resourceId)))
                .build();

        IAuthenticationResult result = app.acquireToken(parameters).join();

        log.info("Retrieved Authentication token");
        return result.accessToken();
    }
}
```

I tried to acquire token like these:
```java
        AccessTokenManager m = new AccessTokenManager(privateKey, cert);
        String token = m.retrieveAccessToken("https://my.server.com/adfs", "my-client-id","http://myResource" );
```
The server's response is something like "client_id is missing".
I checked the actual httpRequest and the body is like this:
```
client_info=1&client_assertion_type=urn%3Aietf%3Aparams%3Aoauth%3Aclient-assertion-type%3Ajwt-bearer&grant_type=client_credentials&scope=http%3A%2F%2FmyResource%2Fopenid+openid+profile+offline_access&
client_assertion=eyJ4NXQ<---removed--->vxTwXfw
```
The client_id could not be found in the request body. However, according to Microsoft documentation from here: https://docs.microsoft.com/en-us/azure/active-directory/develop/v2-oauth2-client-creds-grant-flow, the request body should contains client_id in the body like this:
```
  POST /{tenant}/oauth2/v2.0/token HTTP/1.1           //Line breaks for clarity
  Host: login.microsoftonline.com
  Content-Type: application/x-www-form-urlencoded
  
[...]

PR #488, opened 2022-04-06
Add client_id to http request

siddhijain:
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/472

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
+++ b/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
@@ -16,11 +16,7 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.net.MalformedURLException;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -80,6 +76,10 @@ OAuthHttpRequest createOauthHttpRequest() throws SerializeException, MalformedUR
                         .createClientAuthFromClientAssertion((ClientAssertion) ((ClientCredentialRequest) msalRequest).parameters.clientCredential())
                         .applyTo(oauthHttpRequest);
             } else {
```
Avery-Dunn, line 78: It looks like client ID would only be added to the request if the request has a client assertion (the main 'if' block, `if (msalRequest.application().clientAuthentication() != null)`), but not if a client assertion was set as a parameter for the request (the nested if/else, `if (msalRequest instanceof ClientCredentialRequest && ... `).

Is some behavior different when a developer wants a to override their client app's assertion for a certain request? If not, then this new code should probably be in the main 'if' block instead of the nested 'if/else' block.

PR #496, opened 2022-04-25
Improve error message when Authority does not contain path

mattmazzola:
I was developing an application using `com.microsoft.azure:msal4j:1.11.3` and came across this `AUTHORITY_URI_EMPTY_PATH_SEGMENT` and thought it was misleading.

image

You can see that there are no "empty" paths in the URL given, so the error message is no appropriate or true. The issue is that the msal4j expects the authority configured to include the path such as "common" or a tenant id.

I added a new constant to explain this case more clearly.

Also, I am new to mal4j but I think it should be more clearly stated in the documentation this library expects different configuration.  In some of the other AAD libraries you configure the authority differently, specifically the path is a separate configuration field. Notably the dotnet library https://github.com/AzureAD/microsoft-identity-web (which is likely more commonly used for Microsoft Auth) constructs the authority from two fields.

Example:
```
{
  "Instance": "https://login.microsoftonline.com/",
  "ClientId": "<removed>",
  "TenantId": "common"
}
```

However, using msal4j it expects the authority to be specified completely like this: `https://login.microsoftonline.com/common`
This difference means this error is more likely to occur and it should have a very clear error message.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -117,7 +117,7 @@ abstract class Authority {
 
         if (segments.length == 0) {
             throw new IllegalArgumentException(
-                    IllegalArgumentExceptionMessages.AUTHORITY_URI_EMPTY_PATH_SEGMENT);
+                    IllegalArgumentExceptionMessages.AUTHORITY_URI_MISSING_PATH_SEGMENT);
         }
 
         for (String segment : segments) {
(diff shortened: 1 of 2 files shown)


siddhijain:
@mattmazzola Is the reason for this PR to improve the error message?

mattmazzola:
Yes. Although it adds a new message it doesn't improve an existing message. I updated the title to use the word "improve" instead of "update" though if you prefer that word. The old message was wrong.

siddhijain:
Thanks for clarifying. It makes sense. I will approve the PR and merge it in once all the checks passed. However, please note that the link pasted above https://github.com/AzureAD/microsoft-identity-web is not the official .NET MSAL but it is identity-web (middleware library) and is different from msal4j. msal4j equivalent in .NET is https://github.com/AzureAD/microsoft-authentication-library-for-dotnet.

Issue #492, opened 2022-04-20
Could not authenticate via Azure Directory Interactive with MFA


luedi:
I want to login to an Azure SQL Server managed instance via Azure Directory Interactive. The account has MFA via Microsoft Authenticator enabled. The login screen (https://uniper-onmicrosoft-com.access.mcas.ms/aad_login) is displayed and I enter my credentials and approve them with the Microsoft authenticator app. After a while i get the response:

`The specified database user/password combination is rejected: Fehler beim Authentifizieren des Benutzers "xxx@yyy" in Active Directory (Authentication=ActiveDirectoryInteractive). No Authorization code was returned from the server`

I'm using Jetbrains DataGrip and am in contact with their support for a while. We have tried many different configuration settings but nothing worked. At least we checked the connection with a simple Java program with no other dependencies. The support guy at Jetbrains could not find any reason for the issue in DataGrip and now guess that there is a problem in the authentication library.

With Sql Server Management Studio login works fine.

I didn't find something related here. Does anyone have an idea why this occurs?

Avery-Dunn:
Hello @luedi : Is there any more to that error message, like an error code, files/line numbers for the package that threw the error, etc.? If so, could you provide it?

Also, could you describe how you're using MSAL Java in your application? You mentioned you're trying to access an Azure SQL server ([this](https://docs.microsoft.com/en-us/azure/azure-sql/database/sql-database-paas-overview), right?), and you're connecting to it using Jetbrains DataGrip but having a problem? And by "With Sql Server Management Studio login works fine", you mean you can connect to the Azure SQL server using Server Management Studio, but not with DataGrip?

luedi:
To be more precise: MFA-login via Azure Active Directory Interactive is possible with SSMS. I can login and work with the database. It's a Java problem. Jetbrains support sent me a testl program which shows the same behavior as DataGrip.

[Test-MS-ActiveDirectoryInteractive-master.zip](https://github.com/AzureAD/microsoft-authentication-library-for-java/files/8533631/Test-MS-ActiveDirectoryInteractive-master.zip)

DataGrip itself uses the follwing files:
![image](https://user-images.githubusercontent.com/12098314/164522417-f430c4f8-8ffb-4a9f-b200-0ba65a9bc875.png)

[...]

PR #511, opened 2022-05-30
Fixed localized strings for htmlMessageSuccess

tkyc:
This PR fixes issue https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/492.

So the error happened because in `send200Response` it was setting the response length using the length of the string. This is incorrect as it should be set using the length of the response string's byte array. We also need to specify/set encoding in the response header. Otherwise by default it will decode using iso_8859_1.


Avery-Dunn (approved):
Sorry for the delay on this. Looks good to me, thanks for investigating and putting this together!

PR #527, opened 2022-08-05
Fix PKCS12 certs not allowed to contain CA certs

agostonbarna:
It is a common scenario that a PKCS12 keystore contains only one private key, but multiple aliases/entries: for example we need to import the CA certs before we could import the renewed cert.
So the idea is to only look for private key type entires in the keystore and do not throw an exception if it also contains CA or other non-private key certs, but only if it contains zero or multiple private keys as those scenarios are indeed invalid.

Note that I had trouble mocking the java KeyStore aliases() method in the unit tests with @PrepareForTest({KeyStore.class}) + EasyMock/PowerMock.createMock even tough PowerMock should support mocking final methods, but I could mock the KeyStoreSpi.
If someone knows a way to mock the KeyStore class with these libraries please let me know and will improve the code, but I can also live with this approach.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/ClientCertificate.java
+++ b/src/main/java/com/microsoft/aad/msal4j/ClientCertificate.java
@@ -3,6 +3,9 @@
 
 package com.microsoft.aad.msal4j;
 
+import lombok.Getter;
+import lombok.experimental.Accessors;
+
 import java.io.IOException;
 import java.io.InputStream;
 import java.lang.reflect.InvocationTargetException;
@@ -19,10 +22,11 @@ import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
 import java.security.cert.X509Certificate;
 import java.security.interfaces.RSAPrivateKey;
-import java.util.*;
-
-import lombok.Getter;
-import lombok.experimental.Accessors;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.Enumeration;
+import java.util.List;
 
 final class ClientCertificate implements IClientCertificate {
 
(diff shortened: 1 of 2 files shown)


Avery-Dunn:
Thanks for making this PR! We'll get it reviewed and tested, and it should be in our next release.

agostonbarna:
Hi,
Any update on this?
Thanks,
Barna

Avery-Dunn:
> Hi, Any update on this? Thanks, Barna

Hello, sorry for the delay on this, it's been a busy couple of months and this fell through the cracks. Before we add this to our library, I just wanted to confirm what the intent here is:
- In the current version of our library, we're defining a PKCS12 keystore and loading an InputStream (assumed to be a PKCS12 certificate) into it. Then, we grab the first alias in it, and treat it as a PrivateKey when creating the client certificate. If there are 0 or multiple aliases, we thrown an exception
- In your version, we set up the keystore in the same way. However, rather than grabbing the first alias, we grab the first alias that is a PrivateKeyEntry, and treat that as a PrivateKey when creating the client certificate. If there are 0 or multiple PrivateKeyEntry aliases, we throw an exception

In short, it changes our "only 1 alias" check into an "only 1 alias that is a PrivateKeyEntry" check. Am I understanding it right?

agostonbarna:
Hey @Avery-Dunn 
Thanks for the update. I can relate to that :)
You are correct about the change, I couldn't have described it better.
And the motivation is to make it possible use the MSAL library with the PKCS12 keystore even if it contains a CA (or any non-private key) cert beside the private key cert which is the case for all our keystore files.
Thanks,
Barna

Avery-Dunn:
Hey @agostonbarna, we've reviewed the PR and it all looks good, these changes should get into our next release, which we expect to have done today. 

However, since you made this PR our library when through some file restructuring due to other work, and just to make the merging easier I've made a new PR (https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/549) that has the exact same changes in it. Feel free to review that, and once it's approved I'll merge it in and close the PR when the release is done.

agostonbarna:
Thanks @Avery-Dunn for aligning it with the new structure and doing the merge :)

Avery-Dunn:
Closing this PR since these changes made it into our recent 1.13.2 release via https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/549. Thanks for the contribution, and if you have any other improvements in mind feel free to open up new issues/PRs.

PR #530, opened 2022-08-31
MSALRuntime integration

Avery-Dunn:
Draft PR for the work-in-progress MSALRuntime integration. Most of the new files will be split off into either a new Broker repository or the existing MSAL C++ repository, and are just here for easier initial code review

This PR currently has:
- IBroker.java an interface used to reference methods implemented by a broker package, such as our upcoming WAMBroker that allows using WAM via MSALRuntime
- Changes in some existing flows to demonstrate the use of IBroker
- MSALRuntimeLibrary.java, which will be used to make the actual calls to the MSALRuntime API and will be placed in the MSAL C++ repo (where it will be known as the 'interop layer', and the PR for it is: https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3091)


src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,21 @@
+package com.microsoft.aad.msal4j;
+
+/**
+ * Used to define the basic set of methods that all Brokers must implement
+ */
+public interface IBroker {
```
JonathanGiles, line 6: The I prefix is quite an outdated approach. I would suggest you consider calling this `Broker` instead.

src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,21 @@
...
+    Account signInInteractively(PublicClientApplication clientApplication, InteractiveRequestParameters interactiveRequestParameters);
+
+    AuthenticationResult acquireTokenSilently(PublicClientApplication clientApplication, InteractiveRequestParameters interactiveRequestParameters, Account account);
+
+    AuthenticationResult acquireTokenInteractively(PublicClientApplication clientApplication, InteractiveRequestParameters interactiveRequestParameters, Account account);
+
+    void signOutSilently(PublicClientApplication clientApplication, InteractiveRequestParameters interactiveRequestParameters);
+
+    //TODO: too specific to MSALRuntime? Should it be split into a generic getAccount(something) and getAllAccounts(clientId)?
+    void discoverAccounts(PublicClientApplication clientApplication, InteractiveRequestParameters interactiveRequestParameters);
```
JonathanGiles, line 20: Because all methods seem to require a `PublicClientApplication`, I wonder if this should be a constructor parameter in implementations of this interface? Do you expect for an instance of `Broker` that the `PublicClientApplication` will always be the same?

src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -34,9 +34,35 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
 
     @Override
     AuthenticationResult execute() throws Exception {
-        AuthorizationResult authorizationResult = getAuthorizationResult();
-        validateState(authorizationResult);
-        return acquireTokenWithAuthorizationCode(authorizationResult);
+        if (clientApplication.allowBroker()) {
```
JonathanGiles, line 37: We should have boolean getters prefixed with `is`, so perhaps this method should be named `isBrokerAllowed()` or similar?

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
```diff
@@ -0,0 +1,112 @@
...
+@Structure.FieldOrder({"authParamsHandle", "accountHandle", "authResultHandle", "idToken", "accessToken", "accessTokenExpirationTime", "accountId", "accountClientInfo"})
+public class MSALRuntimeAuthResult extends Structure implements MSALRuntimeLibrary.AuthResultCallbackData {
+    public LongByReference authParamsHandle;
+    public LongByReference accountHandle;
+    public LongByReference authResultHandle;
+    public MSALRuntimeUtils.StringByReference idToken = new MSALRuntimeUtils.StringByReference(new WString(""));
+    public MSALRuntimeUtils.StringByReference accessToken = new MSALRuntimeUtils.StringByReference(new WString(""));
+    public LongByReference accessTokenExpirationTime;
+    public MSALRuntimeUtils.StringByReference accountId = new MSALRuntimeUtils.StringByReference(new WString(""));//TODO: split AuthenticationResult into SignInResult (id token/account) and TokenResult (access tokens)? Or keep it simple all-in-one like MSAL Java's AuthenticationResult?
+    public MSALRuntimeUtils.StringByReference accountClientInfo = new MSALRuntimeUtils.StringByReference(new WString(""));
```
JonathanGiles, line 20: We shouldn't have public instance fields in our APIs, if at all possible

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
```diff
@@ -0,0 +1,112 @@
...
+    private MSALRuntimeLibrary MSALRuntime_INSTANCE;
+    private MSALRuntimeError MSALRuntimeErrorHandler;
+
+    public void setMSALRuntimeLibraryInstance (MSALRuntimeLibrary MSALRuntime_INSTANCE) {
```
JonathanGiles, line 24: Is it convention in this library to upper case MSAL? In the Azure SDKs we would likely call this method `setMsalRuntimeLibraryInstance`, so we would only upper-case the M in MSAL.

JonathanGiles: Also, the parameter name here (and below) is weird - can we just call it `instance`?

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
```diff
@@ -0,0 +1,112 @@
...
+        this.MSALRuntime_INSTANCE = MSALRuntime_INSTANCE;
+    }
+
+    public void setMSALRuntimeErrorHandler (MSALRuntimeLibrary MSALRuntime_INSTANCE) {
+        this.MSALRuntimeErrorHandler = new MSALRuntimeError();
+        this.MSALRuntimeErrorHandler.MSALRuntime_INSTANCE = MSALRuntime_INSTANCE;
+    }
```
JonathanGiles, line 31: This method doesn't do what the name implies?

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
```diff
@@ -0,0 +1,112 @@
...
+        System.out.println("===Setting parameters");
+        MSALRuntimeErrorHandler.printAndReleaseError(MSALRuntime_INSTANCE.MSALRUNTIME_SetRedirectUri(authParamsHandle.getValue(), new WString("https://login.microsoftonline.com/common/oauth2/nativeclient")));
+        MSALRuntimeErrorHandler.printAndReleaseError(MSALRuntime_INSTANCE.MSALRUNTIME_SetRequestedScopes(authParamsHandle.getValue(), new WString(scopes)));
+        MSALRuntimeErrorHandler.printAndReleaseError(MSALRuntime_INSTANCE.MSALRUNTIME_SetRedirectUri(authParamsHandle.getValue(), new WString(redirectUri)));
+        if (claims != null) {
+            MSALRuntimeErrorHandler.printAndReleaseError(MSALRuntime_INSTANCE.MSALRUNTIME_SetDecodedClaims(authParamsHandle.getValue(), new WString(claims)));
+        }
+    }
+
+    public void parseAndReleaseAuthResult_SignInSilently() {
```
JonathanGiles, line 50: I'm confused why a class named *Result has API to parseAndReleaseAuthResult_SignInSilently as public API? Same goes for a lot of the other API in this class.

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeUtils.java
```diff
@@ -0,0 +1,65 @@
...
+        //TODO: Should we allow customer to change it? Any use case where they want a different mdsalruntime.dll
+        if (System.getProperty("jna.library.path") != null) {
+            System.setProperty("jna.library.path", System.getProperty("user.dir"));
+        }
+    }
+
+    static MSALRuntimeLibrary setMSALRuntimeLibrary(){
+        //Will look for dll at jna.library.path, plus some other common dll/Java paths specified by JNA
+        return Native.load("msalruntime.dll", MSALRuntimeLibrary.class);
+    }
```
JonathanGiles, line 64: Even though this isn't public API, it is weird that a set method is actually not setting anything, and instead returning an instance.

src/main/java/com/microsoft/aad/msal4j/IBroker.java
JonathanGiles: I would argue to not follow conventions of another language, if you can help it.

Avery-Dunn: I also agree that it's outdated, but it's also something that all the other interfaces in our library follow. Consistency with the other MSALs is also a big focus for this feature, since most of the libraries will be also be integrating with MSALRuntime we want to keep things as similar as we can

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
Avery-Dunn: These fields (and the class itself) need to be public due to how JNA handles converting Java classes to C structs. Basically, JNA will ignore any field that isn't public when creating the native struct, so anything on the Java side tried to access native structs JNA is handling you'd get an error about the struct not having a field named XYZ.

src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
Avery-Dunn: This is just the flag decided on in the [design doc](https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/%5BXplat%5D%20MSALRuntime/Windows%20MSALRuntime/IntegrationWithMsalRuntime.md&version=GCd5b0360363a21b7c812d6303c911f1c420117dcc&_a=preview) (mainly according to conventions in Python and .NET since they were the frontrunners on this). Normally I'd agree that it's weird and change it to something like 'isBrokerAllowed', but consistency across MSALs is a focus on this project, and we other parameters like this in our library already so it shouldn't be too confusing.

src/main/java/com/microsoft/aad/msal4j/IBroker.java
Avery-Dunn: This has been completely refactored from this initial generic design to something more like the rest of MSAL Java, where there are a series of acquireToken() methods overloaded for each type of request (leaving it up to a broker to figure out how to handle each type of request).

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
Avery-Dunn: It did in an earlier version, but I missed renaming it to something like 'initializeErrorHandler'. Regardless, this method has  been removed entirely as the error handler class has been refactored to just a helper class with static methods.

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeUtils.java
Avery-Dunn: After some refactoring this initialization has been moved to the main WAMBroker class, and renamed to hopefully avoid confusion.

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
Avery-Dunn: As I mentioned in another comment, because of how JNA handles conversion between Java and C/C++, any of the fields in this class that will be referenced in a native context need to be public. To be safe, I just made all of the methods public as well to avoid exceptions during development, but will revisit the scoping once we get some more thorough tests done.

src/main/java/com/microsoft/aad/msal4j/MSALRuntimeAuthResult.java
Avery-Dunn: Agreed, and in the latest commit I refactored these sorts of variables to be things like 'setMsalRuntime...' and 'msalRuntimeLibraryInstance'. When initially naming these variables I just followed how I'd normally write MSAL, and was following some conventions I saw in the JNA documentation and MSALRuntime code whenever I did something that referenced something on the C++ side of things.

Avery-Dunn:
Enough design changes have happened to make this version obsolete, new PR IBroker and it's implementation for MSALRuntime can be found here: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/563

PR #533, opened 2022-09-13
Configurable timeout for InteractiveRequestParameters

### LINKED ISSUE #532: Polling Timeout for Login should be configurable
Hi! Looking at [msal.js](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/client/DeviceCodeClient.ts) and msal dotnet, I noticed that they have configurable timeouts for the user to login using the browser prompt. 

For our scenario, we are using Azure AD B2C. Including a first-time sign up, users can easily exceed the [hardcoded 120s timeout](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java#L128) period. 

I will submit a PR to patch this by allowing a configurable timeout if the issue is acceptable.

eginwong:
Current polling timeout for login is hardcoded to 120s. This change allows the timeout to be configured via the InteractiveRequestParametersBuilder. Fixes #532.

Looking for suggestions on how to test this code.

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
@@ -125,12 +125,15 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
         AuthorizationResult result = null;
         try {
             LOG.debug("Listening for authorization result");
-            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 120;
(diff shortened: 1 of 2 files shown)


Avery-Dunn:
Thanks for making this PR! However, having the timeout be 'infinite' by default instead of a hardcoded 120s would likely be a breaking change for some customers, as they may rely on the current timeout behavior or could start experiencing infinite wait times without realizing why.

Making that timeout configurable would definitely be useful however, so I've made a PR (https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/536) that allows it to be configurable, but maintains the current behavior by default.

Issue #531, opened 2022-09-02
MSAL should not perform Instance Discovery if azureRegion is set


siddhijain:
Some of the apps are flagged in S360 KPI for not using regional endpoints. On debugging, it was found that msal makes call to a global endpoint for instance discovery even when app specifies a region. Confirmed with .NET team that if a region is defined, instance discovery should not be done instead the method call can return 

return new InstanceDiscoveryMetadataEntry()
            {
                Aliases = new[] { regionalEnv, originalEnv },
                PreferredCache = originalEnv,
                PreferredNetwork = regionalEnv
            };

https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PinAuthToRegion/AAD%20SDK%20Proposal%20to%20Pin%20Auth%20to%20region.md&_a=preview&anchor=instance-discovery-and-authority-validation

siddhijain:
Released the bug fix in the latest Release 1.13.1.. Closing the issue.

PR #535, opened 2022-09-14
Improvements and bug fixes for regional endpoint support

Avery-Dunn:
Fixes issue in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/531, and includes a number of other fixes and improvements to regional endpoint support:
- No longer makes unnecessary instance discovery call when using the region API. Now, instance metadata is formatted and cached by the library according to URL formats expected by ESTS-R, so that the original request and any future requests correctly use regional endpoints
- Although the number of supported scenarios is growing, currently ESTS-R only supports a limited number of auth flows, however most failures when using an unsupported flow were either silent or unclear. Exception handling, logging, and javadocs have been improved to be more clear about failures, along with with links to more information

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -141,7 +141,7 @@ class AadInstanceDiscoveryProvider {
             } else {
                 //Avoid unnecessary warnings when looking for cached tokens by checking if request was a silent call
                 if (msalRequest.getClass() != SilentRequest.class) {
-                    log.warn("Regional endpoints are only available for client credential flow. Request will fall back to using the global endpoint.");
+                    log.warn("Regional endpoints are only available for client credential flow, request will fall back to using the global endpoint. See here for more information about supported scenarios: https://aka.ms/msal4j-azure-regions");
                 }
                 return false;
             }
--- a/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -3,14 +3,13 @@
 
 package com.microsoft.aad.msal4j;
 
-import com.nimbusds.oauth2.sdk.http.HTTPResponse;
-
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import java.net.URL;
 import java.util.Arrays;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.Map;
@@ -22,9 +21,9 @@ class AadInstanceDiscoveryProvider {
     private final static String DEFAULT_TRUSTED_HOST = "login.microsoftonline.com";
     private final static String AUTHORIZE_ENDPOINT_TEMPLATE = "https://{host}/{tenant}/oauth2/v2.0/authorize";
     private final static String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE = "https://{host}:{port}/common/discovery/instance";
-    private final static String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE_WITH_REGION = "https://{region}.r.{host}:{port}/common/discovery/instance";
-    private final static String INSTANCE_DISCOVERY_SOVEREIGN_ENDPOINT_TEMPLATE_WITH_REGION = "https://{region}.{host}:{port}/common/discovery/instance";
     private final static String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
+    private final static String HOST_TEMPLATE_WITH_REGION = "{region}.r.{host}";
+    private final static String SOVEREIGN_HOST_TEMPLATE_WITH_REGION = "{region}.{host}";
     private final static String REGION_NAME = "REGION_NAME";
     private final static int PORT_NOT_SET = -1;
     // For information of the current api-version refer: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#versioning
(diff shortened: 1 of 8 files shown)


src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -56,14 +55,35 @@ static InstanceDiscoveryMetadataEntry getMetadataEntry(URL authorityUrl,
                                                            boolean validateAuthority,
                                                            MsalRequest msalRequest,
                                                            ServiceBundle serviceBundle) {
+        String host = authorityUrl.getHost();
+
+        if (useRegionalEndpoint(msalRequest)) {
```
siddhijain, line 60: nit - consider naming it to 'shouldUseRegionalEndpoint' / 'hasRegionalEndpoint' / 'isRegional' to signify that the method returns boolean.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -108,6 +128,68 @@ static void cacheInstanceDiscoveryMetadata(String host,
                 build());
     }
 
+
+    private static boolean useRegionalEndpoint(MsalRequest msalRequest){
+        if (msalRequest.application().azureRegion() != null || msalRequest.application().autoDetectRegion()){
+            //This class type check is a quick and dirty fix to accommodate changes to the internal workings of the region API
+            //
+            //ESTS-R only supports a small, but growing, number of scenarios, and the original design failed silently whenever
```
siddhijain, line 136: Nice detailed comments!

src/main/java/com/microsoft/aad/msal4j/IAuthenticationResult.java
```diff
@@ -44,4 +44,15 @@ public interface IAuthenticationResult extends Serializable {
...
+    /**
+     * The source of the tokens in a result. Will simply be the word "cache" if retrieved from the in-memory token cache,
+     * otherwise it will be the actual authority used by the library for the token request
+     *
+     * In general an authority value here will match the authority configured in the client app or acquire token request,
+     * however there are certain cases, such as when using regional endpoints, where the library may use a different authority
+     *
+     * @return source of tokens in result
+     */
+    String tokenSource();
```
siddhijain, line 57: Is this newly introduced as part of the regional api design doc?

Avery-Dunn: This wasn't part of the region design doc, it's just a little info field to give customers some better feedback. When troubleshooting this issue with a customer there was confusion about where exactly the error was happening because we didn't have an easy way to see exactly where the library was sending requests (the tokens don't contain any regional info, and we were caching the global info regardless). They pointed out [this field](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/b23bbce16567739e895185a0ffb7dcc25d55663e/src/client/Microsoft.Identity.Client/AuthenticationResult.cs#L162) in .NET's AuthenticationResult and asked if we had anything similar.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -56,14 +55,35 @@ static InstanceDiscoveryMetadataEntry getMetadataEntry(URL authorityUrl,
...
                                                            MsalRequest msalRequest,
                                                            ServiceBundle serviceBundle) {
+        String host = authorityUrl.getHost();
+
+        if (shouldUseRegionalEndpoint(msalRequest)) {
+            //Server side telemetry requires the result from region discovery when any part of the region API is used
+            String detectedRegion = discoverRegion(msalRequest, serviceBundle);
+
+            if (msalRequest.application().azureRegion() != null) {
+                host = getRegionalizedHost(authorityUrl.getHost(), msalRequest.application().azureRegion());
```
jennyf19, line 65: nit: use the `host` variable here for `authorityUrl.getHost()`

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -108,6 +128,68 @@ static void cacheInstanceDiscoveryMetadata(String host,
...
+            //
+            //ESTS-R only supports a small, but growing, number of scenarios, and the original design failed silently whenever
+            //  regions could not be used. To avoid a breaking change this check will allow supported flows to use regions,
+            //  and unsupported flows will continue to fall back to global, but with an added warning and a link to more info
+            if (msalRequest.getClass() == ClientCredentialRequest.class) {
+                return true;
+            } else {
+                //Avoid unnecessary warnings when looking for cached tokens by checking if request was a silent call
+                if (msalRequest.getClass() != SilentRequest.class) {
+                    log.warn("Regional endpoints are only available for client credential flow, request will fall back to using the global endpoint. See here for more information about supported scenarios: https://aka.ms/msal4j-azure-regions");
```
jennyf19, line 144: +1 to aka.ms links and actionable log messages! :)

jennyf19 (approved):
:shipit:

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: Changed to shouldUseRegionalEndpoint in the latest commit

src/main/java/com/microsoft/aad/msal4j/IAuthenticationResult.java
Avery-Dunn: Discussed this, and decided to not add this field to AuthenticationResult. Instead, if an app developer wants this information a new debug-level log was added to TokenRequestExecutor which will print out exactly where the request is going.

src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
Avery-Dunn: Good catch, I missed that in a refactoring. Updated in the latest commit to just be 'host'

siddhijain (approved):
LGTM

PR #536, opened 2022-09-15
Allow interactive request timeout to be configurable

Avery-Dunn:
Add a new field `httpPollingTimeoutInSeconds` to `InteractiveRequestParameters` to allow an app developer to override the timeout for receiving an auth result. Previously the timeout was hardcoded to 120 seconds, but 120 seconds is now just the default. Also adjusts the logging to include the timeout value.

This addresses the request from https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/532, and adds the same core functionality as in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/533

### DIFF
--- a/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
+++ b/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
@@ -124,8 +124,11 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
     private AuthorizationResult getAuthorizationResultFromHttpListener() {
         AuthorizationResult result = null;
         try {
-            LOG.debug("Listening for authorization result");
-            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 120;
+            LOG.debug(String.format("Listening for authorization result. Listener will timeout after %S seconds.",
+                    interactiveRequest.interactiveRequestParameters().httpPollingTimeoutInSeconds()));
+
+            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) +
+                    interactiveRequest.interactiveRequestParameters().httpPollingTimeoutInSeconds();
 
             while (result == null && !interactiveRequest.futureReference().get().isCancelled() &&
                     TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) < expirationTime) {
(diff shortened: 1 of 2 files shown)


src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -124,8 +124,11 @@ private void openDefaultSystemBrowser(URL url) {
     private AuthorizationResult getAuthorizationResultFromHttpListener() {
         AuthorizationResult result = null;
         try {
-            LOG.debug("Listening for authorization result");
-            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 120;
+            LOG.debug(String.format("Listening for authorization result. Listener will timeout after %S seconds.",
+                    interactiveRequest.interactiveRequestParameters().httpPollingTimeoutInSeconds()));
+
+            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) +
```
siddhijain, line 130: Before using the httpPollingTimeout from the request, can we check if it contains a valid finite value?

Avery-Dunn: I manually tested both 0 and negative values and it seemed to just timeout immediately, and [the javadocs for the poll method](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/BlockingQueue.html#poll(long,%20java.util.concurrent.TimeUnit)) just say it will wait "up to the specified wait time if necessary" so it seems like their valid inputs.

However, having it wait a 0 or negative amount of time doesn't really make sense and would probably a race condition over whether or not the auth result is returned before the poll method reaches its logic for checking the wait time. 

I can't think of a use case where someone would want that behavior, so in the latest commit I enforced a 1 second minimum, with javadocs and a new log.warn to explain it.

siddhijain: That's a good find.

siddhijain (approved):

src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
eginwong: Hi there, I wanted to make sure we account for the option where we want to set an infinite timeout. Can the code be amended to allow for that option?

Avery-Dunn: I don't think we need to add something to specifically handle if someone wants the time out to be 'infinite'.

The new `httpPollingTimeoutInSeconds` field is an int, so you can set a timeout of up to 2147483647 seconds, or roughly 70 years. Even a much more reasonable timeout value of 86400 seconds would be a full 24 hours.

PR #540, opened 2022-09-20
Sjain/restructure library

siddhijain:


msal4j-sdk/pom.xml
```diff
@@ -0,0 +1,276 @@
...
+        <artifactId>com.microsoft.aad.msal4j</artifactId>
+        <groupId>com.microsoft.azure</groupId>
+        <version>1.13.1</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>com.microsoft.aad.msal4j-sdk</artifactId>
+
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
```
Avery-Dunn, line 15: I don't think we were specifying this in our current pom.xml, but this should probably be 1.8 or 8 instead of 11

siddhijain: sounds good. I removed it altogether.

Avery-Dunn (approved):

PR #541, opened 2022-09-21
Moving IBroker interface file from Avery's PR to this one.

siddhijain:
Moving IBroker interface file to this PR

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
@@ -115,4 +115,12 @@ public class AuthenticationErrorCode {
      * A JWT parsing failure, indicating the JWT provided to MSAL is of invalid format.
      */
     public final static String INVALID_JWT = "invalid_jwt";
+    /**
+     * Indicates that a Broker implementation is missing from the device, such as when an app developer
+     * does not include one of our broker packages as a dependency in their project, or otherwise cannot
+     * be accessed by MSAL Java
+     *
+     * TODO: link to repository wiki page describing this more and how to fix
+     */
+    public final static String MISSING_BROKER = "missing_broker";
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
@@ -118,9 +118,6 @@ public class AuthenticationErrorCode {
     /**
      * Indicates that a Broker implementation is missing from the device, such as when an app developer
      * does not include one of our broker packages as a dependency in their project, or otherwise cannot
-     * be accessed by MSAL Java
-     *
-     * TODO: link to repository wiki page describing this more and how to fix
-     */
+     * be accessed by MSAL Java*/
     public final static String MISSING_BROKER = "missing_broker";
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
@@ -15,15 +15,10 @@ import java.util.concurrent.CompletableFuture;
 public interface IBroker {
 
     /**
-     * Called when the 'useBroker' flag is first set, and will perform any initialization needed to make the broker available for future calls
-     *
-     * Will also detect dependency or accessibility errors with the Broker package, allowing an exception to provide the error to the app developer
-     * before any actual calls to a broker are made
+     * checks if a IBroker implementation exists
      */
-    default void initializeBroker(){
-        throw new MsalClientException("Broker implementation missing", AuthenticationErrorCode.MISSING_BROKER);
-    }
 
+    boolean isAvailable;
     /**
      * Acquire a token silently, i.e. without direct user interaction
      *
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
@@ -18,7 +18,7 @@ public interface IBroker {
      * checks if a IBroker implementation exists
      */
 
-    default boolean isAvailable{
+    default boolean isAvailable(){
         return false;
     }
     /**
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
@@ -1,0 +1,85 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+import java.util.Set;
+
+/**
+ * Used to define the basic set of methods that all Brokers must implement
+ *
+ * All methods are so they can be referenced by MSAL Java without an implementation, and by default simply throw an
+ * exception saying that a broker implementation is missing
+ */
+public interface IBroker {
+
+    /**
+     * Called when the 'useBroker' flag is first set, and will perform any initialization needed to make the broker available for future calls
+     *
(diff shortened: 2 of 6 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,85 @@
...
+
+import java.util.Set;
+
+/**
+ * Used to define the basic set of methods that all Brokers must implement
+ *
+ * All methods are so they can be referenced by MSAL Java without an implementation, and by default simply throw an
+ * exception saying that a broker implementation is missing
+ */
+public interface IBroker {
```
bgavrilMS, line 14: You are missing a `RemoveAccount` functionality.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * Acquire a token silently, i.e. without direct user interaction, using a refresh token
+     *
+     * This requires the broker implementation to cache the refresh tokens, either in its own cache or by populating MSAL Java's cache,
+     * or otherwise have some way of retrieving refresh tokens cached somewhere else
+     *
+     * @param requestParameters MsalRequest object which contains everything needed for the broker implementation to make a request
+     * @return IBroker implementations will return an AuthenticationResult object
+     */
+    default IAuthenticationResult acquireToken(PublicClientApplication application, RefreshTokenParameters requestParameters) {
```
bgavrilMS, line 48: This is not a supported WAM operation. WAM only supports `SilentAuth` as defined by the operation above. Bypassing the msal_runtime's token cache is not supported.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
```diff
@@ -115,4 +115,12 @@ public class AuthenticationErrorCode {
      * A JWT parsing failure, indicating the JWT provided to MSAL is of invalid format.
      */
     public final static String INVALID_JWT = "invalid_jwt";
+    /**
+     * Indicates that a Broker implementation is missing from the device, such as when an app developer
+     * does not include one of our broker packages as a dependency in their project, or otherwise cannot
+     * be accessed by MSAL Java
+     *
+     * TODO: link to repository wiki page describing this more and how to fix
```
bgavrilMS, line 123: TODO should be resolved

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+public interface IBroker {
+
+    /**
+     * Called when the 'useBroker' flag is first set, and will perform any initialization needed to make the broker available for future calls
+     *
+     * Will also detect dependency or accessibility errors with the Broker package, allowing an exception to provide the error to the app developer
+     * before any actual calls to a broker are made
+     */
+    default void initializeBroker(){
```
bgavrilMS, line 22: Generally you don't need to put `Init` methods on interfaces. Implementations will use a ctor which can achieve this. 

In MSAL we expose a `IsInstalled` method, because there are cases where the broker is not installed. On Windows, this older versions like Win8 and Win Server 2016.

MSAL's logic is: 


```
Msal.AcquireToken(...) {
    if (broker.IsConfigured && broker.IsInstalled)  {
         return broker.AcquireToken();
    }
    
    return cache_or_aad.AcquireToken();
}
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,85 @@
...
+    default IAuthenticationResult acquireToken(PublicClientApplication application, UserNamePasswordParameters requestParameters) {
+        throw new MsalClientException("Broker implementation missing", AuthenticationErrorCode.MISSING_BROKER);
+    }
+
+    /**
+     * Retrieve an account information from the broker based on an account ID
+     *
+     * @return IAccount implementations will return an Account object
+     */
+    default Account getAccount(String id) {
```
bgavrilMS, line 77: For now, I suggest you don't add `getAccount` / `getAccounts` methods because they are no-op.

You'll just have to rely on your own cache, i.e. when a token is obtained, you need to persist the id token and account metadata in your own cache.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
siddhijain: Makes sense. Shouldn't these booleans be defined in IBroker interface? In .net IBroker interface, I see only 1 boolean IsBrokerInstalledAndInvokable. Is that the one to be used here? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Broker/IBroker.cs

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
siddhijain: Added.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
siddhijain: So 3 variables in IBroker interface - isAvailable, isConfigured and isInstalled? Are some of these redundant?

bgavrilMS: No, just one, `IsAvailable` which means: 

- on Windows, it is a version of windows which has the broker 
- on Mac and Linux (which we don't integrate to right now), it'll mean "broker is installed"

"IsConfigured" is a setting on the PublicClientApplication object.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
siddhijain: Removed TODO. Added a task in ADO to take care of this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,60 @@
...
+     * Acquire a token silently, i.e. without direct user interaction
+     *
+     * This may be accomplished by returning tokens from a token cache, using cached refresh tokens to get new tokens,
+     * or via any authentication flow where a user is not prompted to enter credentials
+     *
+     * @param requestParameters MsalRequest object which contains everything needed for the broker implementation to make a request
+     * @return IBroker implementations will return an AuthenticationResult object
+     */
+    default IAuthenticationResult acquireToken(PublicClientApplication application, SilentParameters requestParameters) {
+        throw new MsalClientException("Broker implementation missing", AuthenticationErrorCode.MISSING_BROKER);
```
bgavrilMS, line 34: Consider adding aka.ms link and providing more info about how to fix this error, i.e. "install this jar file, call this API"

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * Acquire a token interactively, by prompting users to enter their credentials in some way
+     *
+     * @param requestParameters MsalRequest object which contains everything needed for the broker implementation to make a request
+     * @return IBroker implementations will return an AuthenticationResult object
+     */
+    default IAuthenticationResult acquireToken(PublicClientApplication application, InteractiveRequestParameters requestParameters) {
```
bgavrilMS, line 43: Are these all async?

siddhijain: I believe all the acquireToken calls as async by default in MSAL Java. I will double check.

siddhijain: I have made a note to ensure these calls are in sync with existing acquireToken implementations in MSAL Java.

Issue #542, opened 2022-09-21
Update the Logging Level to VERBOSE (Msal Default) for Token Caching Callback API


g2vinay:
Currently, ERROR levels logs are logged for Exception in Token Caching Callback API,
<img width="1398" alt="image" src="https://user-images.githubusercontent.com/5430778/191592872-09b3300f-b99e-4303-a3b6-c199db84b63a.png">

It is expected to follow the default logging level of VERBOSE as discussed with @siddhijain

siddhijain:
This issue is fixed in Release v1.13.2. Closing the issue.

PR #547, opened 2022-10-05
App token provider change log level

siddhijain:
Addresses https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/542

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalAzureSDKException.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalAzureSDKException.java
@@ -1,0 +1,14 @@
+package com.microsoft.aad.msal4j;
+
+/**
+ * Exception type thrown when Azure SDK returns an error response.
+ */
+public class MsalAzureSDKException extends MsalException{
+    public MsalAzureSDKException(Throwable throwable) {
+        super(throwable);
+    }
+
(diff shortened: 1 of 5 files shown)


Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalAzureSDKException.java
```diff
@@ -0,0 +1,14 @@
+package com.microsoft.aad.msal4j;
+
+/**
+ * Exception type thrown when Azure SDK returns an error response.
+ */
+public class MsalAzureSDKException extends MsalException{
```
Avery-Dunn, line 6: This whole AppTokenProvider flow is intended for the Azure SDK, but in theory anyone who wants to implement their own auth logic could use it, right? 

If so, then maybe this should just be something like 'TokenProviderException'. On the other hand, maybe it's better to leave this specific to Azure SDK to discourage other people from trying to use it.

Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalAzureSDKException.java
siddhijain: That's a good point. This functionality is meant to be used by the Azure SDK team only. I have added documentation to discourage any other teams from using it.

PR #553, opened 2022-10-07
Replace 'compile' dependency with 'implementation'

mpdelbuono:
The 'compile' dependency is deprecated in Gradle. Modern code should use 'implementation' dependency instead.


Avery-Dunn (approved):
Looks good to me, thanks for updating this!

PR #557, opened 2022-10-10
plugins for GPG signing

siddhijain:
This PR makes following changes - 
1. include dependencies in broker pom to build javadoc and sources jars.
2. add gitignore to msal4j-broker module.
3. Added 1 dummy java file to msal4j-broker module for javadoc to successfully run.

### DIFF
--- a/msal4j-brokers/pom.xml
+++ b/msal4j-brokers/pom.xml
@@ -14,11 +14,40 @@
                 <artifactId>msal4j</artifactId>
                 <version>1.13.1</version>
             </dependency>
-            <dependency>
-                <groupId>net.java.dev.jna</groupId>
-                <artifactId>jna-platform</artifactId>
-                <version>5.12.1</version>
-            </dependency>
     </dependencies>
+
+    <build>
+        <plugins>
+        <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-javadoc-plugin</artifactId>
+            <version>3.1.0</version>
+            <configuration>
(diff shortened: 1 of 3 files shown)


msal4j-brokers/pom.xml
```diff
@@ -3,22 +3,135 @@
...
+    <name>msal4j-brokers</name>
+    <description>
+        Microsoft Authentication Library for Java - Brokers helps you integrate with the broker
+        on windows machine to secure Access tokens and refresh tokens.
+    </description>
+    <url>https://github.com/AzureAD/microsoft-authentication-library-for-java</url>
+    <developers>
+        <developer>
+            <id>msopentech</id>
+            <name>Microsoft Open Technologies, Inc.</name>
```
Avery-Dunn: It was mentioned in a comment on your PR for the build script in the interop layer that 'Microsoft Open Technologies, Inc.' seems like a weird thing to put here, and honestly I'm not sure why this is even in our pom.xml for MSAL Java.

The <developer> tag is meant to be an optional documentation thing to just describe the developers, and I think we can just remove it from our pom's since it's not really necessary: https://maven.apache.org/pom.html#Developers

Avery-Dunn (approved):

msal4j-brokers/pom.xml
siddhijain: Good point. It is interesting that some of the tags are optional but were giving me errors if not included. I dont remember if developer tag was one of them but to make it work, I just included everything from msal java pom xml. Lets talk and research more on why it was included in the first place before removing them. If they can be removed, we will remove them from both the modules.

PR #563, opened 2022-10-23
Add IBroker implementation for MSALRuntime

Avery-Dunn:
Draft PR for the work-in-progress MSALRuntime integration. This PR covers
- The IBroker interface, which defines a set of acquireToken methods that are to be implemented in the msal4jbrokers package
- An implementation of IBroker that uses the javamsalruntime package (related PR here: https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3091)
- Changes to MSAL Java's acquireToken flows to allow use of an IBroker implementation (in next commit)
- Exception handling to covert errors from an IBroker implementation into MSAL Java-style exceptions and error messages (in next commit)


msal4j-brokers/pom.xml
```diff
@@ -30,7 +30,12 @@
         <dependency>
             <groupId>com.microsoft.azure</groupId>
             <artifactId>msal4j</artifactId>
-            <version>1.13.2</version>
+            <version>1.13.3</version>
+        </dependency>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>javamsalruntime</artifactId>
+            <version>1.0.0</version>
```
siddhijain: Commenting to ensure we correct the version number here once interop's beta version is published to maven.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -0,0 +1,192 @@
...
+public class MsalRuntimeBroker implements IBroker {
+    private static MsalRuntimeInterop interop;
+
+    static {
+        try {
+            //MsalRuntimeInterop performs various initialization steps in a similar static block,
+            // so when an MsalRuntimeBroker is created this will cause the interop layer to initialize
+            interop = new MsalRuntimeInterop();
+        } catch (MsalInteropException e) {
+            //TODO: Meaningful error message explaining why MSALRuntime failed to initialize (unsupported platform, error from MSALRuntime, etc.)
```
siddhijain: Lets complete this TODO as it is very important for the app developer to understand the next steps.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -3,58 +3,84 @@
...
      * Acquire a token silently, i.e. without direct user interaction
      *
      * This may be accomplished by returning tokens from a token cache, using cached refresh tokens to get new tokens,
      * or via any authentication flow where a user is not prompted to enter credentials
-     *
-     * @param requestParameters MsalRequest object which contains everything needed for the broker implementation to make a request
-     * @return IBroker implementations will return an AuthenticationResult object
      */
-    default IAuthenticationResult acquireToken(PublicClientApplication application, SilentParameters requestParameters) {
+    default void acquireToken(PublicClientApplication application, SilentParameters requestParameters, CompletableFuture<IAuthenticationResult> future) {
```
siddhijain: Should not the methods' return type be CompletableFuture<IAuthenticationResult>? Keeping the return type as void will make these apis different from the MSAL Java acquireToken apis.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -3,58 +3,84 @@
...
+                            Authority.createAuthority(new URL(authority)), JsonHelper.convertJsonToObject(idTokenJson,
+                                    IdToken.class), null));
+                }
+            }
+            if (accessToken != null) {
+                builder.accessToken(accessToken);
+                builder.expiresOn(accessTokenExpirationTime);
+            }
+        } catch (Exception e) {
+            //TODO: throw new msalexception. Could a valid broker result be an invalid MSAL Java result?
```
siddhijain: Ideally, the code should not reach here, but an exception could happen while converting json to object, parsing an id token, etc - in which case it would come under an invalid broker result. But it is good to throw a meaningful MSALException here, and we can figure out if we should pass this error back to MSALRuntime somehow to let them know about the invalid broker results.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -0,0 +1,192 @@
...
+
+        try (AuthParameters authParameters =
+                     new AuthParameters
+                             .AuthParametersBuilder(application.clientId(),
+                             application.authority(),
+                             parameters.scopes().toString())
+                             .build()) {
+
+            //Account was not cached in MSALRuntime, must perform sign in first to populate account info
+            if (accountResult == null) {
```
siddhijain, line 62: nit - It's a better practice to check for nulls this way - https://stackoverflow.com/questions/10983573/checking-for-null-what-order

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
Avery-Dunn: I originally had it returning a CompletableFuture, however that made the code in MSAL Java itself kind of messy. If you look at `MsalRuntimeBroker`, each of the long `.thenApply(...);` sections would need to move into PublicClientApplication (or somewhere else in MSAL Java).

Doing that would also mean splitting up where the interop's result is retrieved and where it is parsed: rather than the IBroker implementation handling that entirely, it would be up to MSAL Java to call the correct methods to translate whatever the broker returned into an IAuthenticationResult

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.concurrent.ExecutionException;
+
+public class MsalRuntimeBroker implements IBroker {
+    private static MsalRuntimeInterop interop;
+
+    static {
+        try {
+            //MsalRuntimeInterop performs various initialization steps in a similar static block,
+            // so when an MsalRuntimeBroker is created this will cause the interop layer to initialize
+            interop = new MsalRuntimeInterop();
```
siddhijain, line 36: Where is the logic to determine which flavor of msalruntime to use? Is it in the java interop layer?

Avery-Dunn: Yes, interop will handle 100% of any platform-specific stuff, and it choses the correct MSALRuntime dll here: https://github.com/AzureAD/microsoft-authentication-library-for-cpp/blob/9534b41ac5a168641c80f3ecfc97d9a6e765da19/msalruntime/interop/java/src/main/java/com/microsoft/azure/javamsalruntime/MsalRuntimeInterop.java#L252

The exact dll that gets loaded doesn't matter to the broker, it's all just the same Java object with the same set of methods either way (JNA handles actually loading and calling the methods in the dll)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
```diff
@@ -112,17 +122,48 @@ public CompletableFuture<IAuthenticationResult> acquireToken(InteractiveRequestP
...
-        CompletableFuture<IAuthenticationResult> future = executeRequest(interactiveRequest);
-        futureReference.set(future);
+        CompletableFuture<IAuthenticationResult> future = new CompletableFuture<>();
+
+        if (this.broker != null) {
+            broker.acquireToken(this, parameters, future);
+            futureReference.set(future);
+        } else {
+            future = executeRequest(interactiveRequest);
+            futureReference.set(future);
```
siddhijain: Since this is common between both conditions, it can be moved out of the if-else block.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -3,58 +3,80 @@
...
     }
 
     /**
      * Acquire a token silently, i.e. without direct user interaction, using username/password authentication
-     *
-     * @param requestParameters MsalRequest object which contains everything needed for the broker implementation to make a request
-     * @return IBroker implementations will return an AuthenticationResult object
      */
-    default IAuthenticationResult acquireToken(PublicClientApplication application, UserNamePasswordParameters requestParameters) {
+    default void acquireToken(PublicClientApplication application, UserNamePasswordParameters parameters, CompletableFuture<IAuthenticationResult> future) {
```
siddhijain: It does not look right to have 'void' as the return type where a token is expected. These methods should still return CompletableFuture<IAuthentication> result.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
```diff
@@ -115,9 +115,17 @@ public class AuthenticationErrorCode {
...
-     * be accessed by MSAL Java*/
+     * be accessed by MSAL Java
+     */
     public final static String MISSING_BROKER = "missing_broker";
+
+    /**
+     * Indicates an error from the Java/MSALRuntime interop layer used by the Java Brokers package,
+     * and will generally just be forwarding an error message from the interop layer or MSALRuntime itself
+     */
+    public final static String MSALRUNTIME_INTEROP_ERROR = "missing_broker";
```
siddhijain: shouldn't this constant have a different value than MISSING_BROKER?

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
siddhijain: Thanks. Looking at this class MsalRuntimeInterop.java, I see a lot of TODOs. Lets make sure they are taken care of before we move on to the next phase of implementing logging and tracing.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
Avery-Dunn: In the latest commit this throws a new exception explaining that the interop could not be initialized, and it includes the message from the MsalInteropException thrown here.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
Avery-Dunn: Fixed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java
Avery-Dunn: Yep, just a copy/paste error. Fixed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
Avery-Dunn: In the latest commit this has been adjusted: 
- Previously, MSAL Java initialized a CompletableFuture<IAuthenticationResult>, and sent it to a broker package as a parameter to get it configured with competition/cancelation steps. The broker methods had void return types, as the future object was already initialized in MSAL Java
- Now, MSAL Java just creates an uninitialized CompletableFuture<IAuthenticationResult>, and sets the value of that future to the return value of a broker method. The methods in the broker package create their own internal CompletableFuture<IAuthenticationResult>, and return that future after configuring the completion/cancelation steps

The underlying process for configuring the future objects is the same, except it should now be clearer that the methods in the broker package are actually producing the result returned to an app developer.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -0,0 +1,178 @@
...
+import com.microsoft.aad.msal4j.SilentParameters;
+import com.microsoft.aad.msal4j.UserNamePasswordParameters;
+import com.microsoft.aad.msal4j.MsalClientException;
+import com.microsoft.aad.msal4j.AuthenticationErrorCode;
+import com.microsoft.aad.msal4j.IAccount;
+import com.microsoft.azure.javamsalruntime.Account;
+import com.microsoft.azure.javamsalruntime.AuthParameters;
+import com.microsoft.azure.javamsalruntime.AuthResult;
+import com.microsoft.azure.javamsalruntime.MsalInteropException;
+import com.microsoft.azure.javamsalruntime.MsalRuntimeInterop;
```
rayluo, line 19: Is the MsalRuntimeInterop package always available? Will the [line 36](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/563/files#diff-fb3a656752687c9bd4ce7d08e3912e0fc520cf78a3838e79fff5cce47ce541dbR36) below always succeed?

Avery-Dunn: Yes, the msal4jbrokers package has javamsalruntime [as a dependency](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dc79a19d0f937e73e1d60c8c2cf23e3f9cdbe4ff/msal4j-brokers/pom.xml#L35), so it and all of its classes (like MsalRuntimeInterop) will always be available.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
Avery-Dunn: This has been replaced with `isBrokerAvailable()`. Rather than throw an exception, this will only determine whether we use the broker or if we fallback to regular MSAL Java.

For MSALRuntime, if we can call their startup API then we assume the broker is available on the platform and try to use it, otherwise we just log a message saying it's unavailable and will use MSAL Java instead.

msal4j-brokers/pom.xml
Avery-Dunn: In the latest commit I changed this to [0.13.3,) . This will start with version 0.13.3 of the interop (the current MSALRuntime version), and will get the latest version from then on (so that we don't need to update the broker package when we update the interop)

Issue #481, opened 2022-03-22
Consistency track: please provide consistent APIs to disable authority validation and instance discovery


xiangyan99:
Now we have different API/behaviors for customers to disable authority validation and instance discovery.

Please sync across languages to provide consistent ones.

This is important because JEDI scenarios are blocked by current behavior.

siddhijain:
This feature is in 'private preview now. Version 1.0.0-beta.1 contains the latest changes.

siddhijain:
The feature is now released in v1.13.4. Closing the issue.

PR #569, opened 2022-11-24
expose instanceDiscovery flag

siddhijain:
expose instanceDiscovery flag which is set to true by default. 

Reference to python PR https://github.com/AzureAD/microsoft-authentication-library-for-python/commit/dd7b53805efb985f6a85c22f294f20d9adfcc01d

### DIFF
--- a/msal4j-sdk/README.md
+++ b/msal4j-sdk/README.md
@@ -16,7 +16,7 @@ Quick links:
 The library supports the following Java environments:
 - Java 8 (or higher)
 
-Current version - 1.0.0-beta.1
+Current version - 1.13.4
 
 You can find the changes for each version in the [change log](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/master/changelog.txt).
 
@@ -28,13 +28,13 @@ Find [the latest package in the Maven repository](https://mvnrepository.com/arti
 <dependency>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>msal4j</artifactId>
-    <version>1.0.0-beta.1</version>
+    <version>1.13.4</version>
 </dependency>
 ```
 ### Gradle
 
 ```gradle
-compile group: 'com.microsoft.azure', name: 'msal4j', version: '1.0.0-beta.1'
+compile group: 'com.microsoft.azure', name: 'msal4j', version: '1.13.4'
 ```
 
 ## Usage
--- a/msal4j-sdk/README.md
+++ b/msal4j-sdk/README.md
@@ -16,7 +16,7 @@ Quick links:
 The library supports the following Java environments:
 - Java 8 (or higher)
 
-Current version - 1.13.3
+Current version - 1.0.0-beta.1
 
 You can find the changes for each version in the [change log](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/master/changelog.txt).
 
@@ -28,13 +28,13 @@ Find [the latest package in the Maven repository](https://mvnrepository.com/arti
 <dependency>
     <groupId>com.microsoft.azure</groupId>
     <artifactId>msal4j</artifactId>
-    <version>1.13.3</version>
+    <version>1.0.0-beta.1</version>
 </dependency>
 ```
 ### Gradle
 
 ```gradle
-compile group: 'com.microsoft.azure', name: 'msal4j', version: '1.13.3'
+compile group: 'com.microsoft.azure', name: 'msal4j', version: '1.0.0-beta.1'
 ```
 
 ## Usage
--- a/msal4j-sdk/pom.xml
+++ b/msal4j-sdk/pom.xml
@@ -3,7 +3,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>msal4j</artifactId>
-	<version>1.13.3</version>
+	<version>1.0.0-beta.1</version>
 	<packaging>jar</packaging>
 	<name>msal4j</name>
 	<description>
--- a/msal4j-sdk/pom.xml
+++ b/msal4j-sdk/pom.xml
@@ -3,7 +3,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>msal4j</artifactId>
-	<version>1.0.0-beta.1</version>
+	<version>1.13.4</version>
 	<packaging>jar</packaging>
 	<name>msal4j</name>
 	<description>
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
@@ -30,7 +30,7 @@ public class TestConstants {
     public final static String COMMON_AUTHORITY_WITH_PORT = MICROSOFT_AUTHORITY_HOST_WITH_PORT + "msidlab4.onmicrosoft.com";
     public final static String MICROSOFT_AUTHORITY = MICROSOFT_AUTHORITY_HOST + "microsoft.onmicrosoft.com";
     public final static String TENANT_SPECIFIC_AUTHORITY = MICROSOFT_AUTHORITY_HOST + MICROSOFT_AUTHORITY_TENANT;
-    public final static String REGIONAL_MICROSOFT_AUTHORITY_BASIC_HOST_WESTUS = "westus.r." + MICROSOFT_AUTHORITY_BASIC_HOST;
+    public final static String REGIONAL_MICROSOFT_AUTHORITY_BASIC_HOST_WESTUS = "westus.login.microsoft.com";
 
     public final static String ARLINGTON_ORGANIZATIONS_AUTHORITY = ARLINGTON_MICROSOFT_AUTHORITY_HOST + "organizations/";
     public final static String ARLINGTON_COMMON_AUTHORITY = ARLINGTON_MICROSOFT_AUTHORITY_HOST + "common/";
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
@@ -45,7 +45,7 @@ public class TestConstants {
     public final static String B2C_AUTHORITY_SIGN_IN = B2C_AUTHORITY + B2C_SIGN_IN_POLICY;
     public final static String B2C_AUTHORITY_ROPC = B2C_AUTHORITY + B2C_ROPC_POLICY;
     public final static String B2C_READ_SCOPE = "https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read";
-    public final static String B2C_MICROSOFTLOGIN_AUTHORITY = "https://login.microsoftonline.com/tfp/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_MICROSOFTLOGIN_AUTHORITY = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
     public final static String B2C_MICROSOFTLOGIN_ROPC = B2C_MICROSOFTLOGIN_AUTHORITY + B2C_ROPC_POLICY;
 
     public final static String LOCALHOST = "http://localhost:";
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -22,7 +22,7 @@ class AadInstanceDiscoveryProvider {
     private static final String AUTHORIZE_ENDPOINT_TEMPLATE = "https://{host}/{tenant}/oauth2/v2.0/authorize";
     private static final String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE = "https://{host}:{port}/common/discovery/instance";
     private static final String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
-    private static final String HOST_TEMPLATE_WITH_REGION = "{region}.{host}";
+    private static final String HOST_TEMPLATE_WITH_REGION = "{region}.r.{host}";
     private static final String SOVEREIGN_HOST_TEMPLATE_WITH_REGION = "{region}.{host}";
     private static final String REGION_NAME = "REGION_NAME";
     private static final int PORT_NOT_SET = -1;
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -67,7 +67,7 @@ class AadInstanceDiscoveryProvider {
 
             //If region autodetection is enabled and a specific region not already set,
             // set the application's region to the discovered region so that future requests can skip the IMDS endpoint call
-            if (msalRequest.application().azureRegion() == null && msalRequest.application().autoDetectRegion()
+            if (null == msalRequest.application().azureRegion() && msalRequest.application().autoDetectRegion()
                         && null != detectedRegion) {
                     msalRequest.application().azureRegion = detectedRegion;
             }
@@ -83,11 +83,11 @@ class AadInstanceDiscoveryProvider {
                 doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
             } else {
                 // instanceDiscovery flag is set to False. Do not perform instanceDiscovery.
-                cache.putIfAbsent(host, InstanceDiscoveryMetadataEntry.builder().
+                return InstanceDiscoveryMetadataEntry.builder().
                         preferredCache(host).
                         preferredNetwork(host).
                         aliases(Collections.singleton(host)).
-                        build());
+                        build();
             }
         }
 
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -22,7 +22,7 @@ class AadInstanceDiscoveryProvider {
     private static final String AUTHORIZE_ENDPOINT_TEMPLATE = "https://{host}/{tenant}/oauth2/v2.0/authorize";
     private static final String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE = "https://{host}:{port}/common/discovery/instance";
     private static final String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
-    private static final String HOST_TEMPLATE_WITH_REGION = "{region}.r.{host}";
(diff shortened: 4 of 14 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -643,6 +648,18 @@ public T azureRegion(String val) {
             return self();
         }
 
+        /** Historically, MSAL would connect to a central endpoint located at
```
Avery-Dunn, line 651: Rather than have a long comment here, it might be better to have a wiki page describing this behavior and make this comment briefer with an aka.ms link to that page.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -186,4 +188,137 @@ public void aadInstanceDiscoveryTest_AutoDetectRegion_NoRegionDetected() throws
...
+    private static Object[][] getAadClouds(){
+        return new Object[][] {{"https://login.microsoftonline.com/common"} , // #Known to Microsoft
+                {"https://private.cloud/foo"}//Private Cloud
+                };
+    }
+
+    @DataProvider(name = "b2cAdfsClouds")
+    private static Object[][] getNonAadClouds(){
+        return new Object[][] {{"https://contoso.com/adfs"}//ADFS
+//                {"https://login.b2clogin.com/contoso/b2c_policy"},//B2C
```
Avery-Dunn: Since B2C is commented out, are B2C and ADFS expected to behave the same way?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -80,7 +79,16 @@ static InstanceDiscoveryMetadataEntry getMetadataEntry(URL authorityUrl,
         InstanceDiscoveryMetadataEntry result = cache.get(host);
 
         if (result == null) {
-            doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
+            if(msalRequest.application().instanceDiscovery()){
+                doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
+            } else {
+                // instanceDiscovery flag is set to False. Do not perform instanceDiscovery.
```
Avery-Dunn, line 86: Very minor, but missing a 't': 'Do no**t** perform instanceDiscovery'

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
```diff
@@ -45,7 +45,7 @@ public class TestConstants {
     public final static String B2C_AUTHORITY_SIGN_IN = B2C_AUTHORITY + B2C_SIGN_IN_POLICY;
     public final static String B2C_AUTHORITY_ROPC = B2C_AUTHORITY + B2C_ROPC_POLICY;
     public final static String B2C_READ_SCOPE = "https://msidlabb2c.onmicrosoft.com/msidlabb2capi/read";
-    public final static String B2C_MICROSOFTLOGIN_AUTHORITY = "https://login.microsoftonline.com/tfp/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_MICROSOFTLOGIN_AUTHORITY = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
```
bgavrilMS: Would be good to have this in a separate PR to help reviewers. I think it's a separate commit, so possibly a rebase error

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -18,22 +18,22 @@
...
-    private final static String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
-    private final static String HOST_TEMPLATE_WITH_REGION = "{region}.r.{host}";
-    private final static String SOVEREIGN_HOST_TEMPLATE_WITH_REGION = "{region}.{host}";
-    private final static String REGION_NAME = "REGION_NAME";
-    private final static int PORT_NOT_SET = -1;
+    private static final String DEFAULT_TRUSTED_HOST = "login.microsoftonline.com";
+    private static final String AUTHORIZE_ENDPOINT_TEMPLATE = "https://{host}/{tenant}/oauth2/v2.0/authorize";
+    private static final String INSTANCE_DISCOVERY_ENDPOINT_TEMPLATE = "https://{host}:{port}/common/discovery/instance";
+    private static final String INSTANCE_DISCOVERY_REQUEST_PARAMETERS_TEMPLATE = "?api-version=1.1&authorization_endpoint={authorizeEndpoint}";
+    private static final String HOST_TEMPLATE_WITH_REGION = "{region}.r.{host}";
```
bgavrilMS: nit: for the other issue - this will be `{region}.login.microsoft.com`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -67,10 +67,9 @@ static InstanceDiscoveryMetadataEntry getMetadataEntry(URL authorityUrl,
 
             //If region autodetection is enabled and a specific region not already set,
             // set the application's region to the discovered region so that future requests can skip the IMDS endpoint call
-            if (msalRequest.application().azureRegion() == null && msalRequest.application().autoDetectRegion()) {
-                if (detectedRegion != null) {
+            if (msalRequest.application().azureRegion() == null && msalRequest.application().autoDetectRegion()
+                        && null != detectedRegion) {
```
bgavrilMS, line 72: nit: keep constant on the right side? https://stackoverflow.com/questions/2369226/object-null-or-null-object :)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -80,7 +79,16 @@ static InstanceDiscoveryMetadataEntry getMetadataEntry(URL authorityUrl,
         InstanceDiscoveryMetadataEntry result = cache.get(host);
 
         if (result == null) {
-            doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
+            if(msalRequest.application().instanceDiscovery()){
+                doInstanceDiscoveryAndCache(authorityUrl, validateAuthority, msalRequest, serviceBundle);
+            } else {
+                // instanceDiscovery flag is set to False. Do not perform instanceDiscovery.
+                cache.putIfAbsent(host, InstanceDiscoveryMetadataEntry.builder().
```
bgavrilMS: This cache is static / shared by all apps. An app which does not use this flag will still skip instance discovery because it will pull the data from this static cache. Probably a very obscure scenario, but still - you can just return the entry and not put it in the cache.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -186,4 +188,137 @@ public void aadInstanceDiscoveryTest_AutoDetectRegion_NoRegionDetected() throws
         Assert.assertTrue(entry.aliases().contains("login.microsoft.com"));
         Assert.assertTrue(entry.aliases().contains("sts.windows.net"));
     }
+
+    @DataProvider(name = "aadClouds")
+    private static Object[][] getAadClouds(){
+        return new Object[][] {{"https://login.microsoftonline.com/common"} , // #Known to Microsoft
+                {"https://private.cloud/foo"}//Private Cloud
```
bgavrilMS: Consider adding some of the sovereign clouds here

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -186,4 +188,137 @@ public void aadInstanceDiscoveryTest_AutoDetectRegion_NoRegionDetected() throws
...
+//                {"https://login.b2clogin.com/contoso/b2c_policy"},//B2C
+                };
+    }
+
+    /**
+     * when instance_discovery flag is set to true (by default), an instance_discovery is performed for authorityType = AAD and
+     *     hence, an exception is thrown while making a call to getMetaDataEntry() if instanceDiscoveryResponse is not mocked.
+    */
+    @Test(  dataProvider = "aadClouds",
+            expectedExceptions = StringIndexOutOfBoundsException.class)
```
bgavrilMS: This is a pretty general-purpose exception, which can be thrown from anywhere. It can potentially hide a bug and it makes the test difficult to read / understand. If `StringIndexOutOfBoundsException` comes from HTTP mock, consider replacing this with a strongly typed exception, e.g. HttpMockOutOfRange or smth.

That said, I think there are already tons of tests for this scenario. MSAL tests work by mocking the HTTP layer, so you add fake (request, response) pairs in a queue of mocked requests. If the mock is not consumed, test should fail.

So better test would be:

1. Add HTTP mock for /token but not for instance discovery and ensure instance discoevry cache is empty (i would imagine this already happens)
2. Acquire a token
3. Assert that HTTP mock queue is empty

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -186,4 +188,137 @@ public void aadInstanceDiscoveryTest_AutoDetectRegion_NoRegionDetected() throws
...
+                .scopes(Collections.singleton("scope")).build();
+
+        MsalRequest msalRequest = new AuthorizationCodeRequest(
+                parameters,
+                app,
+                new RequestContext(app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, parameters));
+
+        URL authorityURL = new URL(authority);
+
+       AadInstanceDiscoveryProvider.getMetadataEntry(
```
bgavrilMS: IMO, it's better to test the end to end rather than this object. So in this case, actually make an `AcquireTokenByAuthCode` and assert that HTTP instance discoevry call is made.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
```diff
@@ -186,4 +188,137 @@ public void aadInstanceDiscoveryTest_AutoDetectRegion_NoRegionDetected() throws
...
+                .scopes(Collections.singleton("scope")).build();
+
+        MsalRequest msalRequest = new AuthorizationCodeRequest(
+                parameters,
+                app,
+                new RequestContext(app, PublicApi.ACQUIRE_TOKEN_BY_AUTHORIZATION_CODE, parameters));
+
+        URL authorityURL = new URL(authority);
+
+        AadInstanceDiscoveryProvider.getMetadataEntry(
```
bgavrilMS: Same as above, better to test the E2E.

bgavrilMS (changes requested):
don't add the new entry to cache, because cache is static
tests can be more useful if they test the end to end experience and mock the http layer

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
siddhijain: sure, let's talk more about this.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
siddhijain: Yes that would have been better. The second PR took all the changes from the first one as well. Sorry about that.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
siddhijain: yes that's a good point. I noticed this recently and this needs to be changed at multiple places in the library. I will do this one to start with.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
siddhijain: That's a valid point. I will make the changes.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
Avery-Dunn: Would the instance discovery behavior for sovereign clouds be different in some way? Is it worth having both global and sovereign cloud tests here?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
siddhijain: Do you not see 't'. I see it :D

Avery-Dunn: Oops, I meant to put this comment in the other PR. The 't' is missing from the word 'not' in that version: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/570/files#r1035238440

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
bgavrilMS: Well, I don't know your codebase, but there may special rules related to public cloud vs sovereign cloud. For example, ESTS-R feature handles these 2 differently and it is also in this general area of the code. So when thinking about different clouds, it's good to have a test case for a sovereign cloud.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
rayluo: FWIW, I use one-or-two-paragraphs long comments all the time, because in-place comment gives better locality to next maintainer (or a future you). And it does not rely on an external link thus insusceptible to future bad-link problem (which can happen to aka.ms links, too). Of course, the long comment itself can end with a link to a long long specs or Microsoft Docs link, it is OK for those links to go stale, because the gist is already captured inside the comment.

Besides, will this long comment be automatically used to generate the API docs? If so, then it  is even preferable to write long description here.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
rayluo: In this Instance Discovery context, neither B2C nor ADFS provides Instance Discovery endpoint.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
siddhijain: Yes for the information presented here, I would prefer having these lines here instead of an aka.ms link. If the information is a page long, then it makes sense to move this to another link.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
siddhijain: That's right. Instance discovery should not be done for B2C or ADFS even if instance_discovery flag is true.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
siddhijain: Made changes in the latest commit.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
siddhijain: Added end-to-end tests in the latest commit.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryTest.java
siddhijain: Added E2E tests in the latest commit.

siddhijain:
> don't add the new entry to cache, because cache is static tests can be more useful if they test the end to end experience and mock the http layer
Done.

siddhijain:
https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/481

bgavrilMS (approved):

Avery-Dunn (approved):

siddhijain:
FYI - This branch has been published as a beta release for Azure SDK team to test.

msal4j-sdk/README.md
```diff
@@ -16,7 +16,7 @@ Quick links:
 The library supports the following Java environments:
 - Java 8 (or higher)
 
-Current version - 1.13.3
+Current version - 1.0.0-beta.1
```
rayluo: I do not know how the versioning work in Java. But if its dependency system follows the [precedence behavior of Semantic Versioning](https://semver.org/#spec-item-11):

> Example: ... < 1.0.0-beta < 1.0.0-beta.2 < 1.0.0-beta.11 < 1.0.0-rc.1 < 1.0.0 < ... < 1.13.3"

then a customer who already making a dependency of some recent version of MSAL Java (for example, ">1.12"), they will not pick up our "1.0.0-beta.1". The general practice is to ship the latest experimental behavior in "1.14.0-beta.1".

In other words, stable version and beta versions all share the same versioning naming space, despite their different suffix.

rayluo (approved):
Overall LGTM. Just added a minor comment. Still approving.

msal4j-sdk/README.md
siddhijain: Yes, since this was our first beta release, we did not know about the nomenclature. I learned about that after publishing the beta release. will keep this in mind next time. Thanks for reviewing!

g2vinay: Yeah, if you intend to release this as beta feature, then it will follow 1.14.0-beta.1, as last stable release is in 1.13.x series, as of this comment.

siddhijain: @g2vinay Do you need this change to happen for testing this feature?

g2vinay: @siddhijain 
the version 1.0.0-beta.1 indicates to user that the beta features will GA in 1.0.0 release, plus v1.0.0-beta.1 is already released, we cannot release it again.
For next beta release, I feel it is a good opportunity to correct the beta versioning and release it as v1.14.0-beta.1 and when the feature is ready for stable release, it can be released in v1.14.0

msal4j-sdk/pom.xml
```diff
@@ -3,7 +3,7 @@
 	<modelVersion>4.0.0</modelVersion>
 	<groupId>com.microsoft.azure</groupId>
 	<artifactId>msal4j</artifactId>
-	<version>1.13.3</version>
+	<version>1.0.0-beta.1</version>
```
g2vinay: This version is already released.
why is it marked to an already released version ?

siddhijain: Talked offline about release versioning. Resolving it here.

PR #583, opened 2023-01-17
update versions for release

siddhijain:


msal4j-sdk/pom.xml
```diff
@@ -240,6 +240,8 @@
                 <configuration>
                     <source>8</source>
                     <target>8</target>
+                    <testSource>11</testSource>
```
Avery-Dunn: Seems like the builds are failing due this trying to use Java 11 to build stuff for the tests.

Is that testng update that needed Java 11 part of this release?

siddhijain: Thanks for checking. This should not be a part of this PR. I will get it removed.

Avery-Dunn (approved):

PR #594, opened 2023-02-09
Add support for both current and legacy B2C authority formats

Avery-Dunn:
Originally, MSALs needed B2C URLs to contain 'tfp' in the path to identify them as B2C authorities, so they needed to look like: `https://<some host>/tfp/<tenant>/<policy>`

However, some recent changes to B2C should ensure that all URLs have 'b2cloginc.com' as part of the host, so current B2C authorities should look like: `https://<something>.b2clogin.com/<tenant>/<policy>`

This PR adds support for the newer format, while retaining support for the legacy format until we're either confident that all customers have been moved to the new format (due to future potential Azure requirements) or we're able to make a breaking change (i.e., we're releasing MSAL Java 2.0)

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
@@ -38,8 +38,8 @@ public class TestConstants {
     public final static String ARLINGTON_GRAPH_DEFAULT_SCOPE = "https://graph.microsoft.us/.default";
 
 
-    public final static String B2C_AUTHORITY = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
-    public final static String B2C_AUTHORITY_URL = "https://msidlabb2c.b2clogin.com/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_AUTHORITY = "https://msidlabb2c.b2clogin.com/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_AUTHORITY_LEGACY_FORMAT = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
     public final static String B2C_ROPC_POLICY = "B2C_1_ROPC_Auth";
     public final static String B2C_SIGN_IN_POLICY = "B2C_1_SignInPolicy";
     public final static String B2C_AUTHORITY_SIGN_IN = B2C_AUTHORITY + B2C_SIGN_IN_POLICY;
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
@@ -373,6 +373,18 @@ public abstract class AbstractClientApplicationBase implements IClientApplicatio
             return self();
         }
 
+        /**
+         * Set URL of the authenticating B2C authority from which MSAL will acquire tokens
+         *
+         * Valid B2C authorities should look like: https://<something>.b2clogin.com/<tenant>/<policy>
+         *
+         * MSAL Java also supports a legacy B2C authority format, which looks like: https://<host>/tfp/<tenant>/<policy>
+         *
+         * However, MSAL Java will eventually stop supporting the legacy format. See here for information on how to migrate to the new format: https://aka.ms/msal4j-b2c
+         *
+         * @param val a boolean value for validateAuthority
+         * @return instance of the Builder on which method was called
+         */
         public T b2cAuthority(String val) throws MalformedURLException {
             authority = Authority.enforceTrailingSlash(val);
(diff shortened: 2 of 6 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -373,6 +373,18 @@ public T authority(String val) throws MalformedURLException {
...
         }
 
+        /**
+         * Set URL of the authenticating B2C authority from which MSAL will acquire tokens
+         *
+         * Valid B2C authorities should look like: https://<something>.b2clogin.com/<tenant>/<policy>
+         *
+         * MSAL Java also supports a legacy B2C authority format, which looks like: https://<host>/tfp/<tenant>/<policy>
+         *
+         * However, MSAL Java will eventually stop supporting the legacy format. See here for information on how to migrate to the new format: https://aka.ms/msal4j-b2c
```
siddhijain, line 383: Nice documentation!

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
```diff
@@ -38,8 +38,8 @@ public class TestConstants {
     public final static String ARLINGTON_GRAPH_DEFAULT_SCOPE = "https://graph.microsoft.us/.default";
 
 
-    public final static String B2C_AUTHORITY = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
-    public final static String B2C_AUTHORITY_URL = "https://msidlabb2c.b2clogin.com/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_AUTHORITY = "https://msidlabb2c.b2clogin.com/msidlabb2c.onmicrosoft.com/";
+    public final static String B2C_AUTHORITY_LEGACY_FORMAT = "https://msidlabb2c.b2clogin.com/tfp/msidlabb2c.onmicrosoft.com/";
```
siddhijain, line 42: Did legacy urls have ".b2clogin.com" in them? If not, we should modify this value so the execution goes to the place where tfp is checked.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/B2CAuthority.java
```diff
@@ -26,27 +26,42 @@ class B2CAuthority extends Authority {
...
+        //
+        // However, legacy URLs using the old format must still be supported by these sorts of checks here and elsewhere, so for the near
+        //   future at least we must consider both formats as valid until we're either sure all customers are swapped,
+        //   or until we're comfortable with a potentially breaking change
         validatePathSegments(segments);
 
-        policy = segments[2];
-
-        final String b2cAuthorityFormat = "https://%s/%s/%s/%s/";
-        this.authority = String.format(
```
siddhijain, line 44: It is better to keep the final constant out of the String.format method call. All the constants should ideally be declared separately.

siddhijain (changes requested):
1. Change the legacy test constant so it actually tests the tfp condition. 
2. Move the constant out of the try catch block.
Rest everything looks good.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
Avery-Dunn: It seems like older B2C URLs could have it, since this constant has been unchanged since the original commit for it back in 2019: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/46c0121e8dbf2609d346565c343385f23c1c3972/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java#L34

I think they just weren't guaranteed to have it, which is why the original B2C_MICROSOFTLOGIN_AUTHORITY used 'login.microsoftonline.com': https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/46c0121e8dbf2609d346565c343385f23c1c3972/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java#L41

Unfortunately it seems like ID labs updated some URLs so the old B2C_MICROSOFTLOGIN_AUTHORITY  doesn't work anymore (you fixed that when you were updating the regional formats in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/574), so I don't think there's a way for us to test a non-'b2clogin.com' URL.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/B2CAuthority.java
Avery-Dunn: Originally I just got rid of the final entirely, because it didn't make sense to have a final in a method and have it only be used once in that method.

In the latest commit there are now constants for both the legacy and current formats (created outside the method). It's definitely more readable/maintainable to have constants for the two formats, even if they're only used once each.

siddhijain (approved):
Looks good to merge once the tests pass.

PR #596, opened 2023-02-13
regional endpoint change to always use login.microsoft.com

siddhijain:
Template updates to always call login.microsoft.com


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -47,7 +47,9 @@ class AadInstanceDiscoveryProvider {
 
         TRUSTED_HOSTS_SET.addAll(Arrays.asList(
                 "login.windows.net",
-                "login.microsoftonline.com"));
+                "login.microsoftonline.com",
+                "login.microsoft.com",
```
siddhijain, line 51: @Avery-Dunn  I added login.microsoft.com and sts.windows.net here. Does this look right? Why did we not have these 2 in the trusted hosts set?

Avery-Dunn: Just had a look through the file's history to confirm, and it seems like that list has been the same since 2019 (I moved things around during the first region PR, but didn't add anything to it)

My best guess is that those URLs didn't exist at the time, and simply never got added.

Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
bgavrilMS: Looks ok

https://login.microsoftonline.com/common/discovery/instance?api-version=1.1&authorization_endpoint=https%3A%2F%2Flogin.microsoftonline.com%2Fcommon%2Foauth2%2Fv2.0%2Fauthorize

bgavrilMS (approved):

PR #603, opened 2023-02-27
Add tests for a CIAM user and reduce test code duplication

Avery-Dunn:
This PR adds tests for a CIAM user to confirm that MSAL Java can handle them as-is, and these tests follow the list laid on in this MSAL Python PR: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/520

When writing these new tests I also simplified some of the existing ones, by reducing code duplication and made some common methods more generic/flexible. Because of this a lot of files are changed in this PR, however the important ones for the CIAM changes are:
- New method to get the CIAM test user: `LabUserProvider.getCiamUser()`
- New tests using that CIAM user:
   - `AcquireTokenInteractive.acquireTokenInteractive_Ciam()`
   - `ClientCredentialsIT.acquireTokenClientCredentials_ClientSecret_Ciam()`
   - `DeviceCodeIT.DeviceCodeFlowCIAMTest()`
   - `UsernamePasswordIT.acquireTokenWithUsernamePassword_Ciam()`

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -346,7 +346,7 @@ public class AcquireTokenSilentIT {
 
     private IConfidentialClientApplication getConfidentialClientApplications() throws Exception {
         String clientId = cfg.appProvider.getOboAppId();
-        String password = cfg.appProvider.getSecret();
+        String password = cfg.appProvider.getOboAppPassword();
 
         IClientCredential credential = ClientCredentialFactory.createFromSecret(password);
 
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -346,7 +346,7 @@ public class AcquireTokenSilentIT {
 
     private IConfidentialClientApplication getConfidentialClientApplications() throws Exception {
         String clientId = cfg.appProvider.getOboAppId();
-        String password = cfg.appProvider.getOboAppPassword();
+        String password = cfg.appProvider.getSecret();
 
         IClientCredential credential = ClientCredentialFactory.createFromSecret(password);
 
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
@@ -5,6 +5,7 @@ package com.microsoft.aad.msal4j;
 
 import labapi.AppCredentialProvider;
 import labapi.AzureEnvironment;
+import labapi.LabUserProvider;
 import org.testng.Assert;
 import org.testng.annotations.BeforeClass;
 import org.testng.annotations.DataProvider;
@@ -24,10 +25,12 @@ import static com.microsoft.aad.msal4j.TestConstants.KEYVAULT_DEFAULT_SCOPE;
 @Test
 public class ClientCredentialsIT {
     private IClientCertificate certificate;
+    private LabUserProvider labUserProvider;
 
     @BeforeClass
     void init() throws CertificateException, UnrecoverableKeyException, NoSuchAlgorithmException, KeyStoreException, NoSuchProviderException, IOException {
         certificate = CertificateHelper.getClientCertificate();
+        labUserProvider = LabUserProvider.getInstance();
     }
 
     @Test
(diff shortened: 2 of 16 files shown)


msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
```diff
@@ -54,7 +54,17 @@ public void acquireTokenClientCredentials_ClientAssertion() throws Exception {
 
         IClientCredential credential = ClientCredentialFactory.createFromClientAssertion(clientAssertion.assertion());
 
-        assertAcquireTokenCommon(clientId, credential);
+        assertAcquireTokenCommon(clientId, credential, TestConstants.MICROSOFT_AUTHORITY);
+    }
+
+    @Test
+    public void acquireTokenClientCredentials_ClientSecret_Ciam() throws Exception {
+        String clientId = "b8e9d222-c4ee-414c-ac29-b0eff1f32400";
```
rayluo: I know this client id showed up in the second URL that I listed in the [MSAL Python PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/520#issue-1487352636), but our goal is to not hardcode such a client id. It should come from the first API.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
```diff
@@ -34,6 +34,8 @@ public class TestConstants {
 
     public final static String REGIONAL_MICROSOFT_AUTHORITY_BASIC_HOST_EASTUS = "eastus.login.microsoft.com";
 
+    public final static String CIAM_AUTHORITY = MICROSOFT_AUTHORITY_HOST + "msidlabciam1.onmicrosoft.com";
```
rayluo, line 37: Can this be loaded from the 2 lab api, rather than hardcoded?

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/Config.java
```diff
@@ -37,6 +37,10 @@ public class Config {
                 appProvider = new AppCredentialProvider(azureEnvironment);
                 tenant = TestConstants.ARLINGTON_AUTHORITY_TENANT;
                 break;
+            case "ciam" :
+                tenantSpecificAuthority = TestConstants.CIAM_AUTHORITY;
+                graphDefaultScope = TestConstants.GRAPH_DEFAULT_SCOPE;
```
rayluo: Can this be loaded from the 2 lab api, rather than hardcoded? Besides, CIAM does not use the graph default scope.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
```diff
@@ -346,7 +346,7 @@ public void acquireTokenSilent_emptyScopeSet(String environment) throws Exceptio
 
     private IConfidentialClientApplication getConfidentialClientApplications() throws Exception {
         String clientId = cfg.appProvider.getOboAppId();
-        String password = cfg.appProvider.getOboAppPassword();
+        String password = cfg.appProvider.getSecret();
```
siddhijain: Why this change?

siddhijain (commented):
Looks good overall. Need to take care of failing build and integration tests.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
Avery-Dunn: I changed that method name just because the 'obo' part of the name seemed misleading: the values were coming from a method called [getSecret()](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/9683c9d0f9b5a4d28438ac63be7d8f44e2b7be51/msal4j-sdk/src/integrationtest/java/labapi/KeyVaultSecretsProvider.java#L38) which isn't restricted to just OBO-related secrets, and it was also being used in some [acquire token silent tests](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/9683c9d0f9b5a4d28438ac63be7d8f44e2b7be51/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java#L349) which aren't using OBO either.

However, the AppCredentialProvider class that has that method also has some other fields/methods with 'obo' in the name. I didn't change those since they weren't used in the new CIAM tests, but changing just one feels like it'd be more confusing than leaving them all. In the latest commit I reverted this change, so it's back to just being 'getOboAppPassword()'

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TestConstants.java
Avery-Dunn: I could get it from the lab API, but we have a lot of other authority constants here. I'll probably rearrange things to retrieve most of these constants from the labs API in some future PR, but for now I'll keep it consistent and hardcode it

rayluo: Fair enough.

Issue #256, opened 2020-06-30
[Enhancement] Add support for extraQueryParameters


navyasric:
Spec: 

- [x] new API at request level (on **all** AcquireToken* methods) should allow passing in a map of key / value string pairs
- [x] in public client, the authorization url which is used to open the browser, must use append these parameters
- [x] in confidential client, the `GetAuthorizationUrl` API should take into account these parameters
- [x] in both clients, the token endpoint must append these parameters (even though the request is a POST request)
- [x] when talking to broker, the parameters must be added the `Properties` bag of auth parameters
- [x] in case of conflict, (e.g. app developer tries to pass in `scope=foo` and call AcquireTokenInteractive), throw exception

Needed for a number of functionalities to be enabled. E.g.: 
Customers needs support of extra query parameters in order to use ROPC flow with B2C

henrik-me:
Similar to: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/255

bgavrilMS:
This will be needed for full WAM support.

bgavrilMS:
@siddhijain - is this done? Can this be closed?

siddhijain:
This will be closed once we update the broker PR with
'when talking to a broker, the parameters must be added to the Properties bag of auth parameters'. Avery is planning to update. Tagging him to close this once done. @Avery-Dunn

Avery-Dunn:
Support for extraQueryParameters in the broker package was add as part of https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/639, so this is fully supported.

PR #607, opened 2023-03-13
expose extraQueryParameters

siddhijain:


Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -67,6 +67,15 @@ OAuthHttpRequest createOauthHttpRequest() throws SerializeException, MalformedUR
             params.put("claims", Collections.singletonList(claimsRequest));
         }
 
+        if(msalRequest.requestContext().apiParameters().extraQueryParameters() != null ){
+            for(String key: msalRequest.requestContext().apiParameters().extraQueryParameters().keySet()){
+                    if(params.containsKey(key)){
```
bgavrilMS, line 72: There is no need to add this validation. If app developer sets this and overrides some params that MSAL computes, it's their problem. 

Consider adding just a `log.warning`

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
```diff
@@ -178,9 +175,12 @@ public void acquireTokenSilent_ConfidentialClient_acquireTokenSilent(String envi
         cfg = new Config(environment);
 
         IConfidentialClientApplication cca = getConfidentialClientApplications();
-
+        //test that adding extra query parameters does not break the flow
+        Map<String, String> extraParameters = new HashMap<>();
+        extraParameters.put("test","test");
         IAuthenticationResult result = cca.acquireToken(ClientCredentialParameters
                 .builder(Collections.singleton(cfg.graphDefaultScope()))
+                        .extraQueryParameters(extraParameters)
```
bgavrilMS, line 183: Is there a way to assert this? Maybe by intercepting the HTTP request message and asserting that the URL indeed contains `test=test` ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
siddhijain: Sure I was going by the requirements, thinking this is what all the MSALs do. https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/256 Point 5.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
siddhijain: I wanted to do that too - One way of intercepting these calls is by mocking HTTP requests (which we do not prefer). These tests do not mock the HTTP requests and hence, the request cannot be intercepted. I will check to see if I can find another way.

bgavrilMS: I believe all MSAL libraries allow injection of custom HTTP interface - https://github.com/AzureAD/microsoft-authentication-library-for-java/wiki/Configure-Http-Client and this seems like a good way to write unit tests ?

siddhijain: That's a good point. By providing a simple CustomHttpClient, I can **build** an app that asserts that the httpRequest in the httpClient contains the extraQueryParameters.  

However, if we need to test end-to-end (and actually acquire tokens using httpClient), providing a custom HTTP client needs quite some code to be written to make get and post requests and process responses. https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java

Is it worth writing this much code to test this? Thoughts?

bgavrilMS: I was thinking just mocking IHttpClient like this - https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/00c991a91f154cb7bad379da21504444c2405ee6/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ConfidentialClientApplicationUnitT.java#L246

Not suggesting you to have an E2E test here, unit tests only. 

Most (>95%) of MSAL.NET's unit tests are written like this: 

- use a mock IHttpClient (see [MockHttpManager ](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/03fd41da88d9c23d15dd800ff57f63d2a2ffecea/tests/Microsoft.Identity.Test.Common/Core/Mocks/MockHttpManager.cs#L21))
- the mock object has a queue of (request, response) objects which the tests sets
- on each `send`, you dequeue a (request, response) and serve it from there. 
- also have an option to assert the request, so for example you can see the actual URI 

There may be a lot of code at first, but it's good infrastructure that you can reuse later.

siddhijain: Got it. Thanks for clarifying. I was reluctant to go down the route of mocking httpClient using EasyMock, as later in those tests, we use PowerMock to assert results. And we are planning to stop using PowerMock in the test cases. Let me explore using the JUnit framework to write this test case and achieve the same result.

siddhijain: Updated an existing test case to capture the HTTP request and validate the parameters. https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/607/files#:~:text=Assert.assertTrue(body.contains(%22test%3Dtest%22))

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
siddhijain: Updated in the latest commit.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
bgavrilMS: LGTM. Sometimes these mocking frameworks are too complex for their own good. At the end of the day, it's just about adding a test implementation of IHttpClient

bgavrilMS (approved):

PR #608, opened 2023-03-16
APIs for toggling MSALRuntime's logging

Avery-Dunn:
MSALRuntime allows logging and PII data to be toggled on and off, however in MSAL Java these are not toggleable: logging is always enabled through SLF4J, and PII is set once [when the client app is made](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java#L55).

Making these toggleable in MSAL Java would require either a breaking change, or a parallel set of logging APIs that would be confusing or useless to anyone not using the broker. 

Because of that, this PR adds new APIs to the broker package, allowing app developers to configure MSALRuntime's logging separately from MSAL Java's. In enabled, app developers will see MSALRuntime's verbose logs through the SLF4J framework in the same way they see MSAL Java's logs.

### DIFF
--- a/msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
+++ b/msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
@@ -187,7 +187,7 @@ public class MsalRuntimeBroker implements IBroker {
      *
      * @param enableLogging true to enable MSALRuntime logs, false to disable it
      */
-    public void toggleLogging(boolean enableLogging) {
+    public void enableBrokerLogging(boolean enableLogging) {
         try {
             MsalRuntimeInterop.toggleMsalRuntimeLogging(enableLogging);
         } catch (Exception ex) {
@@ -202,7 +202,7 @@ public class MsalRuntimeBroker implements IBroker {
      *
      * @param enablePII true to allow PII to appear in logs and error messages, false to disallow it
      */
-    public void togglePII(boolean enablePII) {
+    public void enableBrokerPIILogging(boolean enablePII) {
         try {
             MsalRuntimeInterop.toggleMsalRuntimePIILogging(enablePII);
         } catch (Exception ex) {
--- a/msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
+++ b/msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
@@ -176,4 +176,37 @@ public class MsalRuntimeBroker implements IBroker {
             return false;
         }
     }
+
+    /**
+     * Toggles whether or not detailed MSALRuntime logs will appear in MSAL Java's normal logging framework.
+     *
+     * If enabled, you will see logs directly from MSALRuntime, containing verbose information relating to telemetry, API calls,successful/failed requests, and more.
+     * These logs will appear alongside MSAL Java's logs (with a message indicating they came from MSALRuntime), and will follow the same log level as MSAL Java's logs (info/debug/error/etc.).
+     *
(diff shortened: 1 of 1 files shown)


msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -176,4 +176,37 @@ public boolean isBrokerAvailable() {
...
+     * Toggles whether or not detailed MSALRuntime logs will appear in MSAL Java's normal logging framework.
+     *
+     * If enabled, you will see logs directly from MSALRuntime, containing verbose information relating to telemetry, API calls,successful/failed requests, and more.
+     * These logs will appear alongside MSAL Java's logs (with a message indicating they came from MSALRuntime), and will follow the same log level as MSAL Java's logs (info/debug/error/etc.).
+     *
+     * If disabled (default), MSAL Java will still produce some logs related to MSALRuntime, particularly in error messages, but will be much less verbose.
+     *
+     * @param enableLogging true to enable MSALRuntime logs, false to disable it
+     */
+    public void toggleLogging(boolean enableLogging) {
```
siddhijain: nit - rename the method to enableLog or enableBrokerLog. toggleLogging seems ambiguous. toggleLogging - false -> does this mean do not toggle? toggleLogging - true -> does this mean toggle the current value?

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -176,4 +176,37 @@ public boolean isBrokerAvailable() {
...
+    }
+
+    /**
+     * If enabled, Personal Identifiable Information (PII) can appear in logs and error messages produced by MSALRuntime.
+     *
+     * If disabled (default), PII will not be shown, and you will simply see "(PII)" or similar notes in places where PII data would have appeared.
+     *
+     * @param enablePII true to allow PII to appear in logs and error messages, false to disallow it
+     */
+    public void togglePII(boolean enablePII) {
```
siddhijain: same as above. enableLogPII?

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -176,4 +176,37 @@ public boolean isBrokerAvailable() {
             return false;
         }
     }
+
```
bgavrilMS, line 179: MSAL Java already has logging APIs that control both verbosity and PII. Why aren't those used ? We should not have special APIs for WAM logging.

bgavrilMS (changes requested):
Use existing logging mechanism instead of exposing new one.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
Avery-Dunn: MSAL Java doesn't really have an API to control the verbosity: we send logs through a logging framework called SLF4J and we set the verbosity for those logs (info/debug/error/etc.). However, it's up to an app developer to get a dependency to receive the logs we send through SLF4J, and it's up to them to us that separate dependency to configure what level they want to see. 

Because we'd need a new method to enable logging anyway, and since MSALRuntime logs are very verbose at every level (compare to MSAL Java at least), I figured it'd be better to give app developers an API specifically to turn broker logs on/off rather than create a new API in MSAL Java itself (which wouldn't be relevant unless you're using the broker package).

As for PII, we allow app developers to turn it on at the app level, but there's no way to turn it off without completely recreating the PublicClientApplication object. This one could be easily hooked into MSAL Java's existing flag, but again since the MSALRuntime logs are created separately from MSAL Java and their available APIs behave differently, it made more sense to match the behavior rather than hide it.

PR #628, opened 2023-04-14
add refresh_in logic

siddhijain:


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java
```diff
@@ -35,6 +35,13 @@ private static void validateTokenProviderResult(TokenProviderResult tokenProvide
         if (null == tokenProviderResult.getTenantId() || tokenProviderResult.getTenantId().isEmpty()) {
             handleInvalidExternalValueError(tokenProviderResult.getTenantId());
         }
+
+        if (0 == tokenProviderResult.getRefreshInSeconds()){
+            long expireInSeconds = tokenProviderResult.getExpiresInSeconds();
+            if(expireInSeconds >= TWO_HOURS){
+                tokenProviderResult.setRefreshInSeconds(expireInSeconds/2);
```
bgavrilMS, line 44: nit: the name of the method suggest a read-only operation (validate) but this method actually changes state. Rename or move to a separate private method.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java
```diff
@@ -35,6 +35,13 @@ private static void validateTokenProviderResult(TokenProviderResult tokenProvide
         if (null == tokenProviderResult.getTenantId() || tokenProviderResult.getTenantId().isEmpty()) {
             handleInvalidExternalValueError(tokenProviderResult.getTenantId());
         }
+
+        if (0 == tokenProviderResult.getRefreshInSeconds()){
```
bgavrilMS, line 41: Are there any unit tests for AppTokenProvider? You could check that this refresh_in is set by looking in the token cache - the access token should have a "refresh_in" property in the payload.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java
siddhijain: Makes sense. Update the method name.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java
siddhijain: The access token will have a refresh_in property even without this change. The value of refresh_in would be 0 in that case. I can update a test by not providing a refresh_in value and asserting that refresh_in is not 0 / null / blank in the access token.

siddhijain: Updated test.

Avery-Dunn (approved):

PR #639, opened 2023-05-01
Add support for POP tokens to MSAL Java and MSAL Java Brokers

Avery-Dunn:
Adds POP token support by introducing a new AuthScheme parameter than can be set in supported flows, and calling the relevant interop API to set up the required auth parameters to request POP tokens.

The interop changes in this PR https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3514 need to be released before any of these changes will work outside of a dev environment.


msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -125,6 +131,13 @@ public CompletableFuture<IAuthenticationResult> acquireToken(PublicClientApplica
 
             authParameters.setUsernamePassword(parameters.username(), new String(parameters.password()));
 
+            //If POP auth scheme configured, set parameters to get MSALRuntime to return POP tokens
```
siddhijain, line 137: These lines of code can be moved to a method to prevent duplication.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthScheme.java
```diff
@@ -0,0 +1,65 @@
+package com.microsoft.aad.msal4j;
+
+import java.net.URI;
+
+//TODO: javadocs explaining this class, link to documentation page
+public class AuthScheme {
```
bgavrilMS: Is it necessary to expose a the "Auth Scheme" concept? The only API we need is 

`WithProofOfPosession(Uri uri, HttpMethod m, string nonce)`

bgavrilMS: Internally you can use this concept, but MSAL APIs do not expose it. 

The spec is here - https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PoPTokensProtocol/PoP_API_In_MSAL.md&version=GBdev&anchor=key-case-sensitivity

You should also implement a helper named `IsProofOfPosessionSupportedByClient`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthScheme.java
```diff
@@ -0,0 +1,65 @@
...
+
+    String httpMethod;
+    URI uri;
+    String nonce;
+    SchemeType type;
+
+    public enum SchemeType {
+        POP
```
bgavrilMS: Code smell - 1 element enum. This class isn't actually extensible to other protcols. Bearer scheme does not require HttpMethod / Uri etc.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/UserNamePasswordParameters.java
```diff
@@ -58,6 +58,13 @@ public class UserNamePasswordParameters implements IAcquireTokenParameters {
      */
     private String tenant;
 
+    /**
+     * Sets the auth scheme for this request.
+     *
+     * For more information about auth schemes, see {@link AuthScheme} and (TODO: link to documentation page)
+     */
+    private AuthScheme authScheme;
+
```
bgavrilMS, line 68: There are no tests in this PR. 
The API Review spec has acceptance tests - you should automate as many as possible - https://identitydivision.visualstudio.com/DevEx/_git/AuthLibrariesApiReview?path=/PoPTokensProtocol/PoP_API_In_MSAL.md&version=GBdev&_a=preview


If PR is not ready for review, consider marking it as draft.

bgavrilMS (changes requested):
Not as per spec. 
No tests.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
```diff
@@ -3,15 +3,7 @@
 
 package com.microsoft.aad.msal4jbrokers;
```
bgavrilMS, line 5: Note that msal_runtime authResult will return an unsigned token. This is the correct way to get the token. @MSamWils - I think this is an msal_runtime bug.

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9e4bf560970fb5be9a72b83b3a446ce49c545576/src/client/Microsoft.Identity.Client/Platforms/Features/RuntimeBroker/WamAdapters.cs#L325

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthScheme.java
Avery-Dunn: Removed in the latest commit

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthScheme.java
Avery-Dunn: Originally I just followed the design in [Python's PR](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/511), since it also introduced the concepts of PoP tokens, auth schemes, etc., and that seemed to just have `AuthScheme` as the parameter instead of `WithProofOfPosession` so I figured that was just the minimal design we should aim for too.

I agree it should be more consistent with the design doc though, so in the latest commit the relevant flows have a parameter named `proofOfPossession` (our parameters don't start with 'with'), this `AuthScheme` class was renamed to PopParameters (since we'd only have the one scheme anyway) and it's package-private.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/InteractiveRequestParameters.java
```diff
@@ -130,4 +125,23 @@ public static InteractiveRequestParametersBuilder builder(URI redirectUri) {
...
+    public static class InteractiveRequestParametersBuilder {
+
+        /**
+         * Sets the PopParameters for this request, allowing the request to retrieve proof-of-possession tokens rather than bearer tokens
+         *
+         * For more information, see {@link PopParameters} and https://aka.ms/msal4j-pop
+         *
+         * @param httpMethod a valid HTTP method, such as "GET" or "POST"
+         * @param uri URI to associate with the token
+         * @param nonce optional nonce value for the token, can be empty or null
```
bgavrilMS:
```suggestion
         * @param nonce a string obtained by calling the resource (e.g. Microsoft Graph) un-authenticated and parsing the WWW-Authenticate header associated with pop authentication scheme and extracting the nonce parameter, or, on subsequent calls, by parsing the Autheticate-Info header and extracting the nextnonce parameter.
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/InteractiveRequestParameters.java
```diff
@@ -130,4 +125,23 @@ public static InteractiveRequestParametersBuilder builder(URI redirectUri) {
...
+    //This Builder class is used to override Lombok's default setter behavior for any fields defined in it
+    public static class InteractiveRequestParametersBuilder {
+
+        /**
+         * Sets the PopParameters for this request, allowing the request to retrieve proof-of-possession tokens rather than bearer tokens
+         *
+         * For more information, see {@link PopParameters} and https://aka.ms/msal4j-pop
+         *
+         * @param httpMethod a valid HTTP method, such as "GET" or "POST"
+         * @param uri URI to associate with the token
```
bgavrilMS:
```suggestion
         * @param uri the URI on the downstream protected API which the application is trying to access, e.g. https://graph.microsoft.com/beta/me/profile
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/InteractiveRequestParameters.java
```diff
@@ -130,4 +125,23 @@ public static InteractiveRequestParametersBuilder builder(URI redirectUri) {
...
+
+    //This Builder class is used to override Lombok's default setter behavior for any fields defined in it
+    public static class InteractiveRequestParametersBuilder {
+
+        /**
+         * Sets the PopParameters for this request, allowing the request to retrieve proof-of-possession tokens rather than bearer tokens
+         *
+         * For more information, see {@link PopParameters} and https://aka.ms/msal4j-pop
+         *
+         * @param httpMethod a valid HTTP method, such as "GET" or "POST"
```
bgavrilMS, line 142: Is there an enum you can use instead? Maybe https://docs.oracle.com/javame/8.0/api/httpclient/api/com/oracle/httpclient/HttpMethod.html ?

I'm thinking that if you accept a string here, you'll have to validate it.

Avery-Dunn: I have no idea if this is a realistic scenario, but is there any reason to restrict this to just the common HTTP headers (like get and post) and not allow app developers to pass in custom HTTP headers?

I just tested MSALRuntime's `SetPopParams` API with an HTTP header of "ABCD" and it gave me back a brand new access token without issue. No idea if this token could actually be used to access something via an "ABCD" header, but it seems like MSALRuntime considered it to be valid input.

bgavrilMS: Ack.

Note that this isn't an HTTP header, it's an HTTP method. The variability for methods is much smaller, but the spec does indeed say that new methods can be defined https://www.rfc-editor.org/rfc/rfc7231#section-4.3

siddhijain: It is a great find. Even when a new method can be defined, I think we should just restrict it to the common values as this is what we support for now. Having httpMethod of type HttpMethod rather than String is very common in Java. I will incline towards making it accept HttpMethod too.

Avery-Dunn: When I said "custom HTTP headers" I meant "custom HTTP methods", but yeah I assume the use cases of that are pretty small. 

In the latest commit the parameter now takes in a value from our existing HTTPMethod enum rather than an arbitrary string.

msal4j-brokers/src/main/java/com/microsoft/aad/msal4jbrokers/MsalRuntimeBroker.java
Avery-Dunn: After trying a couple options it seems cleaner to handle this in the interop package rather than this package. 

Basically, if this is some MSALRuntime-specific behavior then the interop should deal with it, and the msal4j-brokers package should just expect a valid result based on the parameters it sends to the interop. I'll take care of this in a PR over in the CPP repo.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/UserNamePasswordParameters.java
Avery-Dunn: As of the latest commit most of the acceptance tests in that design doc are now implemented (the Mac/Linux tests are missing, as our pipelines only run in Windows environments anyway)

msal4j-brokers/src/test/java/test/ProofOfPossessionTest.java
```diff
@@ -0,0 +1,192 @@
...
+        IAuthenticationResult result = acquirePoPTokenUsernamePassword(pca, user, Collections.singleton(GRAPH_DEFAULT_SCOPE));
+    }
+
+    @Test
+    public void acquirePopToken_BrowserAndBroker() throws Exception {
+        User user = labUserProvider.getDefaultUser();
+
+        seleniumDriver = SeleniumExtensions.createDefaultWebDriver();
+
+        //First, get a non-PoP (bearer) token through a browser
```
bgavrilMS, line 62: Why not use non-WAM Username/Password? UI testing is brittle, should be avoided as much as possible.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -90,4 +90,6 @@ private ITenantProfile getTenantProfile() {
     private final Date expiresOnDate = new Date(expiresOn * 1000);
 
     private final String scopes;
+
+    private final Boolean isPopAuthorization;
```
bgavrilMS, line 95: Is this used?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAuthenticationResult.java
```diff
@@ -44,4 +44,9 @@ public interface IAuthenticationResult extends Serializable {
      * @return access token expiration date
      */
     java.util.Date expiresOnDate();
+
+    /**
+     * @return whether this is a proof-of-possession token
+     */
+    Boolean isPopAuthorization();
```
bgavrilMS: 1. Adding a member to an interface is a breaking change (since you'll break ppl using this interface). Can you use default methods instead? 

Alternative: deprecate the interface and just the data object AuthenticationResult, since data objects don't need to be put behind interfaces...

2. .NET uses "string TokenType" instead to differentiate "Bearer", "pop" and potentially others (SSH-cert is the other one).

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
```diff
@@ -44,24 +44,29 @@ default void removeAccount(PublicClientApplication application, IAccount account
...
+        AuthenticationResult.AuthenticationResultBuilder builder = AuthenticationResult.builder();
 
         try {
             if (idToken != null) {
                 builder.idToken(idToken);
-                if (accountId!= null) {
+                if (accountId != null) {
                     String idTokenJson =
                             JWTParser.parse(idToken).getParsedParts()[1].decodeToString();
                     //TODO: need to figure out if 'policy' field is relevant for brokers
```
bgavrilMS: No TODO in public code please. I guess the answer to this is no, but please open GH issues for tracking for TODOs that persist.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
```diff
@@ -216,4 +216,61 @@ protected Builder self() {
...
+        }
+
+        return brokerEnabled;
+    }
+
+    /**
+     * Used to determine whether to call into an IBroker instance instead of standard MSAL Java's normal username/password flow,
+     * and may throw exceptions or log messages if broker-only parameters are used when a broker is not enabled/available
+     */
+    private boolean validateBrokerUsage(UserNamePasswordParameters parameters) {
```
bgavrilMS, line 242: On top of this, the spec states to have an API like "IsPoPSupportedByClient" so that clients can have logic like 

- does RP support POP and does client support POP?

I suppose that `IsBrokerSupported` is public API and it can be used for the same?

msal4j-brokers/src/test/java/test/ProofOfPossessionTest.java
```diff
@@ -0,0 +1,192 @@
...
+import java.util.Collections;
+import java.util.Set;
+import java.util.UUID;
+import java.util.concurrent.ExecutionException;
+
+@Test()
+public class ProofOfPossessionTest {
+    private final static String MICROSOFT_AUTHORITY_ORGANIZATIONS =
+            "https://login.microsoftonline.com/organizations/";
+    private final static String GRAPH_DEFAULT_SCOPE = "https://graph.windows.net/.default";
```
bgavrilMS: Generally better to use "User.Read" scope from Graph, as it's quite harmless, as opposed to ".default" which gives access to "everything"

msal4j-brokers/src/test/java/test/ProofOfPossessionTest.java
```diff
@@ -0,0 +1,192 @@
...
+        User user = labUserProvider.getDefaultUser();
+
+        PublicClientApplication pca = createPublicClientApp(user);
+
+        //Setting UserNamePasswordParameters.proofOfPossession without enabling the broker should result in an exception when trying to get a token
+        IAuthenticationResult result = acquirePoPTokenUsernamePassword(pca, user, Collections.singleton(GRAPH_DEFAULT_SCOPE));
+    }
+
+    @Test
+    public void acquirePopToken_BrowserAndBroker() throws Exception {
```
bgavrilMS, line 57: This is a great test.

bgavrilMS, line 57: I would maybe also add a test where you get a Bearer token from WAM and a POP token from WAM.

MSAL should not cache either on its side.

bgavrilMS (approved):
Some comments, nothing blocking.

Would also like to double check that we have a test that checks: 

- get tokens using non-WAM (ROPC) and store AT and RT in cache
- get tokens using WAM silently. the AT and RT are NOT used at all. We should never fallback to non-WAM tokens in WAM flows.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
Avery-Dunn: Yes, it can be set when creating the AuthenticationResult in the IBroker file: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/8363f81cfd942e269758e9af1e2c2c1c53fca275/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java#LL83C20-L83C20

(the field here is 'private', but we use a dependency called Lombok to generate the getters/setters and other boilerplate code for these fields)

msal4j-brokers/src/test/java/test/ProofOfPossessionTest.java
Avery-Dunn: Changed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/PublicClientApplication.java
Avery-Dunn: Yes, there is an 'isBrokerAvailable' API that essentially just answers the question "could the MSALRuntime startup API be called successfully". 

Unless there is some scenario where POP isn't supported but MSALRuntime works, then for MSAL Java at least an `isBrokerSupported` API and an `isPoPSupported` API would always return the same thing.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IBroker.java
Avery-Dunn: Good catch, this was a leftover from the initial development. Removed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAuthenticationResult.java
Avery-Dunn: Good point about it being a breaking change. We've made a couple additive changes to these sorts of interfaces over the years but it's definitely a risk.

For the this beta release I simply got rid of this new field, and we can discuss more about deprecating interfaces and following best practices.

Issue #447, opened 2021-10-26
Interactive Browser Auth not working on Mac OS


g2vinay:
**Describe the bug**
When trying to authenticate with Azure via Interactive Browser from a Java/Sprint application, the following error is thrown:
```
Caused by: com.microsoft.aad.msal4j.MsalClientException: Unable to open default system browser
        at com.microsoft.aad.msal4j.AcquireTokenByInteractiveFlowSupplier.openDefaultSystemBrowser(AcquireTokenByInteractiveFlowSupplier.java:116)
...
```
And the application halts.

**To Reproduce**
* This was ran many times in OSX with the default browser (in the OS) set to Safari, Chrome or Edge.

***Code Snippet***
```java
        String tenantId = "<Tenant-Id>";
        String clientId = "<Client-Id>";

        String authHostUrl = "https://login.microsoftonline.com/" + tenantId;
        PublicClientApplication publicClientApplication = PublicClientApplication
                .builder(clientId).authority(authHostUrl)
                .build();

        List<String> scopes = new ArrayList<>(1);
        scopes.add("https://vault.azure.net/.default");
        InteractiveRequestParameters parameters =
                InteractiveRequestParameters.builder(URI.create("http://localhost:8080/"))
                        .scopes(new HashSet<>(scopes))
                        .prompt(Prompt.SELECT_ACCOUNT)
                        .build();

        String accessToken = publicClientApplication.acquireToken(parameters).get().accessToken();

```

**Expected behavior**
For my Java application to be able to use the default/any browser to construct a usable `PublicClientApplication` so that I can interactively authenticate.

**Setup (please complete the following information):**
 - OS: OSX Big Sur
 - msal4j:1.11.0 

Referencing original issue: https://github.com/Azure/azure-sdk-for-java/issues/21166

bgavrilMS:
We should fix these bugs. Browser-based auth is important. 

The logic for opening the browser on Windows Mac and Linxu is here: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5effd6c27477b5ffd22aac502420f13d06eec2ac/src/client/Microsoft.Identity.Client/Platforms/netcore/NetCorePlatformProxy.cs#L148

siddhijain:
This issue is fixed and will be available to test in the next release.

Avery-Dunn:
[...]

PR #651, opened 2023-05-11
Issue 447 - support for macOs, Linux and Unix 

siddhijain:
This fix allows opening a default browser depending on the operating system. In the absence of a Mac device, we do not have a way to test it.

https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/447

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
@@ -108,6 +108,17 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
     }
 
     private void openDefaultSystemBrowser(URL url) {
+        String os = System.getProperty("os.name").toLowerCase();
+        if (os.contains("windows")) {
+            openDefaultSystemBrowserInWindows(url);
+        } else if (os.contains("mac")) { // mac os
+            openDefaultSystemBrowserInMac(url);
+        } else if(os.contains("nux") || os.contains("nix")) { //linux or unix os
+            openDefaultSystemBrowserInLinux(url);
+        }
+    }
+
+    private static void openDefaultSystemBrowserInWindows(URL url){
         try {
             if (Desktop.isDesktopSupported() && Desktop.getDesktop().isSupported(Desktop.Action.BROWSE)) {
                 Desktop.getDesktop().browse(url.toURI());
@@ -118,6 +129,24 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
             }
         } catch (URISyntaxException | IOException ex) {
             throw new MsalClientException(ex);
+        }
+    }
+
+    private static void openDefaultSystemBrowserInMac(URL url){
+        Runtime runtime = Runtime.getRuntime();
+        try {
+            runtime.exec("open " + url);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static void openDefaultSystemBrowserInLinux(URL url){
+        Runtime runtime = Runtime.getRuntime();
+        try {
+            runtime.exec("xdg-open " + url);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
         }
     }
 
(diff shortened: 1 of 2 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -108,6 +108,17 @@ private void updateRedirectUrl() {
     }
 
     private void openDefaultSystemBrowser(URL url) {
+        String os = System.getProperty("os.name").toLowerCase();
```
bgavrilMS, line 111: nit cache the result to avoid all the string operations.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -108,6 +108,17 @@ private void updateRedirectUrl() {
     }
 
     private void openDefaultSystemBrowser(URL url) {
```
bgavrilMS: nit cache the result of OS discoevery (maybe put it in a helper, as you're likely to need elsewhere) to avoid all the string operations.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -108,6 +108,17 @@ private void updateRedirectUrl() {
...
+        String os = System.getProperty("os.name").toLowerCase();
+        if (os.contains("windows")) {
+            openDefaultSystemBrowserInWindows(url);
+        } else if (os.contains("mac")) { // mac os
```
bgavrilMS: I'll help test on Mac.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -121,6 +132,24 @@ private void openDefaultSystemBrowser(URL url) {
...
+        Runtime runtime = Runtime.getRuntime();
+        try {
+            runtime.exec("open " + url);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private static void openDefaultSystemBrowserInLinux(URL url){
+        Runtime runtime = Runtime.getRuntime();
```
bgavrilMS: I'd implement the full logic from .NET, took a long time to figure those things out: 

- check for sudo and fail
- try all the browser opening tools. xdg-open works on Ubuntu and a few distros, but not all.

siddhijain: Done in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
siddhijain: Good point. Added an OSHelper class in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
siddhijain: Thank you!

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OSHelper.java
```diff
@@ -0,0 +1,38 @@
+package com.microsoft.aad.msal4j;
+
+public class OSHelper {
+
+    private static String os;
+    private static boolean mac;
```
Avery-Dunn: Seems odd to have three booleans if only one of which will be true at a time. Maybe an enum with MAC, WINDOWS, and LINUX (and maybe 'UNKNOWN' if that could ever happen)? You could still have the isMac, isWindows, etc. methods, but just one field with the result of the static initializer.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OSHelper.java
siddhijain: makes sense. Updated in the latest commit.

Avery-Dunn (approved):

PR #661, opened 2023-06-20
Add initial code for MSI

### LINKED ISSUE #616: [Feature Request] Add App service managed identity with initial managed identity work
Add IMDS managed identity source.

- [x] Add initial managed identity code 
- [x] Add code for App service source
- [x]  Test the flow manually
- [ ]  Add unit tests
- [x]  Add error handling

For more details see: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/615

### LINKED ISSUE #617: [Feature Request] Add IMDS managed identity source to MSI integration
Add IMDS managed identity source.

- [x] Add code for IMDS source
- [x] Test the flow manually
- [ ] Add unit tests 
- [x] Add error handling 

For more details see: #615

neha-bhargava:
Included in this PR:
* Initial code for managed identity 
* Support for App service
* Support for IMDS

Testing
* Unit test
* Sample app for Azure functions to test manually


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
```diff
@@ -0,0 +1,131 @@
...
+
+    protected ManagedIdentityResponse getSuccessfulResponse(IHttpResponse response) {
+
+        ManagedIdentityResponse managedIdentityResponse = JsonHelper
+                .convertJsonToObject(response.body(), ManagedIdentityResponse.class);
+
+        if (managedIdentityResponse == null || managedIdentityResponse.getAccessToken() == null
+                || managedIdentityResponse.getAccessToken().isEmpty() || managedIdentityResponse.getExpiresOn() == null
+                || managedIdentityResponse.getExpiresOn().isEmpty()) {
+            LOG.error("[Managed Identity] Response is either null or insufficient for authentication.");
```
bgavrilMS, line 112: Shouldn't this be an exception?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -0,0 +1,98 @@
...
+                    MsalErrorMessage.SCOPES_REQUIRED);
+        }
+
+        TokenRequestExecutor tokenRequestExecutor = new TokenRequestExecutor(
+                clientApplication.authenticationAuthority,
+                msalRequest,
+                clientApplication.getServiceBundle()
+        );
+
+        if (!managedIdentityParameters.forceRefresh) {
```
bgavrilMS, line 38: Did we expose this flag on MSAL.NET in the public API? I'd rather not make this public from the get go and see if ppl have good scenarios for it.

CC @localden

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -0,0 +1,98 @@
...
+                        this.clientApplication,
+                        context,
+                        null);
+
+                AcquireTokenSilentSupplier supplier = new AcquireTokenSilentSupplier(
+                        this.clientApplication,
+                        silentRequest);
+
+                return supplier.execute();
+            } catch (MsalClientException ex) {
```
bgavrilMS, line 64: This is surprising. Exception should not control the programming flow. 

https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why

If this is the common pattern in MSAL 4J, at least there should be some filtering on ex.Code ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -0,0 +1,98 @@
...
+                .getManagedIdentityResponse(managedIdentityParameters);
+
+        AuthenticationResult authenticationResult = createFromManagedIdentityResponse(managedIdentityResponse);
+        clientApplication.tokenCache.saveTokens(tokenRequestExecutor, authenticationResult, clientApplication.authenticationAuthority.host);
+        return authenticationResult;
+    }
+
+    private AuthenticationResult createFromManagedIdentityResponse(ManagedIdentityResponse managedIdentityResponse) {
+        long expiresOn = Long.valueOf(managedIdentityResponse.expiresOn);
+        long refreshOn = expiresOn > 2 * 3600 ? (expiresOn / 2) : 0L;
```
bgavrilMS, line 93: nice

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -102,6 +102,8 @@ private static IHttpResponse executeHttpRequestWithRetries(HttpRequest httpReque
             }
             Thread.sleep(RETRY_DELAY_MS);
         }
+
+        log.info(httpResponse.body());
```
bgavrilMS: Please do not log HTTP responses, as they contain tokens. Not even as PII. This is a blocking comment.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -73,26 +77,38 @@ private HttpResponse executeHttpPost(HttpRequest httpRequest) throws Exception {
...
 
         connection.setConnectTimeout(connectTimeout);
         connection.setReadTimeout(readTimeout);
 
-        return connection;
+        if (connection instanceof HttpURLConnection) {
+            httpConnection = (HttpURLConnection) connection;
+            isHttp = true;
+        } else {
+            httpsConnection = (HttpsURLConnection) connection;
```
bgavrilMS: I don't think you need this `isHttp` flag, you can just put block at line 104 in here?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
```diff
@@ -87,4 +87,36 @@ private URL createUrlFromString(String stringUrl) {
 
         return url;
     }
+
+    // Adding this for unit tests to be able to compare expected and actual request.
+    @Override
```
bgavrilMS: Use https://projectlombok.org/features/EqualsAndHashCode ?

That said, ideally we should not change product code like this just to make it testable? We will forget about this and add HttpRequsts to some dictionary / map and get wierd bugs.

It'd be better to have a helper in tests that helps with assertions and which can access httpMethod, headers, body etc.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IEnvironmentVariables.java
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+public interface IEnvironmentVariables {
+    String IDENTITY_ENDPOINT = "IDENTITY_ENDPOINT";
```
bgavrilMS: Is this a common pattern to set constants on an interface? In any case, this interface is about env variables and the constants are about MSI. They should be separate?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
neha-bhargava: Yes, I will update it.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: We do expose force refresh in MSAL.Net for managed identity, [see](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs#L56). Should we exclude it for MSAL Java?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: This is how the flow is in client credential. I can add a filter on error code [cache_miss](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/eb36681986d3ae2432e438949efdb9086476e1c4/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationErrorCode.java#L59)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
```diff
@@ -0,0 +1,131 @@
...
+    }
+
+    public ManagedIdentityResponse handleResponse(
+            ManagedIdentityParameters parameters,
+            IHttpResponse response) {
+
+        String message;
+
+        try {
+            if (response.statusCode() == HttpURLConnection.HTTP_OK) {
```
localden, line 79: Can this ever be a different code and still return a good response, like HTTP 304?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
localden: I think we should keep parity with MSAL .NET and keep the force refresh option.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AppServiceManagedIdentity.java
```diff
@@ -0,0 +1,98 @@
...
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Map;
+
+class AppServiceManagedIdentity extends AbstractManagedIdentitySource{
```
localden: Is App Service managed identity implementation that different that it warrants its own code path?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
Avery-Dunn: Yeah, in both client credential and OBO this is the pattern that was added ~2 years ago when we first added this default cache lookup: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/368

Our silent flow will throw an exception if there's a cache miss so this is probably the simplest way to do it, but we can definitely do better messaging in all the places where this happens (filter on cache_miss error code like Neha said with a `LOG.info("abc")` message, and maybe a `LOG.error("xyz")` if there's a different failure?)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
neha-bhargava: I don't think so. There are a set status codes that an endpoint returns. We did not observe this behavior in testing. Also, when we implemented this in MSAL.Net most of the logic was copied from Azure.Identity along with this where we check 200.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: Agree with Den

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IEnvironmentVariables.java
Avery-Dunn: I'm not sure if it's "common", but I've seen interfaces in other Java projects that are just a bunch of fields and no methods. Kind of like an enum, but intended to be extended to be a normal class. The only other place we do this is in [IClientApplicationBase](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IClientApplicationBase.java#L18) where we have a default authority value, but that's only one field. 

All fields in an interface a public, static, and final, so unless this is meant to be a public API for app developers to use then it'd probably be better to havethese fields in a package-private enum (like how our [PublicApi](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/PublicApi.java) class stores telemetry codes)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -329,7 +333,7 @@ public abstract static class Builder<T extends Builder<T>> {
         private String azureRegion;
         private Integer connectTimeoutForDefaultHttpClient;
         private Integer readTimeoutForDefaultHttpClient;
-        private boolean instanceDiscovery = true;
+        protected boolean instanceDiscovery = true;
```
bgavrilMS: nit: boolean fields are easier to read when they are like "IsXYZEnabled"

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
```diff
@@ -0,0 +1,131 @@
...
+import java.beans.Encoder;
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.MalformedURLException;
+import java.net.URISyntaxException;
+
+//base class for all sources that support managed identity
+abstract class AbstractManagedIdentitySource {
+
+    protected static final String TIMEOUT_ERROR = "[Managed Identity] Authentication unavailable. The request to the managed identity endpoint timed out.";
```
bgavrilMS, line 25: The error code might be common, but the message is better to come from each individual source so that there is more context?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
bgavrilMS: I'll let you decide on the best solution here, as I realize there's been a lot of refactoring already. 

Catching a general exception like `MsalClientException` seems brittle - if either: RequestContext, SilentRequest, or AcquireTokenSilentSupplier starts to throw this exception for some other reason, it'll break this logic.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AppServiceManagedIdentity.java
bgavrilMS: Yeah, that's the worse part of MSI - there are 5 different implemenations depending on the compute used - Imds, ServiceFabric, AppService, Arc and CloudShell. 

And the logic to figure out which compute we're on is both brittle and slow, as it's based on reading a bunch of env variables and on making local HTTP calls. 


But my understanding is that with SLC, this will more or less be fixed. @gladjohn to keep me honest.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IEnvironmentVariables.java
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+public interface IEnvironmentVariables {
```
bgavrilMS: This interface should not be public API

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -0,0 +1,98 @@
...
+
+package com.microsoft.aad.msal4j;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.util.HashSet;
+import java.util.Set;
+
+public class AcquireTokenByManagedIdentitySupplier extends AuthenticationResultSupplier {
```
bgavrilMS: This should not public API. The other "supplier" classes are not public API.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/EnvironmentVariables.java
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+public class EnvironmentVariables implements IEnvironmentVariables {
```
bgavrilMS: not public api.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
```diff
@@ -0,0 +1,156 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.net.HttpURLConnection;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class IMDSManagedIdentity extends AbstractManagedIdentitySource{
```
bgavrilMS: 1. Not public
2. nit: Naming - `ImdsManagedIdentitySource` to be consistent with .NET?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
```diff
@@ -0,0 +1,156 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
```
bgavrilMS: The flat packaging in MSAL 4J is not great. Makes the code hard to navigate. Can you create a "folder" at least for MI?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
```diff
@@ -0,0 +1,156 @@
...
+        ManagedIdentityParameters parameters = (ManagedIdentityParameters) msalRequest.requestContext().apiParameters();
+        IEnvironmentVariables environmentVariables = ((ManagedIdentityParameters) msalRequest.requestContext().apiParameters()).environmentVariables == null ?
+                new EnvironmentVariables() :
+                parameters.environmentVariables;
+        if (!StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(IEnvironmentVariables.AZURE_POD_IDENTITY_AUTHORITY_HOST))){
+            LOG.info("[Managed Identity] Environment variable AZURE_POD_IDENTITY_AUTHORITY_HOST for IMDS returned endpoint: " + environmentVariables.getEnvironmentVariable(IEnvironmentVariables.AZURE_POD_IDENTITY_AUTHORITY_HOST));
+            try {
+                imdsEndpoint = new URI(environmentVariables.getEnvironmentVariable(IEnvironmentVariables.AZURE_POD_IDENTITY_AUTHORITY_HOST));
+            } catch (URISyntaxException e) {
+                throw new RuntimeException(e);
```
bgavrilMS: Just for my understanding - we're going with RuntimeException here to keep the code clean? We never expect `URISyntaxException` to occur, so we don't to have it as a checked exception in the public API? (similar to NullPointerException )

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
```diff
@@ -0,0 +1,156 @@
...
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.HashMap;
+import java.util.Map;
+
+public class IMDSManagedIdentity extends AbstractManagedIdentitySource{
+
+    // IMDS constants. Docs for IMDS are available here https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/how-to-use-vm-token#get-a-token-using-http
+    private final static Logger LOG = LoggerFactory.getLogger(IMDSManagedIdentity.class);
+    private static URI DEFAULT_IMDS_ENDPOINT;
```
bgavrilMS: final

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
```diff
@@ -0,0 +1,156 @@
...
+
+    static {
+        try {
+            DEFAULT_IMDS_ENDPOINT = new URI("http://169.254.169.254/metadata/identity/oauth2/token");
+        } catch (URISyntaxException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private String imdsTokenPath = "/metadata/identity/oauth2/token";
```
bgavrilMS: 1. Please review the "static final" modifiers throughout. Is there an analyzer that can help @Avery-Dunn ?
2. Consider having all the constants and fields in one place

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -0,0 +1,91 @@
...
+import lombok.Getter;
+import org.slf4j.LoggerFactory;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Class to be used to acquire tokens for managed identity.
+ * For details see {@link IManagedIdentityApplication}
+ * <p>
+ * Conditionally thread-safe
```
bgavrilMS, line 16: What does this mean?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -0,0 +1,91 @@
...
+ */
+public class ManagedIdentityApplication extends AbstractClientApplicationBase implements IManagedIdentityApplication {
```
bgavrilMS, line 18: Should this be "final" ? in general, should new public API classes be final to prevent ppl from inheriting?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -0,0 +1,91 @@
...
+
+    private String resource;
+
+    @Getter
+    private ManagedIdentityId managedIdentityId;
```
bgavrilMS: is this "final" ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -0,0 +1,91 @@
...
+                        PublicApi.ACQUIRE_TOKEN_BY_USER_ASSIGNED_MANAGED_IDENTITY,
+                managedIdentityParameters);
+
+        ManagedIdentityRequest managedIdentityRequest = new ManagedIdentityRequest(this, requestContext);
+
+        return this.executeRequest(managedIdentityRequest);
+    }
+
+    @Override
+    protected ClientAuthentication clientAuthentication() {
```
bgavrilMS, line 58: What does this do? protected means it's part of the public api

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityIdType.java
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+public enum ManagedIdentityIdType {
```
bgavrilMS: 1. Comments on public APIs
2. Should we try to keep naming similar to .NET, unless Java idioms dictate otherwise? `ManagedIdentityId` is the .NET type

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -0,0 +1,88 @@
...
+
+    public boolean isForceRefresh() {
+        return forceRefresh;
+    }
+
+    public void setForceRefresh(boolean forceRefresh) {
+        this.forceRefresh = forceRefresh;
+    }
+
+    @Override
```
bgavrilMS, line 47: Are these (scopes, claims etc.) supposed to be removed from the public API by the PR from @Avery-Dunn ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -0,0 +1,88 @@
...
+@AllArgsConstructor(access = AccessLevel.PRIVATE)
+public class ManagedIdentityParameters implements IAcquireTokenParameters {
+
+    String resource;
+    
+    boolean forceRefresh;
+
+    IEnvironmentVariables environmentVariables;
+
+    public String getResource() {
```
bgavrilMS, line 31: Do you need to add getters and setters if you used Lombok with builder / accessors etc.  ?

I would think that for public APIs you don't use Lombok magic: 

- it's safer as you can see the actual API 
- you are able to add comments

Thoughts @Avery-Dunn ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRequest.java
```diff
@@ -0,0 +1,56 @@
...
+        String endpoint = this.appendQueryParametersToBaseEndpoint();
+        try {
+            return new URL(endpoint);
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    private String appendQueryParametersToBaseEndpoint(){
+        StringBuilder stringBuilder = new StringBuilder(baseEndpoint.toString());
```
bgavrilMS, line 39: Do you need to do this manually ? Isn't there URL / URI methods that help with this?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRequest.java
```diff
@@ -0,0 +1,56 @@
...
+
+class ManagedIdentityRequest extends MsalRequest {
+
+    URI baseEndpoint;
+
+    HttpMethod method;
+
+    Map<String, String> headers;
+
+    Map<String, String> bodyParameters;
```
bgavrilMS, line 21: headers and body parameters are not used.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalError.java
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+class MsalError {
```
bgavrilMS: This one is actually public ? ;) Nice to see error codes constants.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalErrorMessage.java
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+public class MsalErrorMessage {
```
bgavrilMS: I'd keep this non-public.? We're always changing these "constants".

msal4j-sdk/src/samples/msi-sample/pom.xml
```diff
@@ -0,0 +1,120 @@
...
+        <azure.functions.maven.plugin.version>1.24.0</azure.functions.maven.plugin.version>
+        <azure.functions.java.library.version>3.0.0</azure.functions.java.library.version>
+        <functionAppName>msi-function-java-1687284411707</functionAppName>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>msal4j</artifactId>
+            <version>1.13.8</version>
```
bgavrilMS, line 24: Does the MSI ~sample~  dev app use MSAL4J 1.13.8 fixed version?

msal4j-sdk/src/samples/msi-sample/src/main/java/com/microsoft/identity/msi/HttpTriggerJava.java
```diff
@@ -0,0 +1,52 @@
...
+        if (resource == null) {
+            return request.createResponseBuilder(HttpStatus.BAD_REQUEST).body("Please pass a resource to acquire token for").build();
+        }
+
+        String response;
+
+        try {
+            ManagedIdentityApplication msiApp = ManagedIdentityApplication
+                    .builder(ManagedIdentityId.systemAssigned())
+                    .logPii(true)
```
bgavrilMS, line 40: Q: is the context.GetLogger injected here? How do we pull MSAL 4J logs in an app like this? (comment mostly for @Avery-Dunn )

bgavrilMS (changes requested):
Needs changes particularly on visibility modifiers / public API comments etc. Logic looks good.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -0,0 +1,323 @@
...
+                .httpClient(httpClientMock)
+                .build();
+
+        try {
+            miApp.acquireTokenForManagedIdentity(
+                    ManagedIdentityParameters.builder(resource)
+                            .environmentVariables(environmentVariables)
+                            .build()).get();
+        } catch (Exception exception) {
+            assert(exception.getCause() instanceof MsalManagedIdentityException);
```
bgavrilMS, line 283: Is this what we do with MsalServiceException as well?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
neha-bhargava: Accidental check in

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: Changed for managed identity. Not updating for client credentials flow in this PR.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
neha-bhargava: I did not find a way to add helpers to compare the request in tests since the request is framed in the flow in private methods. This is how the mocking is in tests `when(httpClientMock.send(eq(expectedRequest(source, resource)))).thenReturn(expectedResponse(200, ""));`

I have added the annotation, but not sure if there is a way to write a helper to verify the request.

Another option is to remove the validation for request.

@Avery-Dunn to help confirm if there is any other way.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IEnvironmentVariables.java
neha-bhargava: These constants are specific to Environment Variables so these are here. I can move these to constant file

neha-bhargava: Moved to constant file

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
neha-bhargava: Making it package private.
Missed the naming. Thanks for pointing it out.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
neha-bhargava: I am not sure if this can be done. The issue will be having to reference those classes out of that package since the scope will be package private and not public.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
neha-bhargava: This should be MsalManagedIdentityException. Ideally this will never happen. Updating here and rest of the places.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: The comment is a copy paste from ConfidentialClientApplication. Not sure if it is a common practice in Java @Avery-Dunn do you know?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: I agree that this should be final. but the other public API classes are not. @Avery-Dunn maybe have this as part of refactoring?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: This is further in the flow needed to build the OAuth request and I think not needed for managed identity. https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/6106d521e6760bd5cf17a7b32f410e9e125fd4f1/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java#L81

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityIdType.java
neha-bhargava: This is same as .Net naming wise https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityIdType.cs#L12. I removed the public keyword. It should be package private.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
neha-bhargava: Yes, this should be done as part of refactoring.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRequest.java
neha-bhargava: Headers is used in IMDS and AppService. The bodyParameters will be used in one of the sources that have POST method. So added it in advance.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MsalError.java
neha-bhargava: Updating with comments

msal4j-sdk/src/samples/msi-sample/pom.xml
neha-bhargava: No, the latest jar that is created.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: I think so, in Java the exception thrown from CompletableFuture is wrapped in ExecutionException

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
Avery-Dunn: Does .NET have error messages for each source? If so we could probably just copy them.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
Avery-Dunn: Since the URL, HTTP method, headers, and body are all fields in HTTPRequest, then Lombok's EqualsAndHashCode annotation should create an equals method that's looks like the return method you have here.

If the comparisons still aren't working, then it might be cleaner to just have this method in the relevant tests like @bgavrilMS said.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
Avery-Dunn: This is something that's been talked about before, I'm not sure why it was originally designed this way but it certainly can be confusing. 

For now I think it's best to keep the MI implementation as similar to the rest of the library as possible, and focus on improving the entire library structure in another PR.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
Avery-Dunn: If you're talking about analyzers to find minor issues/make code suggestions I use IntelliJ's 'SonarLint' plugin, it's pretty good at making best-practices suggestions out of the box.

MSAL Java is already riddled with stuff like 'static final' vs. 'final static' so it's not too big of an issue, but I've been fixing those sorts of things whenever SonarLint highlights them in a file I was changing anyway.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
Avery-Dunn: The original commit that added that 'Conditionally thread-safe' language in a few places was part of a large commit of comments/javadocs updates: https://github.com/AzureAD/microsoft-authentication-library-for-java/commit/ef3078b7af247e51f093d81d0420e72154193dec#diff-54124b221a96bb4edb1daaac28130dc448b82d9b9d024dca2d520ee355f0b740R23

Unfortunately it happened a couple weeks before I joined so I have no context for it. It's also used in the TokenCache class, so maybe it's a reference to how we use read/write locks on the token cache?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
Avery-Dunn: Is there are particular issue that could be cause by people inheriting this ManagedIdentityApplication class? As @neha-bhargava said none of the other classes are final.

From a strict Java best-practices standpoint unless there's a planned use case for inheritance then the class should be final, though it might be confusing for only this class to get singled out like that.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
Avery-Dunn: This would be a similar answer to https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/661/files/d7364ba1a13185751f231ea9aaaef2deab3a8ede#r1311659704, but yeah getters aren't needed if you have Lombok's `@Getter` annotation, and the setters aren't needed if you have the `@Builder` annotation. These sorts of issues will be solved in the follow-up refactoring PR.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRequest.java
Avery-Dunn: We use a helper method from Nimbus called `URLUtils.serializeParameters()` to serialize parameters into a string in a few places (see [here](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/4dc7a5e3c29e3995c585648b86cb3a04aef3c39f/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java#L79) for an example of serializing query params), but it takes in a `Map<String, List<String>>` instead of a `Map<String, String>`. This means we'd need to refactor every instance of `queryParameters.put("key", "value")` to `queryParameters.put("key", Collections.singleton("value"))`, which is clunky but we actually do it quite a lot in the library already...

msal4j-sdk/src/samples/msi-sample/pom.xml
Avery-Dunn: Through IDE magic IntelliJ and Maven will either grab the version of the code you have in the editor or the version you built locally rather than trying to get the 'real' 1.13.8 version from the public package repo. We usually update these version numbers as part of a release, and for development they just need to match the version number in your local copy of msal4j's pom.xml

msal4j-sdk/src/samples/msi-sample/src/main/java/com/microsoft/identity/msi/HttpTriggerJava.java
Avery-Dunn: You can find some docs on logging here (it includes a step about spring for some reason, I think that's a mistake): https://learn.microsoft.com/en-us/entra/msal/java/advanced/msal-logging-java

In short, whenever we want to to log something we do it through the [Simple Logging Facade for Java](http://www.slf4j.org/). You've probably seen code like `LOG.info("abc")` and `LOG.error("xyz")`, those are all getting sent through SLF4J. An app developer can use their preferred logging framework, and as long as it's compatible with SLF4J then it will output our logs through that framework in the same way as their own logs.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
Avery-Dunn: I think CompletableFuture will throw a `CompletionException` instead of an `ExecutionException` in some cases, but either way the actual exception that caused it to fail will be wrapped up in that `Exception` you're catching.

We do similar stuff in other tests about exceptions, for example in [AcquireTokenSilentlyTests](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/4dc7a5e3c29e3995c585648b86cb3a04aef3c39f/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java#L30)

Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -673,7 +677,7 @@ public T azureRegion(String val) {
         yet still want MSAL to accept any authority that you will provide,
         you can use a ``False`` to unconditionally disable Instance Discovery. */
         public T instanceDiscovery(boolean val) {
```
bgavrilMS, line 679: Should not be part of the Managed Identity apis... Not sure if it is or not.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -0,0 +1,323 @@
...
+    @ParameterizedTest
+    @MethodSource("com.microsoft.aad.msal4j.ManagedIdentityTestDataProvider#createData")
+    void managedIdentityTest_DifferentScopes_RequestsNewToken(ManagedIdentitySourceType source, String endpoint) throws Exception {
+        String resource = "https://management.azure.com";
+        String anotherResource = "https://graph.microsoft.com";
+
+        IEnvironmentVariables environmentVariables = new EnvironmentVariablesHelper(source, endpoint);
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+
+        lenient().when(httpClientMock.send(eq(expectedRequest(source, resource)))).thenReturn(expectedResponse(200, getSuccessfulResponse(resource)));
```
bgavrilMS, line 178: Can you check that all mocks are consumed? Or some other check that the 2 tokens are not identical?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -0,0 +1,323 @@
...
+import java.util.HashMap;
+import java.util.Map;
+
+import static org.junit.jupiter.api.Assertions.*;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+
+@ExtendWith(MockitoExtension.class)
+@TestInstance(TestInstance.Lifecycle.PER_METHOD)
+public class ManagedIdentityTests {
```
bgavrilMS, line 25: Can you add a test that checks the token cache keys if one doesn't exist? We should make sure that AT key is consistent with .NET for SAMI and for UAMI. I understand it's only in memory.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
neha-bhargava: Updated the HttpRequest to add Lombok's EqualsAndHashCode and it works.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSManagedIdentity.java
neha-bhargava: Agree this is something that will not be a breaking change and can be achieved in another PR.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: We can add final to all the applications but do not want to do this as part of this PR.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: Read a bit about this, https://www.copperspice.com/docs/cs_api/thread-safety.html#:~:text=The%20term%20conditionally%20thread%20safe%20is%20a%20weaker,words%2C%20the%20threads%20are%20not%20sharing%20any%20resources. Basically means different threads will not share any resources. I think it is OK to keep this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
neha-bhargava: It is not. This is part of Abstract class. Will be taken care of in refactoring

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: This task will become simpler with addition of token source. I will add a TODO here until then.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: There is a difference in tenant id that is used in Java to build the cache key vs the tenant id used in .Net. Other than that the cache key is same. Will need some more work to make them same. Something I think need to be consistent with .Net https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Instance/Authority.cs#L45, rules to set the tenant id.

neha-bhargava: Wil create a task on board for this.

Issue #652, opened 2023-05-31
Problem on testNG : CVE-2022-4065


MustanK57:
Hello, 

Microsoft want the migration from ADAL to MSAL, but this migration is blocked by this vulnerability. 
This vulnerability will be resolve soon (before the end of June ?).

Thank you

bgavrilMS:
@Avery-Dunn  -  I see MSAL Java using testng 7.1 - I assume bumping to testng 7.7 should not be a problem. I see that DependaBot proposed a fix https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/648 but the PR was closed without comments. Can you please add your reasoning for this?

@MustanK57 - thanks for pointing out, we'll look at this ASAP. Do you know what the attack vector is, since this is a test dependency only?

MustanK57:
Vulnerability scan detect it on the download, and I need to be sure if it can cause a problem while I use this package.

bgavrilMS:
Ack, an attack vector in unlikely, but security tools prevent from using latest MSAL. We'll fix this with priority (~definitely~ hopefully before end of June).

bgavrilMS:
@siddhijain investigated this and can provide a few options. Seems like it's not an easy fix.

bgavrilMS:
@DidunAyodeji @localden  - I believe this may be higher priority than brokers or Managed Identity, as folks will not pick up a package with a CVE label.

siddhijain:
It is not a simple fix to upgrade the version of this library as the latest secured version of test-ng is on Java 11 and msal java is built using Java 8. 

We can transition from TestNG to a different framework, such as JUnit, which is widely recognized as one of the most popular testing frameworks in the Java world. However, it's important to note that the latest version of JUnit (JUnit 5) does not support the mocking framework currently used in MSAL Java (PowerMock). We have two potential approaches: either we can shift to an older version of JUnit (JUnit 4), or we can consider migrating the MSAL Java library to a mocking framework compatible with JUnit 5.

From my research, the rest of the migration from test-ng to JUnit should be simple.

localden:
@siddhijain @bgavrilMS what is the support expectation around JUnit 4? Want to make sure that we're not taking a dependency on a library that will be deprecated. I see on GitHub that the latest release was in 2021.

I am leaning towards option (2) of upgrading both the testing and mock frameworks, but would also like to understand the cost and timeline implications here.

bgavrilMS:
[...]

PR #672, opened 2023-07-07
Test framework update

Avery-Dunn:
As per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/652, older versions of testng have a vulnerability, and newer versions no longer support Java 8.

This PR changes tests to use JUnit 5 and Mockito:
 - Roughly half the test files didn't use any mocking, and only needed some minor changes to imports, annotations, and assertions
    - The only exception were the `AssertionCredentialTest` and `ClientSecretTest` files, which were so similar that I just combined them into one `ClientCredentialTest` file just for better readability
 - Tests that used Powermock/Easymock to mock public and static methods mostly had a 1:1 equivalent with Mockito, and for the most part were converted without major changes
 - Tests that mocked private methods had no Mockito equivalent, and Mockito has explicitly said they will not support it, so these were either rewritten or removed. Of particular note:
   -  `InstanceDiscoveryTest` shared a lot of coverage with `AadInstanceDiscoveryTest` so they were merged, and some methods in `AadInstanceDiscoveryProvider` were changed to package-private to keep them non-public but allow unit tests
   - `ConfidentialClientApplicationUnitT` and `PublicClientApplicationTest` were both removed. They relied heavily on private mocking and would've needed a complete rewrite, however the integration tests cover everything the unit tests were doing so it did not seem worth rewriting these tests
   - `DefaultHttpClientTest`, `OAuthRequestValidationTest`, `OauthHttpRequestTest`, and `OAuthRequestValidationUnitT` were all removed. They relied heavily on private mocking, and were mainly mocking the sort of HTTP requests/responses and parsing that all of the integrations tests covered

Despite the removal of several test files, test coverage was barely affected. According to IntelliJ's built-in coverage check:
Framework | Class coverage | Method coverage | Line coverage
-----| ---------| ---------| ---------|
TestNG | 89.3% (151/169) | 81.2% (830/1022) | 82.2% (3192/3884)
JUnit | 89.3% (151/169) | 80.2% (819/1021) | 80.5% (3125/3883)


bgavrilMS (approved):

bgavrilMS:
LGTM, seems like there is some Selenium / UI issue that crashes most tests.

msal4j-sdk/pom.xml
```diff
@@ -68,29 +68,41 @@
...
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.mockito</groupId>
+            <artifactId>mockito-junit-jupiter</artifactId>
+            <version>4.7.0</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>net.bytebuddy</groupId>
```
bgavrilMS, line 101: How is this runtime code generator used? Is this replacing mocking of private methods?

Avery-Dunn: Yes, it's needed to support the more advanced mocking options in `mockito-inline`. The `mockito-inline` package seems to require it (based on error messages when not using it), but it doesn't have it as a dependency for some reason so I needed to pull it in separately.

This `byte-buddy` and `mockito-inline` combo is [the same used by Azure SDK](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/pom.xml#L101). It seems like [mockito-inline](https://mvnrepository.com/artifact/org.mockito/mockito-inline) is a side project of useful features (like private mocking and certain annotations) that will be in the main mockito package one day.

However, while looking at your comment I noticed I forgot to scope it to just be a 'test' dependency, and I fixed that in the latest commit.

Issue #679, opened 2023-07-19
Performance: Account class didn't implement equals() and hashCode() => wrong results in getAccounts().join().contains()


gterminator:
Hi,
the class Account didnt implement the equals() and hashCode() methods. Therefore the Set cannot determine that an object is in or not.

**Problem**
`
boolean isAccountInCache = app.getAccounts().join().contains(myAccount);
`
This is not working because of the missing methods. As a Workaround, we have to iterate trough every element in this cache. Which make this library very inefficient and latency expensive.

**Problematic Workaround**
`
boolean isAccountInCache = app.getAccounts().join().stream().anymatch( acc -> acc.username().equals(myusername));
` 
This is working well for ~100 accounts in the cache, but in larger applications with more than 10-100k users in cache this will be a problem. Imagine we iterate for every request 10k times through this accounts set. 

Maybe I find time to open a PR and/or provide a PoC of this problem.

Best regards

Max

gterminator:
**Here the PoC**
```
public class Main
{


    public static void main(String[] args) throws InterruptedException
    {
        Set<Account> accountsBuggy = getExampleCacheOfAccounts(1000);
        Set<AccountFixed> accountsFixed = getExampleCacheOfAccountsFixed( 1000);

        /**
         * JDK normally hashcodes the object reference. In a distributed environment we didn't have the
         * object reference on another server...
         * Creating here a new Object with the same name but another reference.
         */
        Account account = Account.builder()
                                 .name("Max")
                                 .build();

        System.out.println("--Running bug Testcase--");
        boolean isAccountInSet = accountsBuggy.contains( account );
        System.out.println("Is account in Set= " + isAccountInSet + " expected=true");



        System.out.println("--Running fixed Testcase--");
        AccountFixed accountFixed = AccountFixed.builder()
                                 .name("Max")
                                 .build();

        boolean isAccountFixedInSet = accountsFixed.contains( accountFixed );
        System.out.println("Is accountFixed in Set= " + isAccountFixedInSet + " expected=true");




    }


    public static Set<Account> getExampleCacheOfAccounts(int amount)
    {
        Random random = new Random();
        HashSet<Account> accountHashSet = new HashSet<>();

        // example account for fetching
        Account account = Account.builder()
                                 .name("Max")
                                 .build();
        accountHashSet.add(account);

        for (int i = 0; i < amount; i++)
        {
            Account dummyAccount = Account.builder()
[...]

PR #681, opened 2023-07-19
Issue-679: Fix for Account Cache; .contains() was not possible and yo…

gterminator:
…u had to iterate through all elements as workaround.

#679


gterminator:
PoC see #679.

Avery-Dunn:
As discussed in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/679 the number of accounts in the cache should usually be low enough that iterations over it are quick. However, getting the contains() method to behave like you'd expect it to is common sense, so I see no reason not to add this even if it's not strictly necessary.

Thanks for making this PR, and sorry for the delay on addressing it. It will be in the next release, sometime this week.

Issue #682, opened 2023-07-26
AuthenticationResultSupplier fails on Linux - Caused by: java.io.IOException: Cannot run program "xdg-openhttps://login.microsoftonline.com


ricemery:
The AuthenticationResultSupplier.get method fails on Linux with attached exception. The bug is the call to Runtime.exec at line 185 of AuthenticationResultSupplier. The call is passed an invalid command  to execute. The command is passed without a space separating the command from the command arguments. `"xdg-openhttps://login.microsoftonline.com` should be `"xdg-open https://login.microsoftonline.com`. 

```
runtime.exec(openTool +  url);
```
should be changed to
```
runtime.exec(openTool + " " + url);
```
 
Stack Trace -
```
java.util.concurrent.CompletionException: java.lang.RuntimeException: java.io.IOException: Cannot run program "xdg-openhttps://login.microsoftonline.com/XXX/oauth2/v2.0/authorize?scope=openid+profile+offline_access+api%3A%2F%2F4XXX&response_type=code&redirect_uri=http%3A%2F%2Flocalhost%3A42459&code_challenge_method=S256&state=XXX&code_challenge=XXX&client_id=XXX&response_mode=form_post": error=2, No such file or directory
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:99)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:18)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1768)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1760)
	at java.base/java.util.concurrent.ForkJoinTask.doExec$$$capture(ForkJoinTask.java:373)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1182)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1655)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1622)
[...]

PR #683, opened 2023-07-26
Add space between command and arguments when executing linux command …

ricemery:
…to open browser. Refs #682

Fix for issue #682


Avery-Dunn (approved):
Looks good to me, thanks for identifying and fixing this!

Issue #669, opened 2023-06-30
NPE occurs if account has no localAccountId


GregDThomas:
**Steps to reproduce**
- Run the example at https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/samples/public-client/UsernamePasswordFlow.java using the details of an account that has no `localAccountId`.
  - Note that https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/3336 suggests that a localAccountId is not always created. 
  
**Expected results**
 - New token is generated, and then re-used as it's in the cache

**Actual results**
 - New token is generated, then an NPE occurs as there is no `localAccountId` and the following statement throws an NPE;
   - `if (accCached.homeAccountId().contains(accCached.localAccountId())) {`

```
==No accounts in cache
17:46:47.494 [ForkJoinPool.commonPool-worker-3] DEBUG com.microsoft.aad.msal4j.PublicClientApplication - [Correlation ID: cd8843c4-5fc5-4d42-9d65-e6e7675f042b] Execution of class com.microsoft.aad.msal4j.AcquireTokenSilentSupplier failed.
com.microsoft.aad.msal4j.MsalClientException: Token not found in the cache
	at com.microsoft.aad.msal4j.AcquireTokenSilentSupplier.execute(AcquireTokenSilentSupplier.java:75)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:69)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:18)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(CompletableFuture.java:1700)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.exec(CompletableFuture.java:1692)
	at java.base/java.util.concurrent.ForkJoinTask.doExec(ForkJoinTask.java:290)
	at java.base/java.util.concurrent.ForkJoinPool$WorkQueue.topLevelExec(ForkJoinPool.java:1020)
	at java.base/java.util.concurrent.ForkJoinPool.scan(ForkJoinPool.java:1656)
	at java.base/java.util.concurrent.ForkJoinPool.runWorker(ForkJoinPool.java:1594)
	at java.base/java.util.concurrent.ForkJoinWorkerThread.run(ForkJoinWorkerThread.java:183)
[...]

Issue #664, opened 2023-06-29
[BUG] [Repair Item] Remove client timeouts for connecting to the STS


bgavrilMS:
Today, MSAL Java times out after 15s. This timeout should be removed. AAD has its own timeouts which are not public.

Avery-Dunn:
This was fixed as of https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/684, and part of the 1.13.10 release.

PR #684, opened 2023-07-26
Assorted fixes

Avery-Dunn:
This PR handles a few small issues

Remove default HTTP timeout behavior, and improve error messages related to timeouts https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/664

Fix NPE that can occur when attempting to use `getAccounts` call with on-prem ADFS users in the cache https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/669

A version of a test dependency, [Guava ](https://mvnrepository.com/artifact/com.google.guava/guava) 32.0.0, has a vulnerability, so this PR also updates that dependency to version 32.1.1

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
@@ -1,5 +1,8 @@
 package com.microsoft.aad.msal4j;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLSocketFactory;
 import java.io.DataOutputStream;
@@ -14,6 +17,7 @@ import java.nio.charset.StandardCharsets;
 import java.util.Map;
 
 class DefaultHttpClient implements IHttpClient {
+    private final static Logger LOG = LoggerFactory.getLogger(DefaultHttpClient.class);
 
     private final Proxy proxy;
     private final SSLSocketFactory sslSocketFactory;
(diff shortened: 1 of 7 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -346,7 +346,7 @@ Set<IAccount> getAccounts(String clientId) {
                         ((Account) rootAccounts.get(accCached.homeAccountId())).tenantProfiles.put(accCached.realm(), profile);
                     }
 
-                    if (accCached.homeAccountId().contains(accCached.localAccountId())) {
+                    if (accCached.localAccountId() != null && accCached.homeAccountId().contains(accCached.localAccountId())) {
```
bgavrilMS, line 349: Seems like a separate change?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -117,6 +118,10 @@ private HttpResponse readResponseFromConnection(final HttpsURLConnection conn) t
             httpResponse.addHeaders(conn.getHeaderFields());
             httpResponse.body(inputStreamToString(is));
             return httpResponse;
+        } catch (SocketTimeoutException readException) {conn.getURL();
```
bgavrilMS: Changing exception behavior is tricky because exceptions are part of the public api. I would think many apps have retry policies that look for `SocketTimeout` and `Connect` exceptions, and this change will break the retry policies.

Maybe log.error these messages, but don't change exception type.

bgavrilMS (changes requested):
Let HTTP exceptions flow through

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
Avery-Dunn: Good point, in the latest commit I log the message explaining the error but the original exception gets re-thrown.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java
Avery-Dunn: This PR was meant to handle changes for all the small issues/enhancements we've gotten over the last month, but since the PRs made by customers didn't need additional changes this PR ended up handling just the timeout behavior and the NPE exception reported in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/669

Since it was just a one-line change, I figured I'd just add it into this PR anyway.

Issue #699, opened 2023-08-22
Provide a mechanism for static token caching


bgavrilMS:
See .NET equivalent `WithTokenCacheOptions(CacheOptions.SharedTokenCache)`

Issue #697, opened 2023-08-21
Provide a datapoint that shows that the token comes from the cache


bgavrilMS:
MSAL.NET has an additional datapoint in `AuthenticationResult.AuthenticationResultMetadata.TokenSource` (which may not be the best naming, but the team wanted an object to be able to add more properties there) which can be one of "IdP", "Cache", "Broker". 

This will help app developers understand where the token comes from.

Proposal for MSAL Java - put this in AuthenticationResult object directly. It's ok to have just 2 values for now - IdP and Cache. We can add Broker later.

CC @rayluo  as this should also be done in MSAL PY.

Avery-Dunn:
Token source metadata was added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/722 and was released along with MSI support, and cache refresh reason metadata and telemetry improvements were added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/901 and released in version 1.19.0

PR #708, opened 2023-09-11
Cache improvements

Avery-Dunn:
Adds an option to use a static/shared cache in ConfidentialClientApplication instead of the per-instance cache (https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/699), and adds a new field to AuthenticationResult to store helpful metadata about the result (https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/697)

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -69,6 +69,10 @@ class AcquireTokenSilentIT {
         // Check that access and id tokens are coming from cache
         assertEquals(result.accessToken(), acquireSilentResult.accessToken());
         assertEquals(result.idToken(), acquireSilentResult.idToken());
+        assertEquals(TokenSource.IDENTITY_PROVIDER, result.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, result.metadata().cacheRefreshReason());
+        assertEquals(TokenSource.CACHE, acquireSilentResult.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, acquireSilentResult.metadata().cacheRefreshReason());
     }
 
     @ParameterizedTest
@@ -92,6 +96,10 @@ class AcquireTokenSilentIT {
 
         // Check that new refresh and id tokens are being returned
         assertTokensAreNotEqual(result, resultAfterRefresh);
+        assertEquals(TokenSource.IDENTITY_PROVIDER, result.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, result.metadata().cacheRefreshReason());
+        assertEquals(TokenSource.IDENTITY_PROVIDER, resultAfterRefresh.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.FORCE_REFRESH, resultAfterRefresh.metadata().cacheRefreshReason());
     }
 
     @ParameterizedTest
@@ -253,6 +261,11 @@ class AcquireTokenSilentIT {
         //Current time is after refreshOn, so token should be refreshed
         assertNotNull(resultSilentWithRefreshOn);
         assertTokensAreNotEqual(resultSilent, resultSilentWithRefreshOn);
+
+        assertEquals(TokenSource.CACHE, resultSilent.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, resultSilent.metadata().cacheRefreshReason());
+        assertEquals(TokenSource.IDENTITY_PROVIDER, resultSilentWithRefreshOn.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.PROACTIVE_REFRESH, resultSilentWithRefreshOn.metadata().cacheRefreshReason());
     }
 
     @ParameterizedTest
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TokenCacheIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TokenCacheIT.java
@@ -8,7 +8,9 @@ import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.TestInstance;
 
+import static com.microsoft.aad.msal4j.TestConstants.KEYVAULT_DEFAULT_SCOPE;
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.util.Collections;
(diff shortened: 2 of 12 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
```diff
@@ -0,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * List of possible reasons the tokens in an {@link IAuthenticationResult} were refreshed.
+ */
+public enum CacheRefreshReason {
```
bgavrilMS, line 9: You are missing a case here when "Claims" are specified. When this happens, MSAL has to bypass the token cache and to ESTS. 

But not for client capabilities. Client capabilities are internally still claims. But we should not bypass the cache.

bgavrilMS: MSAL.NET bundles Force Refresh + Claims https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/fe3c8eacfb013b73404b3295ec5a36b3a60b3ed8/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs#L18

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenSource.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A list of possible sources for the tokens found in an {@link IAuthenticationResult}
+ */
+public enum TokenSource {
+
+    /**
+     * A default value, likely indicates tokens could not be retrieved
+     */
+    UNKNOWN,
```
bgavrilMS, line 14: MSAL throws exception when token cannot be retried. I don't think you can get this value. Also, there is no unit test covering this path.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TokenCacheIT.java
```diff
@@ -195,6 +197,68 @@ void retrieveAccounts_ADFSOnPrem() throws Exception {
...
+        // and two that use the useSharedCache option to use the same static TokenCache
+        ConfidentialClientApplication cca_notStatic = ConfidentialClientApplication.builder(
+                        clientId, credential).
+                authority(TestConstants.MICROSOFT_AUTHORITY).
+                build();
+
+        ConfidentialClientApplication cca_sharedCache1 = ConfidentialClientApplication.builder(
+                        clientId, credential).
+                authority(TestConstants.MICROSOFT_AUTHORITY).
+                useSharedCache(true).
```
bgavrilMS, line 217: + @localden for the public API aspect. 

Currently, the Java API for specifying the cache is https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/3c00622f7af638dcae2283709a02315db2baac41/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AccountTest.java#L89

In MSAL.NET we went with a property object, so as to be able to add more cache options not just this boolean. Arguably, we never added anything else :)

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
```diff
@@ -69,6 +69,10 @@ void acquireTokenSilent_LabAuthority_TokenNotRefreshed(String environment) throw
         // Check that access and id tokens are coming from cache
         assertEquals(result.accessToken(), acquireSilentResult.accessToken());
         assertEquals(result.idToken(), acquireSilentResult.idToken());
+        assertEquals(TokenSource.IDENTITY_PROVIDER, result.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, result.metadata().cacheRefreshReason());
+        assertEquals(TokenSource.CACHE, acquireSilentResult.metadata().tokenSource());
+        assertEquals(CacheRefreshReason.NOT_APPLICABLE, acquireSilentResult.metadata().cacheRefreshReason());
```
bgavrilMS, line 75: Please add more tests. Consider adding these assertions to all integrationt tests, not just public client ones. Client_credentials is of highest priority. 

In particular:

- client_credentials + all combos of CacheRefreshReason (especially PROACTIVE_REFRESH)
- client_credentials + claims

- obo and auth code

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -60,15 +62,19 @@ AuthenticationResult execute() throws Exception {
...
                 } else if (afterRefreshOn) {
+                    this.silentRequest.requestContext().refreshReason(CacheRefreshReason.PROACTIVE_REFRESH);
                     clientApplication.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
                             CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue);
                 } else if (res.expiresOn() < currTimeStampSec) {
+                    this.silentRequest.requestContext().refreshReason(CacheRefreshReason.EXPIRED);
                     clientApplication.getServiceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
                             CacheTelemetry.REFRESH_ACCESS_TOKEN_EXPIRED.telemetryValue);
                 } else if (StringHelper.isBlank(res.accessToken())) {
+                    this.silentRequest.requestContext().refreshReason(CacheRefreshReason.NO_CACHED_ACCESS_TOKEN);
```
bgavrilMS, line 77: Is MSAL Java now sending CacheRefreshReason telemetry? Are the values different than before?

bgavrilMS (changes requested):
Insufficient tests

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TokenCacheIT.java
Avery-Dunn: Yeah, I talked with @neha-bhargava about that [CacheOptions](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/AppConfig/CacheOptions.cs) that .NET has, but I didn't add it because:
- Shared cache was the only option in there after ~2 years, and if there are still no plans to add any other options then it didn't seem like expandability is more important than simplicity
- Shared cache is a pretty big thing to enable, so even if there are more cache options in the future then having this useSharedCache API alongside a CacheOptions class wouldn't be too confusing for app developers

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: CacheRefreshReason wasn't telemetry we collected before this PR, so this is all new values. I mostly matched values that .NET had just to keep things consistent between SDKs.

bgavrilMS: Thanks Avery. This will be helpful when we write a kusto query that detects cache usage for MSAL Java customers, i.e. we will have to add "SKU=Java" and "version>1.13.10"

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
bgavrilMS: Still open comment :)

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/TokenCacheIT.java
```diff
@@ -195,6 +197,68 @@ void retrieveAccounts_ADFSOnPrem() throws Exception {
...
+        //None of the tokens retrieved using cca_notStatic should be the same as those retrieved using cca_sharedCache1 or cca_sharedCache2
+        assertNotEquals(result_notStatic1.accessToken(), result_sharedCache1.accessToken());
+        assertNotEquals(result_notStatic1.accessToken(), result_sharedCache2.accessToken());
+        assertNotEquals(result_notStatic2.accessToken(), result_sharedCache1.accessToken());
+        assertNotEquals(result_notStatic2.accessToken(), result_sharedCache2.accessToken());
+
+        //Because the confidential client flow has an internal silent call:
+        //  -result_notStatic1 and result_notStatic2 should be the same, because they both used the non-static cache from one ConfidentialClientApplication instance
+        //  -result_sharedCache1 and result_sharedCache2 should be the same, because they both used the static cache shared between two ConfidentialClientApplication instances
+        assertEquals(result_notStatic1.accessToken(), result_notStatic2.accessToken());
```
neha-bhargava, line 258: Since you added token source in the metadata, you can also check the token source as cache or IDP

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -208,6 +215,12 @@ public ConfidentialClientApplication.Builder appTokenProvider(Function<AppTokenP
             throw new NullPointerException("appTokenProvider is null") ;
         }
 
+        public ConfidentialClientApplication.Builder useSharedCache(boolean val) {
```
neha-bhargava, line 218: The user experience in MSAL.net is enable this through cache options. Are we not planning to be consistent with it?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
neha-bhargava, line 9: I could not find where this value is being sent to server side telemetry. The expected format for that is `Expected format: 5|api_id,cache_info,region_used,region_autodetection,region_outcome|platform_config` where this values is sent as cache info. Also when you send it to server side telemetry it should be a numeric value as 0,1, etc. See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ac1cb0522a011e0a6e0802a2535bf03cfd8d9fab/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs#L9

neha-bhargava (commented):
The token source and cache refresh reason is also sent to the server side telemetry. I did not see the code to send it to server side telemetry. See [HttpTelemetryManager.cs](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ac1cb0522a011e0a6e0802a2535bf03cfd8d9fab/src/client/Microsoft.Identity.Client/TelemetryCore/Http/HttpTelemetryManager.cs#L22) for reference.

PR #715, opened 2023-09-25
MSI Refactoring

Avery-Dunn:
Refactoring the codebase to better accommodate the new MSI code. The main goals of this PR:
- Make no changes to the functionality or visibility of any API that MSAL Java has already released, and do not break any of the newer MSI-related tests
- Add a new IApplicationBase interface above the existing IClientApplicationBase, to have better control over the visibility of APIs for the managed identity application vs. the existing public/confidential client applications (commit 3715e77)
- Refactor the new MSI-related classes to better follow Java best practices for naming/visibility/etc., and match the coding styles used throughout the rest of MSAL Java (commit 6835ef0)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+    @Getter
+    private Integer connectTimeoutForDefaultHttpClient;
+
+    @Accessors(fluent = true)
+    @Getter
+    private Integer readTimeoutForDefaultHttpClient;
+
+    @Accessors(fluent = true)
+    @Getter
+    private String applicationName;
```
bgavrilMS: Please leave `appname` and `appver` out, as we're not hitting ESTS. 

PS: they are not actually used :)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+
+    @Accessors(fluent = true)
+    @Getter
+    private String applicationVersion;
+
+    //The following fields are set in only some applications and/or set internally by the library. To avoid excessive
```
bgavrilMS, line 62: There are a bunch of other package internal fields that would fall under this comment. Maybe refactor this file to have all public stuff first and all internal stuff groupped.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+    @Getter(AccessLevel.PACKAGE)
+    private boolean autoDetectRegion;
+
+    @Accessors(fluent = true)
+    @Getter(AccessLevel.PACKAGE)
+    protected String azureRegion;
+
+    @Accessors(fluent = true)
+    @Getter(AccessLevel.PACKAGE)
+    private boolean instanceDiscovery;
```
neha-bhargava: I don't think we need `instanceDiscovery`, `autoDetection` and `azureRegion` in the base.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+
+    public CompletableFuture<Set<IAccount>> getAccounts() {
```
neha-bhargava: This should not be here. `getAccounts()` should not be exposed for managed identity

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AccountsSupplier.java
```diff
@@ -10,10 +10,10 @@
 
 class AccountsSupplier implements Supplier<Set<IAccount>> {
 
-    AbstractClientApplicationBase clientApplication;
+    AbstractApplicationBase clientApplication;
```
neha-bhargava: I think the `AccountSupplier` can still work with the `AbstractClientApplicationBase`. Since this is not needed for managed identity, we do not need any account specific APIs in the base class.

neha-bhargava:
We will also need to hide the API's at the request level. Right now the [ManagedIdentityParameters](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/392c93ef3828559cce69baa23cd426a6867e3608/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java#L23) implements the [IAcquireTokenParamters](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/7d3f2e0a504bde297cbb61edba07f34260bca5b1/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAcquireTokenParameters.java#L12) which have APIs WithClaims, extraQueryParams etc. Which are not relevant to managed identity. Please add a base level to IAcquireTokenParameters as well.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+    }
+
+    @Override
+    public CompletableFuture<IAuthenticationResult> acquireToken(RefreshTokenParameters parameters) {
+
+        validateNotNull("parameters", parameters);
+
+        RequestContext context = new RequestContext(
+                this,
+                PublicApi.ACQUIRE_TOKEN_BY_REFRESH_TOKEN,
```
bgavrilMS: I don't think this is part of the base. It's maybe part of the "client application". The base should not have any AcquireToken* methods.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -0,0 +1,476 @@
...
+    public CompletableFuture<IAuthenticationResult> acquireTokenSilently(SilentParameters parameters)
+            throws MalformedURLException {
+
+        validateNotNull("parameters", parameters);
+
+        RequestContext context;
+        if (parameters.account() != null) {
+            context = new RequestContext(
+                    this,
+                    PublicApi.ACQUIRE_TOKEN_SILENTLY,
```
bgavrilMS: I don't think this should be part of the base.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IApplicationBase.java
```diff
@@ -0,0 +1,68 @@
...
+
+    /**
+     * Acquires a security token from the authority using a refresh token previously received.
+     * Can be used in migration to MSAL from ADAL, and in various integration
+     * scenarios where you have a refresh token available.
+     *
+     * @param parameters {@link RefreshTokenParameters}
+     * @return A {@link CompletableFuture} object representing the {@link IAuthenticationResult} of the call.
+     */
+    CompletableFuture<IAuthenticationResult> acquireToken(RefreshTokenParameters parameters);
```
bgavrilMS: These 2 acquire token methods probably need to go the `client` interface, not here.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
```diff
@@ -9,15 +9,11 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.time.Instant;
-import java.time.format.DateTimeFormatter;
-import java.time.format.DateTimeParseException;
-import java.util.Locale;
-
 @Getter
 public class ManagedIdentityResponse {
```
bgavrilMS: Should not be public. I think @neha-bhargava already fixed this, so just a heads up when merging.

bgavrilMS (changes requested):
SILENT flow and REFREH_TOKEN flow are not part of the base.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Moved it back to AbstractClientApplicationBase in the latest commit (they were referenced in the RequestContext class which now uses AbstractApplicationBase, so I figured they could've been relevant for managed identity).

However, since they aren't actually used should we start marking them as deprecated? (in a different PR)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Moved it back to AbstractClientApplicationBase in the latest commit

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AccountsSupplier.java
Avery-Dunn: Since I also moved the getAccounts() method back into AbstractClientApplicationBase , this can definitely just use AbstractClientApplicationBase like before. Fixed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IApplicationBase.java
Avery-Dunn: I originally thought that ManagedIdentityApplication would have an explicit silent call option even though we do it implicitly, just like what we do with some ConfidentialClient flows (though that's kind of a clunky legacy thing), and so I put the silent/refresh calls into the ApplicationBase class since all the other apps would've had it.

However, after confirming this wasn't needed I just moved it and the refreshParameters method back into AbstractClientApplicationBase in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Moved back to AbstractClientApplicationBase, see this comment: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/715#discussion_r1346086456

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Moved back to AbstractClientApplicationBase, see this comment: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/715#discussion_r1346086456

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
Avery-Dunn: Some other stuff was made non-public in her PR, but this class was still public. Fixed in the latest commit, along with various other merge conflicts.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: I originally put these here to avoid excessive casting in places where the Application class was referenced via an MsalRequest object. However, since this casting would only occur in the AadInstanceDiscoveryProvider class, and it'd be confusing to have these field in AbstractApplicationBase but never used, in the latest commit I moved them back to AbstractClientApplicationBase and just did a bunch of type casting in AadInstanceDiscoveryProvider.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: In the latest commit I refactored it, so that all potentially package-private fields are below that comment. They should only ever be visible in an IDE's autocomplete if they're made public in a Application subclass, so customers should only ever see this clunky API style if they look in the source code and see this comment.

bgavrilMS (approved):

neha-bhargava (approved):

PR #729, opened 2023-10-17
Service Fabric MSI

Avery-Dunn:
Support for Service Fabric and corresponding MSI unit tests.


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
```diff
@@ -38,6 +38,8 @@ private static AbstractManagedIdentitySource createManagedIdentitySource(MsalReq
             return managedIdentitySource;
         } else if ((managedIdentitySource = CloudShellManagedIdentitySource.create(msalRequest, serviceBundle)) != null) {
             return managedIdentitySource;
+        } else if ((managedIdentitySource = ServiceFabricManagedIdentitySource.create(msalRequest, serviceBundle)) != null) {
```
bgavrilMS: In the .NET implementation, the order is different: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/ManagedIdentity/ManagedIdentityClient.cs

Looks like `ServiceFabric` needs to be chcked first. I remember from discussion with Azure SDK that order if important.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
```diff
@@ -0,0 +1,89 @@
...
+
+        managedIdentityRequest.headers = new HashMap<>();
+        managedIdentityRequest.headers.put("secret", identityHeader);
+
+        managedIdentityRequest.queryParameters = new HashMap<>();
+        managedIdentityRequest.queryParameters.put("resource", Collections.singletonList(resource));
+        managedIdentityRequest.queryParameters.put("api-version", Collections.singletonList(SERVICE_FABRIC_MSI_API_VERSION));
+
+        if (idType == ManagedIdentityIdType.CLIENT_ID) {
+            LOG.info("[Managed Identity] Adding user assigned client id to the request.");
```
bgavrilMS: nit: mention the source name in these log messages.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -149,7 +154,7 @@ void managedIdentityTest_UserAssigned_SuccessfulResponse(ManagedIdentitySourceTy
         IEnvironmentVariables environmentVariables = new EnvironmentVariablesHelper(source, endpoint);
         DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
 
-        when(httpClientMock.send(eq(expectedRequest(source, resource, id)))).thenReturn(expectedResponse(200, getSuccessfulResponse(resource)));
+        lenient().when(httpClientMock.send(eq(expectedRequest(source, resource, id)))).thenReturn(expectedResponse(200, getSuccessfulResponse(resource)));
```
bgavrilMS: Why do we use this lenient mocking instead of strict mocking? It feels like we're watering down these tests. 

If it's because of the static caching that MSI now has, which means that a token may be served from cache instead of from HTTP, then consider adding a test API to clear this cache. This is what .NET does and we can even discuss putting in the public API (maybe under some special namespace) for integration testing, if needed.

Generally tests can follow the pattern: 

1. Clear internal cache
2. Setup HTTP mocks
3. Make first token acqusitions and ensure mocks are consumed
4. Make another token acquisition and ensure tokens come from cache.

bgavrilMS (changes requested):
1. Order of sources
2. non lenient mocking if possible

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
```diff
@@ -0,0 +1,89 @@
...
+    private final ManagedIdentityIdType idType;
+    private final String userAssignedId;
+
+    @Override
+    public void createManagedIdentityRequest(String resource) {
+        managedIdentityRequest.baseEndpoint = msiEndpoint;
+        managedIdentityRequest.method = HttpMethod.GET;
+
+        managedIdentityRequest.headers = new HashMap<>();
+        managedIdentityRequest.headers.put("secret", identityHeader);
```
neha-bhargava, line 31: Nit: Make `"secret"` a constant?

neha-bhargava: Looking at the other parameters I think that should be fine.

neha-bhargava (approved):
Looks good with the unit tests and flow. Please do a manual test before merging.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
neha-bhargava: Right, move the service fabric to the top. Wondering if the unit tests are passing with this. If they are then there is something wrong.

neha-bhargava (changes requested):
Missed to catch the flow in ManagedIdentityClient.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
Avery-Dunn: Fixed in the latest commit

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
Avery-Dunn: Added in the latest commit

bgavrilMS (approved):

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
Avery-Dunn: Fixed in a commit in the PR for azure arc: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/730/commits/b81c4f83604b144961522a2fb010520fc8129b7c

neha-bhargava (approved):

PR #750, opened 2023-11-16
Fix MSI error response parsing issue

Avery-Dunn:
Fixes the issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/744

In Cloud Shell scenarios, an error response would have the form of `{"error": {...}}`, where the actual error info was in a JSON within the main JSON. This lead to our JSON parser throwing an exception when it tried to parse the sub-JSON like a primitive value.

This PR fixes that issue with a new method and small nested class. If the `"error"` field of the response is a JSON instead of a String, the JSON parsing package we use will convert it to and `ErrorField` field object that will then have the actual error info we're looking for. If the `"error"` field of the response isn't a JSON, then the existing  `error` field will be populated and that new method won't get called.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityErrorResponse.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityErrorResponse.java
@@ -15,9 +15,26 @@ public class ManagedIdentityErrorResponse {
     @JsonProperty("correlationId")
     private String correlationId;
 
+    //In some MSI scenarios such as Cloud Shell, the actual error info is in a JSON within the main JSON. To parse that second
+    // JSON layer, we need to first pass it into a subclass, parse it using the usual @JsonProperty annotation, and then retrieve the values.
+    @JsonProperty("error")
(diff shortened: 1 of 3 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityErrorResponse.java
```diff
@@ -15,9 +15,26 @@ public class ManagedIdentityErrorResponse {
...
+    }
+
     @JsonProperty("error")
     private String error;
 
     @JsonProperty("error_description")
     private String errorDescription;
+
+    @Getter
+    private static class ErrorField {
```
bgavrilMS, line 33: Is it possible to write a unit test for this?

Avery-Dunn: In the latest commit I expanded one of the unit tests to make sure both the new parsing and old parsing are both covered: the `managedIdentityTest_WrongScopes` test will now conditionally check what source type is used, and if cloud shell the HTTP mocker will return a cloud shell style error response.

Issue #751, opened 2023-11-21
Change log level from `info` to `debug` for `AcquireTokenSilentSupplier`


durimkryeziu:
In one of the projects I'm working on, we're using this library as a transitive dependency of the `azure-identity` and my logs are full of these log statements. Could this be changed to `debug` level, please?

https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/000f4c9baa0353b18cf95880c882134183564fea/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java#L112

Avery-Dunn:
Hello @durimkryeziu : I don't think there would be any issues with lowering the log level, it does seem like there could be a lot of these messages. 

We'll look into getting this fixed in an upcoming release. However, if you're seeing these MSAL Java messages because you're using an Azure SDK then our fix will only appear when they update to use the new version, and we don't have control over their release schedule.

Avery-Dunn:
As of the recent 1.14.1 release, this log should now be at the 'debug' level instead of 'info'

durimkryeziu:
Hi @Avery-Dunn, thanks for making the required change! Yes, I am aware that it doesn't mean that it will be included in the Azure SDK immediately and that you don't have control over their release schedule, so all good 👍

Issue #741, opened 2023-10-28
[Bug] HttpListener (HTTP-Dispatcher thread) stays alive for 120 seconds after client timeout occurs


samsan-nasa:
### Library version used

1.14.0

### Java version

Java 11

### Scenario

PublicClient (AcquireTokenInteractive, AcquireTokenByUsernamePassword)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

This example is using OAuth2 authentication.  I am using the  Interactive method using the default system browser to acquire a token. When I use a timeout value with the call to acquire the token with the .orTImeout() or .completeOnTimeout() methods, the HttpListener thread stays alive after the client timeout.  If you attempt a retry you will get an exception regarding the localhost port being in use because the HttpListener thread stays alive on that port.
<img width="1640" alt="ClientTimeout-HttpListenerStaysAlive" src="https://github.com/AzureAD/microsoft-authentication-library-for-java/assets/115190712/5815000f-b04b-4cd0-bbc3-0de8ad8eb9dc">

### Relevant code snippets

```csharp
// Client code adds a timeout to the acquireToken call:

PublicClientApplication publicClientApplication =
                    PublicClientApplication
                            .builder(CLIENT_ID)
                            .authority(AUTHORITY)
                            .logPii(true)
                            .build();

            InteractiveRequestParameters parameters = InteractiveRequestParameters
                    .builder(new URI("http://localhost:8888"))
                    .scopes(SCOPE)
                    .build();

            IAuthenticationResult result = publicClientApplication.acquireToken(parameters)
                .completeOnTimeout(null, 2, TimeUnit.SECONDS)
                .join();

//AcquireTokenByInteractiveFlowSupplier class hard codes a 120
//second timeout for the HttpListener:

try {
            LOG.debug("Listening for authorization result");
            long expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 120;

            while (result == null && !interactiveRequest.futureReference().get().isCancelled() &&
                    TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) < expirationTime) {

                result = authorizationResultQueue.poll(100, TimeUnit.MILLISECONDS);
            }
        } catch (Exception e) {
            throw new MsalClientException(e);
        }

// Full client code below

import com.fasterxml.jackson.databind.JsonNode;
import com.fasterxml.jackson.databind.ObjectMapper;
[...]

Issue #726, opened 2023-10-11
[Feature Request] Remove x-client-cpu


bgavrilMS:
### MSAL client type

Public, Confidential, Managed identity

### Problem Statement

See https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4167#issuecomment-1736303602 for details 

### Proposed solution

Stop sending this header.

Avery-Dunn:
This issue wasn't updated at the time, but this was resolved by hhttps://github.com/AzureAD/microsoft-authentication-library-for-java/pull/756 and released in 1.14.1 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/759

Issue #662, opened 2023-06-23
Errors is logged and thrown at the same time in AuthenticationResultSupplier resulting in handled exceptions still being logged as errors


snussbaumer:
In [AuthenticationResultSupplier, lines 98-99](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/fe7e69cc645260655a12684f1a23f3367e83fd89/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java#L98-L99) we see :  
```java
    logException(ex);
    throw new CompletionException(ex);
```

[logException](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/fe7e69cc645260655a12684f1a23f3367e83fd89/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java#L155) can in certain cases log an error with `clientApplication.log.error(logMessage, ex);`

This means that even when the exception is properly handled by the calling code (which we do) there is still a remaining error in the logs.

We have this problem and people end up investigating an issue that is not there. 

For the time being we completely disabled logging for ConfidentialClientApplication ... but it would be much preferable that the error log not be done at all (or at least that this logging could be disabled when building the ConfidentialClientApplication)

Avery-Dunn:
Hello @snussbaumer : I plan on looking into why we do that a bit more, but after a quick look I agree it doesn't make sense to both throw an exception and log it at an 'error' level.

In our next release, this will likely be changed to log those errors at a 'debug' level like the other exceptions in that `logException` method. I don't have an ETA on that, but will update this thread once we do.

Avery-Dunn:
As of the latest 1.14.1 release, this info is now logged at the 'debug' level and we don't unnecessarily print out the stack trace, so logs should raise fewer flags and be less verbose.

PR #756, opened 2023-12-05
Various fixes/improvements

Avery-Dunn:
This PR handles a number of small issues from our backlog:
- https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/726 : Removes an unnecessaries telemetry header
- https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/741 : When a future was cancelled it shut down the relevant thread, however Java 9+ introduced timeout behavior that our code didn't consider so threads would continue through a timeout. We now check for exceptions (such as TimeoutException) as well as cancellations
- https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/751 : Lowers the log level for successful cache lookup from 'info' to 'debug' to keep per-request logs more useful
- https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/662 : Reduce verbosity of logs produced right before an exception, and avoid logging at an 'error' level if an exception will be thrown anyway

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
@@ -209,7 +209,7 @@ class AcquireTokenByInteractiveFlowSupplier extends AuthenticationResultSupplier
                 expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 1;
             }
 
-            while (result == null && !interactiveRequest.futureReference().get().isCancelled()) {
+            while (result == null && !interactiveRequest.futureReference().get().isDone()) {
                 if (TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) > expirationTime) {
                     LOG.warn(String.format("Listener timed out after %S seconds, no authorization code was returned from the server during that time.", timeFromParameters));
                     break;
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AccountsSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AccountsSupplier.java
@@ -26,9 +26,10 @@ class AccountsSupplier implements Supplier<Set<IAccount>> {
                     (clientApplication.clientId());
 
         } catch (Exception ex) {
-            clientApplication.log.error(
-                    LogHelper.createMessage("Execution of " + this.getClass() + " failed.",
-                            msalRequest.headers().getHeaderCorrelationIdValue()), ex);
+            clientApplication.log.warn(
+                    LogHelper.createMessage(
+                            String.format("Execution of %s failed: %s", this.getClass(), ex.getMessage()),
+                            msalRequest.headers().getHeaderCorrelationIdValue()));
 
             throw new CompletionException(ex);
         }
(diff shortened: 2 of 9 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
```diff
@@ -209,7 +209,7 @@ private AuthorizationResult getAuthorizationResultFromHttpListener() {
                 expirationTime = TimeUnit.MILLISECONDS.toSeconds(System.currentTimeMillis()) + 1;
             }
 
-            while (result == null && !interactiveRequest.futureReference().get().isCancelled()) {
+            while (result == null && !interactiveRequest.futureReference().get().isDone()) {
```
bgavrilMS, line 212: Timeout in interactive flow is not something MSALs do and it's not something we ask app dev to do either. It can take a really long time to perform auth - e.g. I was just asked to MFA and then to change password.

But anyway, it looks like this is based on public API, so it's fine.

bgavrilMS, line 212: Could you add a comment on what "isDone" is doing here? I think it returns true when cancel() is called, or when the future completes with success?

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByInteractiveFlowSupplier.java
Avery-Dunn: For interactive flow we have a timeout parameter that would cause this loop to end after a default 120 seconds: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/ebdd4fc7299fdf31b2534c25e18da5b036b84292/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/InteractiveRequestParameters.java#L100

This ended up being kind of a weird situation. In Java 8 there wasn't really a timeout option for ending a future (there was [a timeout](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#get-long-java.util.concurrent.TimeUnit-) for waiting for the result), however in Java 9 they started introducing [methods](https://docs.oracle.com/javase%2F9%2Fdocs%2Fapi%2F%2F/java/util/concurrent/CompletableFuture.html#completeOnTimeout-T-long-java.util.concurrent.TimeUnit-) for completing the future after a timeout. Since we target Java 8 as a minimum but also want to support newer versions, this means
- Java 8 customers need that `httpPollingTimeoutInSeconds` as a timeout to end the interactive flow
- Java 9+ customers can use the newer timeout methods in Java's CompletableFuture API, and will be confused if MSAL doesn't follow those timeouts

Avery-Dunn: [isDone](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CompletableFuture.html#isDone--) returns true when a future is cancelled (like the current behavior), plus when it completes normally or with an exception. In other words, `isDone` covers any situation where we can't return the auth result, and there's no point in continuing this loop.

By only checking `isCancelled` there were scenarios where thread spun off to complete futures could continue to pointlessly run (such as when the app developers sets a timeout in the future https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/741).

Issue #690, opened 2023-08-15
Add retry logic for managed identity flow similar to the retry logic in Azure SDK.


neha-bhargava:
Add retry logic for managed identity flow for specific error codes. For more details see: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4078/files#diff-8132e330116f5d7b0add607d2c2ccdb08f2edd20354c6a9244ecaf2511f3a063

- retry on specific HTTP response codes (with default HttpClient or with HttpClient from app developer)
- add a simple API to disable default internal retry policy

Avery-Dunn:
Added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/790 and released in msal4j 1.15.0

PR #790, opened 2024-02-26
Add Managed Identity retry logic

Avery-Dunn:
Implements retry logic for MI scenarios, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/690

The main changes are:
- Refactor HttpHelper to not have static methods, and it now extends a new IHttpHelper interface
- Add new HttpHelperManagedIdentity class that extends HttpHelper and implements MI-specific HTTP behavior (currently just the retry logic)
- Refactor ServiceBundle to take IHttpHelper as a parameter, so that the now non-static methods in HttpHelper can still be accessed

Issue https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/690 also mentioned adding a way to disable this MI retry logic. Since MSAL Java doesn't have an API to disable the normal retry logic, there will be a different PR for the main library to add it


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -346,10 +346,12 @@ private static Authority createDefaultAADAuthority() {
         readTimeoutForDefaultHttpClient = builder.readTimeoutForDefaultHttpClient;
         serviceBundle = new ServiceBundle(
                 builder.executorService,
+                new TelemetryManager(telemetryConsumer, builder.onlySendFailureTelemetry),
                 builder.httpClient == null ?
                         new DefaultHttpClient(builder.proxy, builder.sslSocketFactory, builder.connectTimeoutForDefaultHttpClient, builder.readTimeoutForDefaultHttpClient) :
                         builder.httpClient,
-                new TelemetryManager(telemetryConsumer, builder.onlySendFailureTelemetry));
+                this instanceof ManagedIdentityApplication ? new HttpHelperManagedIdentity() : new HttpHelper()
```
bgavrilMS: It's generally a red flag for a base class to be aware of its implementations. https://softwareengineering.stackexchange.com/questions/219543/should-a-class-know-about-its-subclasses

Is it possible to composition for this instead?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceBundle.java
```diff
@@ -10,13 +10,15 @@ class ServiceBundle {
     private ExecutorService executorService;
     private TelemetryManager telemetryManager;
     private IHttpClient httpClient;
+    private IHttpHelper httpHelper;
     private ServerSideTelemetry serverSideTelemetry;
 
-    ServiceBundle(ExecutorService executorService, IHttpClient httpClient,
-                  TelemetryManager telemetryManager) {
+    ServiceBundle(ExecutorService executorService, TelemetryManager telemetryManager,
+                  IHttpClient httpClient, IHttpHelper httpHelper) {
```
bgavrilMS: Do you still need the httpClient in the bundle if you have the higher level construct in there? (the helper) ?

neha-bhargava:
I would recommend adding unit tests that verify retry / not based on the returned status codes.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Agreed, I just did it for simplicity and to avoid (nearly) duplicate code, but it's definitely a bad practice. 

In the latest commit I did a proper refactor, and moved the ServiceBundle creation down into the implementation classes so these sorts of 'instanceof' checks aren't needed.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceBundle.java
Avery-Dunn: It turns out the the HttpClient set in the ServiceBundle was only used in one place, to pass an instance of it into HttpHelper. I assume it was because (before this PR) all of HttpHelper's methods were used statically, so HttpHelper wasn't really intended to have a constructor where an HttpClient could be set.

However, since this PR made it all non-static, in the latest commit I added HttpClient to a new constructor in HttpHelper, and removed it from ServiceBundle's constructor.

bgavrilMS (approved):

Issue #758, opened 2023-12-11
[Bug] Service Fabric MI Auth doesn't validate server cert.


bgavrilMS:
### Library version used

1.14.4-beta

### Java version

8

### Scenario

ManagedIdentityClient - managed identity

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

Have a look at Azure Identity code here: https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/IdentityClient.java#L1032

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Solution and workarounds

Note that any solution should still offer HttpClient extensiblity. So we can't just use MSAL's internal HTTP Client stack. Needs a bit of design first. 

See equivalent .NET issue: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4462

bgavrilMS:
We have agreed with Azure SDK to not extend the HttpClient factory for this. It is acceptable for MSAL use it's own HttpClient that doesn't go through the extensibility pipeline.

Retry policies etc. are still required.

Avery-Dunn:
Fixed in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/791 and released in msal4j 1.15.0

PR #791, opened 2024-02-27
Validate Service Fabric certs

Avery-Dunn:
Adds validation for Service Fabric server certs as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/758, by bringing in the logic from Azure SDK's [IdentitySslUtil](https://github.com/Azure/azure-sdk-for-java/blob/main/sdk/identity/azure-identity/src/main/java/com/azure/identity/implementation/util/IdentitySslUtil.java) and adding a new DefaultHttpClientManagedIdentity class to handle any MI-specific HTTP client behavior


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -347,7 +356,7 @@ private static Authority createDefaultAADAuthority() {
         serviceBundle = new ServiceBundle(
                 builder.executorService,
                 builder.httpClient == null ?
-                        new DefaultHttpClient(builder.proxy, builder.sslSocketFactory, builder.connectTimeoutForDefaultHttpClient, builder.readTimeoutForDefaultHttpClient) :
+                        determineHttpClientType(builder) :
                         builder.httpClient,
```
neha-bhargava: Is builder.httpClient the http client passed in by the user? If so, this would mean that we will skip cert validation when a httpClient is passed in by the user?

Avery-Dunn: Yes, HttpClients can either be MSAL's defaults or one passed in by an app developer. Essentially there are three varieties of HTTP client that MSAL could deal with: a non-MSI default with no cert validation, an MSI default with cert validation, and an app developer's implementation of the non-MSI default interface, which they can add cert validation to.

That design was my understanding from the discussion in [.NET's thread for this](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/4462): @bgavrilMS said MSAL can "use it's own HttpClient that doesn't go through the extensibility pipeline", i.e. this new cert validation shouldn't be part of the default HttpClient that a user could extend. 

(@bgavrilMS feel free to correct me if I got that backwards, if so I can just move it to the HttpClient interface and give app developers access to it by default in their implementations)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -168,6 +168,15 @@ private AuthenticationResultSupplier getAuthenticationResultSupplier(MsalRequest
         return supplier;
     }
 
+    private IHttpClient determineHttpClientType(Builder<?> builder) {
+
+        if (this instanceof ManagedIdentityApplication) {
```
bgavrilMS: This is an inheritance warning sign. Base class should not be aware of its implementations.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
bgavrilMS: Let me try to clarify @Avery-Dunn 

- MSI flows should use either user configured HttpClient or a default HttpClient
- For ServiceFabric and only for ServiceFabric, it is acceptable to use a different HttpClient which perfoms extra validation of the SSL cert. In this case the user configured HttpClient can be ignored. 
- Ideally the design should allow for unit tests to continue to work.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Agreed, I just did it for simplicity and to avoid (nearly) duplicate code, but it's definitely a bad practice. 

In the latest commit I did a proper refactor, and moved the ServiceBundle creation down into the implementation classes so these sorts of 'instanceof' checks aren't needed.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -24,9 +24,15 @@ public class ManagedIdentityApplication extends AbstractApplicationBase implemen
 
     private ManagedIdentityApplication(Builder builder) {
         super(builder);
+        super.tokenCache = sharedTokenCache;
+        super.serviceBundle = new ServiceBundle(
+                builder.executorService,
+                builder.httpClient == null ?
```
bgavrilMS: I think this is still a problem. MSI should perform this extra check even if the app developer customized their `HTTP Client`. And it should only perform it on Service Fabric.

bgavrilMS (changes requested):
MSI should make the extra check irrespective of the user having passed in an HttpClient of their own. 
Check must be made only on SF.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
Avery-Dunn: In the latest commits I believe I've added the correct behavior:
- For all applications (public/confidential/managed identity) an `HttpClient` can be set by the app developer, or left as default
- For all flows except Service Fabric, we will use the `HttpClient` set at the app-level
- For Service Fabric, we will not use the app-level `HttpClient`, and will instead use one that has the cert validation logic

Essentially, for only Service Fabric we use a new `DefaultHttpClientManagedIdentity` that has the cert validation logic, and if needed that class can be extended to handle any other bespoke MSI logic. Additionally that class is package-private so the app developer should not need to know about it.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
```diff
@@ -22,6 +23,13 @@ class ServiceFabricManagedIdentitySource extends AbstractManagedIdentitySource {
     private final ManagedIdentityIdType idType;
     private final String userAssignedId;
 
+    //Service Fabric requires a special check for an environment variable containing a certificate thumbprint used for validating requests.
+    //
```
neha-bhargava: nit: Remove?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClientManagedIdentity.java
```diff
@@ -0,0 +1,163 @@
...
+        ALL_HOSTS_ACCEPT_HOSTNAME_VERIFIER = new HostnameVerifier() {
+            @SuppressWarnings("BadHostnameVerifier")
+            @Override
+            public boolean verify(String hostname, SSLSession session) {
+                return true;
+            }
+        };
+    }
+
+    DefaultHttpClientManagedIdentity(Proxy proxy, SSLSocketFactory sslSocketFactory, Integer connectTimeout, Integer readTimeout) {
```
neha-bhargava, line 44: Is this client created only to validate the server certificate?

neha-bhargava: Maybe you can have a default constructor here and call the parent class constructor with all nulls.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClientManagedIdentity.java
```diff
@@ -0,0 +1,163 @@
...
+import javax.net.ssl.X509TrustManager;
+import java.security.KeyManagementException;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
+import java.security.cert.Certificate;
+import java.security.cert.CertificateEncodingException;
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
+class DefaultHttpClientManagedIdentity extends DefaultHttpClient {
```
neha-bhargava, line 30: Add a comment that this client adds a server cert validation

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClientManagedIdentity.java
Avery-Dunn: Currently it is only used to add the cert validation to the service fabric flow, however I wanted to make it a more generic and extendable class to handle any other things that an MSI flow might need in the future.

Also, I just tried making a default constructor that sends nulls, but something felt wrong about sending only nulls the default option. I think it's better to have the caller be very explicit about sending all of the parameters it wants to have set, even if for this scenario `ServiceFabricManagedIdentitySource` just needs nulls.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
Avery-Dunn: It was meant as a line break, but yeah that's not really something you should do for that type of comment. Removed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClientManagedIdentity.java
Avery-Dunn: Added in the latest commit.

PR #797, opened 2024-03-11
Bind to Only Loopback Address(es) When Making Interactive Request

crimsonvspurple:
Currently, we are creating Socket using InetSocketAddress(port) which is using wildcard address, so it will attempt to bind to all available network interfaces that includes private IP like 192.168.x.x, 10.x.x.x, etc; which in turns, may/will trigger the firewall prompt.

In this commit, we explicitly bind to loopback addresses only which will not trigger firewall.


Avery-Dunn (approved):
Sorry for the delay on this, but it looks good to me. This fix will be in our next release this week.

PR #810, opened 2024-04-16
Move resource from header to query for Cloudshell

Avery-Dunn:
Per feedback from Azure SDK, the 'resource' value should be a query parameter instead of a header.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CloudShellManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CloudShellManagedIdentitySource.java
@@ -25,10 +25,12 @@ class CloudShellManagedIdentitySource extends AbstractManagedIdentitySource{
         managedIdentityRequest.headers = new HashMap<>();
         managedIdentityRequest.headers.put("ContentType", "application/x-www-form-urlencoded");
         managedIdentityRequest.headers.put("Metadata", "true");
-        managedIdentityRequest.headers.put("resource", resource);
 
         managedIdentityRequest.bodyParameters = new HashMap<>();
         managedIdentityRequest.bodyParameters.put("resource", Collections.singletonList(resource));
+
+        managedIdentityRequest.queryParameters = new HashMap<>();
+        managedIdentityRequest.queryParameters.put("resource", Collections.singletonList(resource));
     }
 
     private CloudShellManagedIdentitySource(MsalRequest msalRequest, ServiceBundle serviceBundle, URI msiEndpoint)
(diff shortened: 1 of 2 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CloudShellManagedIdentitySource.java
```diff
@@ -25,10 +25,12 @@ public void createManagedIdentityRequest(String resource) {
         managedIdentityRequest.headers = new HashMap<>();
         managedIdentityRequest.headers.put("ContentType", "application/x-www-form-urlencoded");
         managedIdentityRequest.headers.put("Metadata", "true");
-        managedIdentityRequest.headers.put("resource", resource);
 
         managedIdentityRequest.bodyParameters = new HashMap<>();
         managedIdentityRequest.bodyParameters.put("resource", Collections.singletonList(resource));
+
+        managedIdentityRequest.queryParameters = new HashMap<>();
+        managedIdentityRequest.queryParameters.put("resource", Collections.singletonList(resource));
```
rayluo, line 33: Looks like the `resource` parameter has already been put into the HTTP body (in [line 30](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/810/files#diff-796de7a4af7314c25e61c692f3f01197fc51b0bf87a647c7c424f035e512cc65R30)) and will be sent by HTTP POST (shown in [test case line 92](https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/810/files#diff-c1e78d16db8b6cb86ef4e37c4f14c7950221b0fd619b2ab501499c588b5c0833R92)). Why do we still need to put it into query string? It shouldn't need to.

P.S.: Cloud Shell's IMDS does not have explicit documentation but, for what it's worth, this was the [Cloud Shell IMDS implementation in MSAL Python](https://github.com/AzureAD/microsoft-authentication-library-for-python/blame/1.28.0/msal/cloudshell.py#L41-L47), running in production (supporting Azure Portal's one-click ssh into VM) for two years now. It is probably safe to follow that empirical implementation.

Avery-Dunn: Thanks for the Python reference, I made this PR due to an issue Azure SDK was having during testing. Essentially, they were getting an error response from the server saying that the resource was missing from the request, and I believe some other implementation they were looking at sent it as a query parameter.

Some further testing by them suggests there might be another root cause, so I've decided to just close this PR until we're more confident about what's going on.

Issue #794, opened 2024-02-27
Force a token refresh when claims are part of a silent request


Avery-Dunn:
Currently MSAL Java will return a cached token even if the request has some non-empty claims. The request could be expecting an access token with different claims than the one that was cached, leading to a new claims challenge when the token is used.

To avoid this, the silent flow behavior should match the behavior in MSAL .NET, which is to force a refresh if there are claims in the request: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Internal/Requests/Silent/CacheSilentStrategy.cs#L47

crimsonvspurple:
I was gonna request something like this too. Would love a way to skip the cache on demand.

Avery-Dunn:
Hello @crimsonvspurple : This is for a specific case where an access token may have been invalidated so the cached one might have the wrong claims, and the `acquireTokenSilently` API can/should handle it. Is that the same scenario you're working with, or is there some other scenario where you'd want the silent flow to skip the cache?

Because when using any other `acquireToken` API that isn't `acquireTokenSilently`, you should already be able to skip the cache:
 - For some confidential client applications we automatically try to return a cached token: if you're acquiring the tokens using the [on-behalf-of](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/bc6ab89c9caec96f312ed2e1d7dc0637a166ea74/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfParameters.java#L29) or [client credential](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/bc6ab89c9caec96f312ed2e1d7dc0637a166ea74/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java#L30) flows there is a 'skipCache' flag that you can use to bypass the cache.
 - For other confidential client flows and all public client flows (username/password, device code, etc.) we don't try to return a cached token ~automatically, so you can always skip the cache by simply not using the acquireTokenSilently API.~. So only `acquireTokenSilently`, `acquire token on behalf of` and `acquire token for client` need to be modified with "force refresh". Using  a cache is very important, ESTS will throttle apps that request tokens too often.

bgavrilMS:
Marking this is a bug, as it is inconsistent with the other MSALs. Thanks for discovering this. 

[...]

PR #811, opened 2024-04-16
Refresh tokens when request contains claims

Avery-Dunn:
The first commit in this PR refactors the logic for determining when a refresh should be performed, since it had gotten complex as different scenarios and edge cases were introduced over the years. That refactor should not affect the existing refresh behavior. 

The second commit adds the existence of the claims parameter as reason to refresh, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/794. With this change app developers will no longer need to explicitly use the forceRefresh() API when dealing with tokens that have claims/client capabilities/etc.

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
@@ -193,7 +193,7 @@ class AcquireTokenSilentIT {
     }
 
     @Test
-    public void acquireTokenSilent_ConfidentialClient_acquireTokenSilentDifferentScopeThrowsException()
+    void acquireTokenSilent_ConfidentialClient_acquireTokenSilentDifferentScopeThrowsException()
             throws Exception {
         cfg = new Config(AzureEnvironment.AZURE);
 
@@ -342,6 +342,48 @@ class AcquireTokenSilentIT {
 
         assertResultNotNull(silentResult);
         assertEquals(result.accessToken(), silentResult.accessToken());
+    }
+
+    @Test
+    public void acquireTokenSilent_ClaimsForceRefresh() throws Exception {
+        cfg = new Config(AzureEnvironment.AZURE);
+        User user = labUserProvider.getDefaultUser(AzureEnvironment.AZURE);
+
+        Set<String> scopes = new HashSet<>();
+        PublicClientApplication pca = PublicClientApplication.builder(
+                        user.getAppId()).
+                authority(cfg.organizationsAuthority()).
+                build();
+
+        IAuthenticationResult result = pca.acquireToken(UserNamePasswordParameters.
+                        builder(scopes,
+                                user.getUpn(),
+                                user.getPassword().toCharArray())
+                        .build())
+                .get();
+
+        assertResultNotNull(result);
+
+        IAuthenticationResult silentResultWithoutClaims = pca.acquireTokenSilently(SilentParameters.
+                        builder(scopes, result.account())
+                        .build())
+                .get();
+
+        assertResultNotNull(silentResultWithoutClaims);
+        assertEquals(result.accessToken(), silentResultWithoutClaims.accessToken());
+
+        //If claims are added to a silent request, it should trigger the refresh flow and return a new token
+        ClaimsRequest cr = new ClaimsRequest();
+        cr.requestClaimInAccessToken("email", null);
+
+        IAuthenticationResult silentResultWithClaims = pca.acquireTokenSilently(SilentParameters.
+                        builder(scopes, result.account())
+                        .claims(cr)
+                        .build())
+                .get();
+
+        assertResultNotNull(silentResultWithClaims);
+        assertNotEquals(result.accessToken(), silentResultWithClaims.accessToken());
     }
 
     private IConfidentialClientApplication getConfidentialClientApplications() throws Exception {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
@@ -131,12 +131,13 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
     //Handles any logic to determine if a token should be refreshed, based on the request parameters and the status of cached tokens
     private boolean shouldRefresh(SilentParameters parameters, AuthenticationResult cachedResult) {
 
-        //If there is a refresh token but no access token, we should use the refresh token to get the access token
-        if (StringHelper.isBlank(cachedResult.accessToken()) && !StringHelper.isBlank(cachedResult.refreshToken())) {
+        //If forceRefresh is true, no reason to check any other option
+        if (parameters.forceRefresh()) {
             return true;
         }
 
         //If the request contains claims then the token should be refreshed, to ensure that the returned token has the correct claims
+        //  Note: these are the types of claims found in (for example) a claims challenge, and do not include client capabilities
         if (parameters.claims() != null) {
             return true;
         }
@@ -149,7 +150,11 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
             return true;
         }
 
-        //Finally, simply return the value of the request's forceRefresh parameter
-        return parameters.forceRefresh();
+        //If there is a refresh token but no access token, we should use the refresh token to get the access token
+        if (StringHelper.isBlank(cachedResult.accessToken()) && !StringHelper.isBlank(cachedResult.refreshToken())) {
+            return true;
+        }
+
+        return false;
     }
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
@@ -13,8 +13,7 @@ import java.util.Date;
 class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
 
     private SilentRequest silentRequest;
-    private boolean shouldRefresh;
-    private boolean afterRefreshOn;
+    protected static final int ACCESS_TOKEN_EXPIRE_BUFFER_IN_SEC = 5 * 60;
 
     AcquireTokenSilentSupplier(AbstractApplicationBase clientApplication, SilentRequest silentRequest) {
         super(clientApplication, silentRequest);
@@ -24,6 +23,7 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
 
     @Override
     AuthenticationResult execute() throws Exception {
+        boolean shouldRefresh;
         Authority requestAuthority = silentRequest.requestAuthority();
         if (requestAuthority.authorityType != AuthorityType.B2C) {
             requestAuthority =
@@ -57,7 +57,7 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
 
             shouldRefresh = shouldRefresh(silentRequest.parameters(), res);
 
-            if (shouldRefresh || afterRefreshOn) {
+            if (shouldRefresh || clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo() == CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue) {
                 if (!StringHelper.isBlank(res.refreshToken())) {
                     //There are certain scenarios where the cached authority may differ from the client app's authority,
                     // such as when a request is instance aware. Unless overridden by SilentParameters.authorityUrl, the
@@ -100,11 +100,11 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
             return refreshedResult;
         } catch (MsalServiceException ex) {
             //If the token refresh attempt threw a MsalServiceException but the refresh attempt was done
-            // only because of refreshOn, then simply return the existing cached token
-            if (!afterRefreshOn && (silentRequest.parameters().forceRefresh() || StringHelper.isBlank(cachedResult.accessToken()))) {
-                throw ex;
+            // only because of refreshOn, then simply return the existing cached token rather than throw an exception
+            if (clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo() == CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue) {
+                return cachedResult;
             }
-            return cachedResult;
+            throw ex;
         }
     }
 
@@ -114,6 +114,7 @@ class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
         //If forceRefresh is true, no reason to check any other option
         if (parameters.forceRefresh()) {
             setCacheTelemetry(CacheTelemetry.REFRESH_FORCE_REFRESH.telemetryValue);
+            log.debug("Refreshing access token because forceRefresh parameter is true.");
             return true;
         }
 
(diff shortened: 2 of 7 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -120,4 +82,74 @@ AuthenticationResult execute() throws Exception {
...
+            clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
+                    CacheTelemetry.REFRESH_NO_ACCESS_TOKEN.telemetryValue);
+        }
+    }
+
+    //Handles any logic to determine if a token should be refreshed, based on the request parameters and the status of cached tokens
+    private boolean shouldRefresh(SilentParameters parameters, AuthenticationResult cachedResult) {
+
+        //If there is a refresh token but no access token, we should use the refresh token to get the access token
+        if (StringHelper.isBlank(cachedResult.accessToken()) && !StringHelper.isBlank(cachedResult.refreshToken())) {
```
bgavrilMS: highly unlikely, I recommend moving to the bottom as optimization.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -120,4 +82,74 @@ AuthenticationResult execute() throws Exception {
...
+            return true;
+        }
+
+        //If the request contains claims then the token should be refreshed, to ensure that the returned token has the correct claims
+        if (parameters.claims() != null) {
```
bgavrilMS, line 123: I assume `claims()` does not contain the client capabilities? Maybe add to the comment above.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -120,4 +82,74 @@ AuthenticationResult execute() throws Exception {
...
+            //If the token refresh attempt threw a MsalServiceException but the refresh attempt was done
+            // only because of refreshOn, then simply return the existing cached token
+            if (!afterRefreshOn && (silentRequest.parameters().forceRefresh() || StringHelper.isBlank(cachedResult.accessToken()))) {
+                throw ex;
+            }
+            return cachedResult;
+        }
+    }
+
+    private void setRefreshTelemetry(AuthenticationResult cachedResult) {
```
bgavrilMS: This duplicates some of logic in `shouldRefresh`. Can you make it more [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself), for example by having `shouldRefresh` return the enum value instead of boolean?

Also, it doesn't take into account claims.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -120,4 +82,74 @@ AuthenticationResult execute() throws Exception {
...
+        if (cachedResult.refreshOn() != null && cachedResult.refreshOn() > 0 &&
+                cachedResult.refreshOn() < currTimeStampSec && cachedResult.expiresOn() >= currTimeStampSec){
+            afterRefreshOn = true;
+            return true;
+        }
+
+        //Finally, simply return the value of the request's forceRefresh parameter
+        return parameters.forceRefresh();
+    }
 }
```
bgavrilMS, line 160: Missing unit tests.

bgavrilMS (changes requested):
Missing unit tests is primary concern.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Yeah, the only reasonable scenario I can see this one happening is if someone configures a cache where they wiped all of the access tokens but left the refresh tokens.

In the latest commit I moved the check for forceRefresh to the top (since it can simply override any other option if true) and this check to the bottom.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Yes, the claims parameter and clientCapabilities parameter only get merged right before they're used in the request, so this would not have client capabilities.

In the latest commit I added an extra comment to clarify that client capabilities should not trigger this refresh.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: In the latest commit I moved this code into the shouldRefresh() method so that it also sets the telemetry when determining whether or not to refresh (and claims are treated just like a force refresh)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Added test in the latest commit specifically for claims being a reason to refresh, other refresh reasons are already covered by existing tests.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -13,6 +13,8 @@
 class AcquireTokenSilentSupplier extends AuthenticationResultSupplier {
 
     private SilentRequest silentRequest;
+    private boolean shouldRefresh;
```
bgavrilMS: This seems to be a local variable in the "execute" method, no need to have this as obj state.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -13,6 +13,8 @@
...
+    private boolean afterRefreshOn;
```
bgavrilMS: I think you can avoid adding this boolean to the object state by checking the telemetry flag is set to `REFRESH_REFRESH_IN`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -53,29 +55,9 @@ AuthenticationResult execute() throws Exception {
...
-                    clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
-                            CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue);
-                } else if (res.expiresOn() < currTimeStampSec) {
-                    clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
-                            CacheTelemetry.REFRESH_ACCESS_TOKEN_EXPIRED.telemetryValue);
-                } else if (StringHelper.isBlank(res.accessToken())) {
-                    clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo(
-                            CacheTelemetry.REFRESH_NO_ACCESS_TOKEN.telemetryValue);
-                }
+            shouldRefresh = shouldRefresh(silentRequest.parameters(), res);
```
bgavrilMS, line 58: The reason for the cache refresh should be captured in logging satements.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -53,29 +55,9 @@ AuthenticationResult execute() throws Exception {
                 clientApplication.serviceBundle().getServerSideTelemetry().incrementSilentSuccessfulCount();
             }
 
-            //Determine if the current token needs to be refreshed according to the refresh_in value
```
bgavrilMS, line 56: nit: control flow via exception should be avoided (see line 48). 

https://stackoverflow.com/questions/729379/why-not-use-exceptions-as-regular-flow-of-control

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -84,29 +66,7 @@ AuthenticationResult execute() throws Exception {
                         requestAuthority = Authority.createAuthority(new URL(requestAuthority.authority().replace(requestAuthority.host(),
                                 res.account().environment())));
                     }
```
bgavrilMS, line 68: nit: as per comment above, control flow using exception is not great.

This is for line 76 (for some reason GH refuses to let me comment on that line...)

https://stackoverflow.com/questions/729379/why-not-use-exceptions-as-regular-flow-of-control

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -120,4 +80,75 @@ AuthenticationResult execute() throws Exception {
...
+        long currTimeStampSec = new Date().getTime() / 1000;
+
+        if (cachedResult.expiresOn() < currTimeStampSec) {
+            setCacheTelemetry(CacheTelemetry.REFRESH_ACCESS_TOKEN_EXPIRED.telemetryValue);
+            return true;
+        }
+
+        //Certain long-lived tokens will have a 'refresh on' time that indicates a refresh should be attempted long before the token would expire
+        if (cachedResult.refreshOn() != null && cachedResult.refreshOn() > 0 &&
+                cachedResult.refreshOn() < currTimeStampSec && cachedResult.expiresOn() >= currTimeStampSec){
```
bgavrilMS: `cachedResult.expiresOn() >= currTimeStampSec` - not sure this is correct. MSALs use a 5 min buffer for this comparison.  Does MSAL Java not do this elsewhere? If not, please create a bug / consistency item to be fixed later (p2 at most). 5 min buffer is needed to account for time discrepancies.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
```diff
@@ -344,6 +344,48 @@ void acquireTokenSilent_emptyScopeSet(String environment) throws Exception {
         assertEquals(result.accessToken(), silentResult.accessToken());
     }
 
+    @Test
+    public void acquireTokenSilent_ClaimsForceRefresh() throws Exception {
+        cfg = new Config(AzureEnvironment.AZURE);
+        User user = labUserProvider.getDefaultUser(AzureEnvironment.AZURE);
+
+        Set<String> scopes = new HashSet<>();
+        PublicClientApplication pca = PublicClientApplication.builder(
```
bgavrilMS, line 353: I think unit tests should cover this not integration tests. Integration tests are expensive. There is no integration point that is worth testing unless you pass in some well-known claims and can assert the result. 

Unit tests should cover: 

- cca.AcquireTokenSilent 
- cca.AcquireTokenForClient
- pca.AcquireTokenSilent

bgavrilMS (changes requested):
More test coverage (unit test) needed.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: This logic was copied from where we set telemetry, which appears to have been missing the 5 minute buffer. However, [the method in TokenCache](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/fb028675c22d406ca6fdc9bcda6fa210d47fa806/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java#L483) that actually determines what tokens were pulled from the cache does correctly have the buffer.

In the latest commit I moved the logic for determining if a token is expired into AcquireTokenSilentSupplier to ensure the refresh logic is all in one place and telemetry is properly set.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Added a logging statement for each refresh reason (at the same logging level as the 'returned token from cache' message)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Because the telemetry flag is stored as an int several layers into the object passed into this class, I'd have to replace the use of `afterRefreshOn` with the full `clientApplication.serviceBundle().getServerSideTelemetry().getCurrentRequest().cacheInfo() == CacheTelemetry.REFRESH_REFRESH_IN.telemetryValue`

This makes one of the if statements where it's used messier and the other one cleaner, and since that evens out well enough I just got rid of `afterRefreshOn` in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Fixed in latest commit.

bgavrilMS:
Any updates on this @Avery-Dunn ?

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenSilentIT.java
Avery-Dunn: In the latest commit I added extra tests using some dummy cache data, to show that a token is returned when using the AcquireTokenSilent API without claims and not returned when claims are added to the request.

I did not add a specific test for the implicit silent call in the 'cca.AcquireTokenForClient' scenario, because internally it reaches the exact same code as the explicit AcquireTokenSilent call and that behavior is covered by other tests.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -45,7 +56,71 @@ void confidentialAppAcquireTokenSilently_emptyCache_MsalClientException() throws
...
+        assertEquals("token", result.accessToken());
+
+        ClaimsRequest cr = new ClaimsRequest();
+        cr.requestClaimInAccessToken("something", null);
+
+        parameters = SilentParameters.builder(Collections.singleton("scopes"), basicAccount).claims(cr).build();
+        CompletableFuture<IAuthenticationResult> future = application.acquireTokenSilently(parameters);
+
+        //Confirm cached dummy token ignored when claims are part of request
+        ExecutionException ex = assertThrows(ExecutionException.class, future::get);
```
bgavrilMS, line 88: These asserts are strange. Would it make sense to just merge the improvment where you expose "TokenSource" and assert on that?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -45,7 +56,71 @@ void confidentialAppAcquireTokenSilently_emptyCache_MsalClientException() throws
...
+
+        SilentParameters parameters = SilentParameters.builder(Collections.singleton("scopes"), basicAccount).build();
+
+        IAuthenticationResult result = application.acquireTokenSilently(parameters).get();
+
+        //Confirm cached dummy token returned from silent request
+        assertNotNull(result);
+        assertEquals("token", result.accessToken());
+
+        ClaimsRequest cr = new ClaimsRequest();
```
bgavrilMS, line 81: This is not good API experience ( I know it exists already).

 Is there an overload where the claims are injected as a string? After all, this is what the app developers will use.|
 
 Later edit: can we pls deprecate Claims|Request and associated APIs? And just keep `claims(string)` ?"

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
```diff
@@ -30,30 +31,25 @@ void testClaimsRequest_Format() {
...
         PublicClientApplication pca = PublicClientApplication.builder(
                 "client_id").
-                clientCapabilities(new HashSet<>(Collections.singletonList("llt"))).
+                clientCapabilities(capabilities).
                 build();
 
         InteractiveRequestParameters parameters = InteractiveRequestParameters.builder(new URI("http://localhost:8080"))
-                .claimsChallenge("{\"id_token\":{\"auth_time\":{\"essential\":true}},\"access_token\":{\"auth_time\":{\"essential\":true},\"xms_cc\":{\"values\":[\"abc\"]}}}")
+                .claimsChallenge(TestConfiguration.CLAIMS_CHALLENGE)
                 .claims(cr)
```
bgavrilMS, line 53: Why both claims and claims channelge?

Can we deprecate this `claims` and ClaimsRequest object model?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
```diff
@@ -30,30 +31,25 @@ void testClaimsRequest_Format() {
...
-        cr.requestClaimInAccessToken("email", null);
-        cr.requestClaimInIdToken("acr", new RequestedClaimAdditionalInfo(false, null, values));
-        cr.requestClaimInIdToken("sub", new RequestedClaimAdditionalInfo(true, "248289761001", null));
-        cr.requestClaimInIdToken("auth_time", new RequestedClaimAdditionalInfo(false, null, null));
+        cr.requestClaimInAccessToken("nbf", new RequestedClaimAdditionalInfo(true, "1701477303", null));
+
+        Set<String> capabilities = new HashSet<>();
+        capabilities.add("cp1");
 
         PublicClientApplication pca = PublicClientApplication.builder(
```
bgavrilMS, line 46: Prefer to write tests for confidential client. Public client should go through broker.

bgavrilMS (changes requested):
Unsure about claims(string) and claimsrequest(CR) - propose to deprecate strongly typed model and it is inconsistent with other MSALs.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
Avery-Dunn: For simplicity I'm not deprecating in this PR/release, but I created this task to get it deprecated in the next one: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/855

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
Avery-Dunn: For simplicity I'm not deprecating in this PR/release, but I created this task to get it deprecated in the next one: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/855

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClaimsTest.java
Avery-Dunn: This was an existing test that was adjusted to show claims behave correctly, but going forward I'll avoid adjusting old public client tests and continue to avoid adding new ones.

PR #814, opened 2024-05-02
Change Cloudshell request method from POST to GET

Avery-Dunn:
Testing by the Azure SDK team has shown an issue when making the POST request for Cloudshell, where the server responds with a error message saying the resource is missing from the request. This only occurs when using Netty, a common networking framework for Java, and POST requests work fine in other MSALs.

However, further testing has shown it works fine in Netty and other common frameworks if it is a GET request with resource as a query parameter, and this PR makes that change as a workaround. This change should not affect any existing working customers, and we should be able to swap back to a POST request in the future if we need to.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
@@ -49,12 +49,11 @@ abstract class AbstractManagedIdentitySource {
         IHttpResponse response;
 
         try {
-
-            HttpRequest httpRequest = managedIdentityRequest.method.equals(HttpMethod.GET) ?
-                    new HttpRequest(HttpMethod.GET,
+            HttpRequest httpRequest = StringHelper.isNullOrBlank(managedIdentityRequest.getBodyAsString()) ?
+                    new HttpRequest(managedIdentityRequest.method,
                             managedIdentityRequest.computeURI().toString(),
                             managedIdentityRequest.headers) :
-                    new HttpRequest(HttpMethod.POST,
+                    new HttpRequest(managedIdentityRequest.method,
                             managedIdentityRequest.computeURI().toString(),
                             managedIdentityRequest.headers,
                             managedIdentityRequest.getBodyAsString());
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
@@ -49,14 +49,9 @@ abstract class AbstractManagedIdentitySource {
         IHttpResponse response;
 
         try {
-            HttpRequest httpRequest = StringHelper.isNullOrBlank(managedIdentityRequest.getBodyAsString()) ?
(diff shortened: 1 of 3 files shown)


rayluo:
> when making the POST request for Cloudshell, where the server responds with a error message saying the resource is missing from the request. This only occurs when using Netty, a common networking framework for Java, and POST requests work fine in other MSALs.

That sounds so weird. HTTP stuffs are supposed to be language and platform-agnostic.

Can we have a short(est) code snippet written with Netty to repro that? That way, we can then have either Cloud Shell folks or Netty folks to troubleshoot to see whether it is a bug on their sides.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CloudShellManagedIdentitySource.java
```diff
@@ -20,15 +20,17 @@ class CloudShellManagedIdentitySource extends AbstractManagedIdentitySource{
...
-        managedIdentityRequest.method = HttpMethod.POST;
+        managedIdentityRequest.method = HttpMethod.GET;
 
         managedIdentityRequest.headers = new HashMap<>();
         managedIdentityRequest.headers.put("ContentType", "application/x-www-form-urlencoded");
         managedIdentityRequest.headers.put("Metadata", "true");
-        managedIdentityRequest.headers.put("resource", resource);
 
         managedIdentityRequest.bodyParameters = new HashMap<>();
         managedIdentityRequest.bodyParameters.put("resource", Collections.singletonList(resource));
```
neha-bhargava: Is this still needed with GET? I don't think so

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
```diff
@@ -49,12 +49,11 @@ public ManagedIdentityResponse getManagedIdentityResponse(
         IHttpResponse response;
 
         try {
-
-            HttpRequest httpRequest = managedIdentityRequest.method.equals(HttpMethod.GET) ?
-                    new HttpRequest(HttpMethod.GET,
+            HttpRequest httpRequest = StringHelper.isNullOrBlank(managedIdentityRequest.getBodyAsString()) ?
```
neha-bhargava: The cloud shell was the only post method. If you change this to GET I don't think we need this check anymore.

bgavrilMS (approved):
Approved with suggestions from Neha

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CloudShellManagedIdentitySource.java
Avery-Dunn: Removed in the latest commit, I was initially worried that it was still needed in the body in some scenarios but the query param seems to cover everything.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractManagedIdentitySource.java
Avery-Dunn: Yep, it's no longer needed. Removed in the latest commit.

neha-bhargava (approved):

Issue #820, opened 2024-05-22
[Bug] DefaultHttpClient ignores supplied SSLSocketFactory


stefanlourens:
### Library version used

1.15.0

### Java version

21.0.2

### Scenario

ConfidentialClient - web site (AcquireTokenByAuthCode)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

Supplying a SSLSocketFactory to the  ConfidentialClientApplication builder has no effect. 

I traced this down to the DefaultHttpClient's handling of SSL connections: 

It currently checks if the connection is an instance of ```HttpURLConnection```, but since ```HttpsURLConnection``` extends ```HttpURLConnection``` it's always true and the else is never executed. 

```java
if (connection instanceof HttpURLConnection) {
    return (HttpURLConnection) connection;
} else {
    HttpsURLConnection httpsConnection = (HttpsURLConnection) connection;

    if (sslSocketFactory != null) {
        httpsConnection.setSSLSocketFactory(sslSocketFactory);
    }

    return httpsConnection;
}
```
https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java#L93-L103

I suggest changing this to:

```java
if (connection instanceof HttpsURLConnection) {
    HttpsURLConnection httpsConnection = (HttpsURLConnection) connection;

    if (sslSocketFactory != null) {
        httpsConnection.setSSLSocketFactory(sslSocketFactory);
    }

    return httpsConnection;
} else {
    return (HttpURLConnection) connection;
}
```

### Expected behavior

The supplied SSLSocketFactory should be set on the HttpsUrlConnection ```httpsConnection.setSSLSocketFactory(sslSocketFactory);``` or setting the default ssl handing externally.

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

### Solution and workarounds

Currently the only workaround is to supply your own http client, with the logic fixed.

Avery-Dunn:
Hello @stefanlourens : Thanks for bringing this to our attention! I was able to reproduce the issue, and this is definitely unintended behavior: if the URL is 'https' then the `instance of Http...` and `instance of Https...` checks are both true and the else block will never be reached.

I don't have an ETA but the fix will be in either our next release or the one after, and I'll update this thread once it's done.

[...]

PR #821, opened 2024-05-24
Fix for supplied SSLSocketFactory being ignored

stefanlourens:
It was checking if the connection is an instance of ```HttpURLConnection```, but since ```HttpsURLConnection``` extends ```HttpURLConnection``` it's always true and the else is never executed. 

https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/820


bgavrilMS (approved):

Avery-Dunn (approved):

Avery-Dunn:
Thanks for making this! We have some specific fixes planned for our next release, so if this does not get merged in for that release then it will be part of a hotfix release immediately after.

stefanlourens:
@bgavrilMS Do I need to do something to get the build to trigger to update the status?

neha-bhargava (approved):

bgavrilMS:
@Avery-Dunn - any issues with merging this? I don't think it's unit testable.

deepak-dhull-02 (commented):
good change

Issue #822, opened 2024-05-28
[Feature Request]  Expose refreshOn in IAuthenticationResult


g2vinay:
### MSAL client type

Public, Confidential, Managed identity

### Problem Statement

The refreshOn property of the token isn't returned as part of IAuthenticationResult.
It prevents users from accessing and honoring this property.

This feature is required for :
1) Cross-language consistency with .NET and other Azure languages which support this capability already. 
2) Enterprise Azure customers need this capability exposed so that token cache refreshes can happen more aggressively and reduce the chance of performance issues with near-expired or expired tokens

Today, Azure SDK has its Token Cache implementation which defaults to a refresh offset of 5 minutes.
This cache caters to all implementations of TokenCredential which may or may not be using Msal.
So, today the refresh offset of 5 minutes overrides the Msal's refresh_on , as the refresh_on is not available at Azure SDK cache layer.
We need the refresh_on information to correctly invoke the TokenCredential at its desired refresh_on time.

Further, the default refresh offset of 5 minutes in Azure SDK token cache is creating reliability issues for our enterprise customers and the ask is for Azure SDK to use the refresh on info in their cache implementation to invoke the token credential at the given refresh on time.

### Proposed solution

Expose the refreshOn property of the token similar to the expiry time.

bgavrilMS:
Hi @g2vinay - not understanding how "refreshOn" is useful on AuthenticationResult object? We can of course add it in case ppl want to log this value for their own telemetry purposes. But does it have any functional value? 

RefreshOn is for MSAL to perform background token refresh.

g2vinay:
> Hi @g2vinay - not understanding how "refreshOn" is useful on AuthenticationResult object? We can of course add it in case ppl want to log this value for their own telemetry purposes. But does it have any functional value?
> 
> RefreshOn is for MSAL to perform background token refresh.

Yeah we have a cache layer in azure-core, that defaults to refresh offset of 5 minutes. 
Having this info available will allow the cache in azure-core to honor it.

g2vinay:
Based on the discussion with @Avery-Dunn 
This feature will align with already exposed property in Msal.NET.
This feature is expected to be included in next release of msal4j.

To further highlight,
Enterprise Azure customers need this capability exposed so that token cache refreshes can happen more aggressively and reduce the chance of performance issues with near-expired or expired tokens

bgavrilMS:
[...]

PR #829, opened 2024-06-24
Expose refreshon parameter in authentication result

Avery-Dunn:
Exposes the refresh on parameter in the AuthenticationResultMetadata field of an AuthenticationResult, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/822

Also replaces the (package-private) AuthenticationResultMetadata constructors with lombok's builder annotation to better handle setting the new refreshOn field and any others that may get added.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
@@ -155,7 +155,10 @@ class TokenRequestExecutor {
                     refreshOn(response.getRefreshIn() > 0 ? currTimestampSec + response.getRefreshIn() : 0).
                     accountCacheEntity(accountCacheEntity).
                     scopes(response.getScope()).
-                    metadata(new AuthenticationResultMetadata(TokenSource.IDENTITY_PROVIDER)).
+                    metadata(AuthenticationResultMetadata.builder()
+                            .tokenSource(TokenSource.IDENTITY_PROVIDER)
+                            .refreshOn(response.getRefreshIn() > 0 ? currTimestampSec + response.getRefreshIn() : 0)
+                            .build()).
                     build();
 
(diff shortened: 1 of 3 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -155,7 +155,10 @@ private AuthenticationResult createAuthenticationResultFromOauthHttpResponse(
                     refreshOn(response.getRefreshIn() > 0 ? currTimestampSec + response.getRefreshIn() : 0).
```
neha-bhargava, line 155: Is this already exposed in `AuthenticationResult`? If it is why do we need it in the metadata as well? If this makes more sense in metadata, maybe remove it from here?

neha-bhargava (approved):
Approved with comments.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
Avery-Dunn: `refreshOn` is a field in `AuthenticationResult` but not the `IAuthenticationResult` interface, and the return type of our `acquireToken` methods is `IAuthenticationResult`. `AuthenticationResult` itself is a package-private, so other projects can only see the fields in `IAuthenticationResult`

For various historical design reasons `IAuthenticationResult` has most of the fields listed out individually but that's not very expandable: adding a new field to a public interface is considered a breaking change unless we also add a (potentially meaningless) default value, but that's considered a bad practice. 

That's the reason the `AuthenticationResultMetadata` class was introduced: adding new fields to that is not a breaking change, and it can be a bucket for more niche things like `refreshOn` that most users will not need to worry about.

bgavrilMS (changes requested):
No unit tests

Issue #836, opened 2024-07-11
[Bug] RefreshOn info isn't wired in AcquireTokenByTokenProvider supplier.


g2vinay:
### Library version used

1.16.0

### Java version

17

### Scenario

ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

The app is in production, and I have upgraded to a new version of MSAL

### Issue description and reproduction steps

The refreshOn information isn't wired in AuthenticationResult -> Metadata -> RefreshOn, it comes back as null. 

It is missing the wiring [here](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/8d67dca83c56f90aea49e57d06b5056d0afbffef/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java#L89)

This issue is a blocker for Cosmos team.

### Relevant code snippets

```csharp
Steps to reproduce:


        ConfidentialClientApplication confidentialClientApplication =  getManagedIdentityConfidentialClient();

        TokenRequestContext tokenRequestContext = new TokenRequestContext().addScopes("https://vault.azure.net/.default");
        ClientCredentialParameters.ClientCredentialParametersBuilder builder =
                ClientCredentialParameters.builder(new HashSet<>(tokenRequestContext.getScopes()))
                        .tenant("organizations");

        IAuthenticationResult authenticationResult = confidentialClientApplication.acquireToken(builder.build()).get();

        System.out.println(authenticationResult.metadata().refreshOn());


```java
    static ConfidentialClientApplication getManagedIdentityConfidentialClient() {
        String authorityUrl = AzureAuthorityHosts.AZURE_PUBLIC_CLOUD + "organizations";

        // Temporarily pass in Dummy Client secret and Client ID. until MSal removes its requirements.
        IClientCredential credential = ClientCredentialFactory
                .createFromSecret("dummy-secret");
        ConfidentialClientApplication.Builder applicationBuilder =
                ConfidentialClientApplication.builder( "SYSTEM-ASSIGNED-MANAGED-IDENTITY", credential);

        applicationBuilder
                .instanceDiscovery(false)
                .validateAuthority(false)
                .logPii(true);

        try {
            applicationBuilder = applicationBuilder.authority(authorityUrl);
        } catch (MalformedURLException e) {
            throw (new IllegalStateException(e));
        }

[...]

PR #838, opened 2024-07-12
Add missing refreshOn metadata

Avery-Dunn:
A new field of auth result metadata was recently added to AuthenticationResults in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/829

However, as described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/836, there are some other code paths where the AuthenticationResult is built and was missing the metadata object in the builder.


bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAppProviderSupplier.java
```diff
@@ -87,6 +87,9 @@ public AuthenticationResult fetchTokenUsingAppTokenProvider(AppTokenProviderPara
                 .idToken(null)
                 .expiresOn(tokenProviderResult.getExpiresInSeconds())
                 .refreshOn(tokenProviderResult.getRefreshInSeconds())
+                .metadata(AuthenticationResultMetadata.builder()
```
bgavrilMS, line 90: THis is ok, but how do you know it works? I.e. is there no testing that can be done here?

bgavrilMS (changes requested):
No unit tests.

Issue #833, opened 2024-06-27
[BUG] AcquireTokenSilentSupplier failed: Token not found in the cache


joshfree:
Ported issue filed by @krysiuda at https://github.com/Azure/azure-sdk-for-java/issues/40654 to https://github.com/AzureAD/microsoft-authentication-library-for-java

**Describe the bug**
Between msal4j versions 1.14.0 and 1.15.0 logging for AuthenticationErrorCode.CACHE_MISS has changed. I believe the change was accidental, resulting in additional (and potentially misleading) WARN messages being log.

***Exception or Stack Trace***
Our production system is logging "Execution of class com.microsoft.aad.msal4j.AcquireTokenSilentSupplier failed: Token not found in the cache" on "com.microsoft.aad.msal4j.ConfidentialClientApplication" logger (severity: WARN) every hour.

**Expected behavior**
Since cache miss simply triggers retrieval of a new token, I would expect this event (cache miss) to be logged with low severity. Preferably DEBUG level.

**Setup (please complete the following information):**
 - OS: linux
 - IDE: IntelliJ
 - Library/Libraries: msal4j-1.15.0-sources.jar
 - Java version: openjdk-21
 - App Server/Environment: N/A
 - Frameworks: N/A 

**Additional context**
in msal4j-1.14.0 AuthenticationResultSupplier there was:
```
            if (exception.errorCode() != null && exception.errorCode().equalsIgnoreCase(AuthenticationErrorCode.CACHE_MISS)) {
                clientApplication.log.debug(logMessage, ex);
                return;
            }
            ...
            clientApplication.log.error(logMessage, ex);
```
logging CACHE_MISS on DEBUG level instead of ERROR.

in msal4j-1.15.0, there is only:
```
                clientApplication.log.warn(
                        LogHelper.createMessage(
                                String.format("Execution of %s failed: %s", this.getClass(), ex.getMessage()),
                                msalRequest.headers().getHeaderCorrelationIdValue()));
```
logging cache misses always on WARN level.

Avery-Dunn:
[...]

PR #844, opened 2024-07-23
Reduce logging level of CACHE_MISS <GitHub Issue #833>

papaya2k:
Addresses GitHub Issue #833 by reducing the logging level of a CACHE_MISS in the AuthenticationResultSupplier to debug level.

Avery-Dunn:
> I agree that a cache miss exception should cause a debug log, since it doesn't necessarily mean that something wrong. It'll be a quick fix so will likely be in our next release, and I'll update this thread once we have some ETAs

Without this patch, the logging output is overly verbose about CACHE_MISS events.  This should resolve that issue by changing the logging level.


bgavrilMS (approved):

Avery-Dunn:
Looks good to me @papaya2k, thanks for fixing this!

Avery-Dunn (approved):

Issue #732, opened 2023-10-19
[Feature Request] MSAL Java support for CIAM custom authority


bgavrilMS:
### MSAL client type

Public, Confidential

### Problem Statement

CIAM will support custom authorities, which will require MSAL to perform OIDC discovery of the endpoints.

Blocked until we have a test environment.

### Proposed solution

Recommendation is for MSAL to add support to WithGenericAuthority for user flows (consider renaming to WithOidcAuthority).

Since this is a general-purpose mechanism, we should also try to use a non-Microsoft authority, such as Facebook or Google. Concern is that client_info missing will mess up token caching logic

Avery-Dunn:
Support added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/851 and released in 1.17.0

PR #851, opened 2024-08-06
Add support for CIAM custom authority

Avery-Dunn:
Adds support for CIAM custom authorities through a new API in the client application builder that handles OIDC compliant authorities, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/732

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
@@ -89,6 +89,25 @@ class ClientCredentialsIT {
     }
 
     @Test
+    void acquireTokenClientCredentials_Certificate_CiamCud() throws Exception {
+        String authorityCud = "https://login.msidlabsciam.com/fe362aec-5d43-45d1-b730-9755e60dc3b9/v2.0/";
+        String clientId = "b244c86f-ed88-45bf-abda-6b37aa482c79";
+
+        ConfidentialClientApplication cca = ConfidentialClientApplication.builder(
+                        clientId, CertificateHelper.getClientCertificate())
+                .oidcAuthority(authorityCud)
+                .build();
+
+        IAuthenticationResult result = cca.acquireToken(ClientCredentialParameters
+                        .builder(Collections.singleton(TestConstants.DEFAULT_SCOPE))
+                        .build())
+                .get();
+
+        assertNotNull(result);
+        assertNotNull(result.accessToken());
+    }
+
+    @Test
     void acquireTokenClientCredentials_Callback() throws Exception {
         String clientId = TestConstants.MSIDLAB_CLIENT_ID;
 
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
@@ -179,7 +179,7 @@ class ClientCredentialsIT {
                 clientId,
                 (ClientCertificate) certificate,
                 "https://login.microsoftonline.com/common/oauth2/v2.0/token",
-                true);
+                true, false);
     }
 
     private void assertAcquireTokenCommon(String clientId, IClientCredential credential, String authority) throws Exception {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
@@ -29,4 +29,13 @@ class AadInstanceDiscoveryResponse {
 
     @JsonProperty("correlation_id")
     private String correlationId;
+
+    @JsonProperty("authorization_endpoint")
+    private String authorizationEndpoint;
+
+    @JsonProperty("token_endpoint")
+    private String tokenEndpoint;
+
+    @JsonProperty("device_authorization_endpoint")
+    private String deviceCodeEndpoint;
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
@@ -29,13 +29,4 @@ class AadInstanceDiscoveryResponse {
 
     @JsonProperty("correlation_id")
     private String correlationId;
-
-    @JsonProperty("authorization_endpoint")
-    private String authorizationEndpoint;
-
-    @JsonProperty("token_endpoint")
-    private String tokenEndpoint;
-
-    @JsonProperty("device_authorization_endpoint")
-    private String deviceCodeEndpoint;
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
@@ -558,21 +558,16 @@ public abstract class AbstractClientApplicationBase extends AbstractApplicationB
                         builder.httpClient)
         );
 
-        //In most cases we will retrieve various auth/token/etc. via instance discovery during a token request,
-        //  the format is well known enough to be stored in the codebase itself, or the instance discovery result is
-        //  provided at app creation so we can cache it and skip discovery.
-        //However, if the authority being set is a generic OIDC authority (rather than Azure AD-style) then we need to perform
-        //  instance discovery when the client app is created, as there are some app-level methods that require the endpoints from
-        //  the instance discovery metadata.
         if (aadAadInstanceDiscoveryResponse != null) {
             AadInstanceDiscoveryProvider.cacheInstanceDiscoveryResponse(
                     authenticationAuthority.host,
                     aadAadInstanceDiscoveryResponse);
-        } else if (authenticationAuthority.authorityType == AuthorityType.OIDC) {
-            AadInstanceDiscoveryProvider.doOidcInstanceDiscoveryAndCache((OidcAuthority) authenticationAuthority, this, serviceBundle);
+        }
+
+        if (authenticationAuthority.authorityType == AuthorityType.OIDC) {
             ((OidcAuthority) authenticationAuthority).setAuthorityProperties(
-                    AadInstanceDiscoveryProvider.doOidcInstanceDiscoveryAndCache(
-                            (OidcAuthority) authenticationAuthority, this, serviceBundle));
+                    OidcDiscoveryProvider.performOidcDiscovery(
+                            (OidcAuthority) authenticationAuthority, this));
         }
     }
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
@@ -4,22 +4,17 @@
 package com.microsoft.aad.msal4j;
 
 import com.nimbusds.oauth2.sdk.auth.ClientAuthentication;
-import lombok.AccessLevel;
 import lombok.Getter;
 import lombok.experimental.Accessors;
-import org.slf4j.Logger;
 
 import javax.net.ssl.SSLSocketFactory;
 import java.net.MalformedURLException;
 import java.net.Proxy;
 import java.net.URL;
 import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
 import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ExecutorService;
-import java.util.function.Consumer;
 
 import static com.microsoft.aad.msal4j.ParameterValidationUtils.validateNotBlank;
 import static com.microsoft.aad.msal4j.ParameterValidationUtils.validateNotNull;
@@ -563,10 +558,21 @@ public abstract class AbstractClientApplicationBase extends AbstractApplicationB
                         builder.httpClient)
         );
 
+        //In most cases we will retrieve various auth/token/etc. via instance discovery during a token request,
+        //  the format is well known enough to be stored in the codebase itself, or the instance discovery result is
+        //  provided at app creation so we can cache it and skip discovery.
+        //However, if the authority being set is a generic OIDC authority (rather than Azure AD-style) then we need to perform
+        //  instance discovery when the client app is created, as there are some app-level methods that require the endpoints from
+        //  the instance discovery metadata.
         if (aadAadInstanceDiscoveryResponse != null) {
             AadInstanceDiscoveryProvider.cacheInstanceDiscoveryResponse(
                     authenticationAuthority.host,
                     aadAadInstanceDiscoveryResponse);
+        } else if (authenticationAuthority.authorityType == AuthorityType.GENERIC) {
+            AadInstanceDiscoveryProvider.doOidcInstanceDiscoveryAndCache((GenericAuthority) authenticationAuthority, this, serviceBundle);
+            ((GenericAuthority) authenticationAuthority).setAuthorityProperties(
+                    AadInstanceDiscoveryProvider.doOidcInstanceDiscoveryAndCache(
+                            (GenericAuthority) authenticationAuthority, this, serviceBundle));
         }
     }
 }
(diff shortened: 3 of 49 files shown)


msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
```diff
@@ -88,6 +88,25 @@ void acquireTokenClientCredentials_ClientSecret_Ciam() throws Exception {
         assertNotNull(result.accessToken());
     }
 
+    @Test
+    void acquireTokenClientCredentials_Certificate_CiamCud() throws Exception {
```
bgavrilMS, line 92: Although this is useful, the more important flow is auth code. Are there any tests for auth code ? CCA auth code please, not PCA

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/GenericAuthority.java
```diff
@@ -0,0 +1,42 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+
+public class GenericAuthority extends Authority {
```
bgavrilMS: OidcAuthority ? To match the name of the public API?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -79,11 +80,11 @@ static AuthorityType detectAuthorityType(URL authorityUrl) {
 
         final String path = authorityUrl.getPath().substring(1);
         if (StringHelper.isBlank(path)) {
-            if(isCiamAuthority(authorityUrl.getHost())){
+            if (isCiamAuthority(authorityUrl.getHost())) {
                 return AuthorityType.CIAM;
+            } else {
+                return AuthorityType.GENERIC;
```
bgavrilMS: I don't think this is correct. You cannot make any assumption on the format of the generic authority. This is why we decided to use a different API "oidcAuthority" - the app developer has to tell MSAL about it. 

if the existing "authority" API is used and MSAL can't detect the authority, then it defaults to AAD.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -63,8 +62,10 @@ static Authority createAuthority(URL authorityUrl) throws MalformedURLException{
             createdAuthority = new B2CAuthority(authorityUrl);
         } else if (authorityType == AuthorityType.ADFS) {
             createdAuthority = new ADFSAuthority(authorityUrl);
-        } else if(authorityType == AuthorityType.CIAM){
```
bgavrilMS, line 66: The fact that you modified this existing code and no unit test broke is an indication of a test gap. Please add a unit test to get coverage of this logic.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/GenericAuthority.java
```diff
@@ -0,0 +1,42 @@
...
+    static final String AUTHORIZATION_ENDPOINT = "oauth2/v2.0/authorize";
```
bgavrilMS: This is not correct. You have to make a call to oidc discovery endpoint (authority + ".well-known/openid-configuration"), and read the authorization endpoint and token endpoint from there. this call should be cached statically (and maybe unit tests need to reset this cache).

Pls see .NET implementation.

Avery-Dunn: In my original implementation I saw that the endpoints in the ID labs test environment followed the same "oauth2/v2.0/authorize" we used elsewhere and assumed that was part of the OIDC standard we could expect.

In the latest commit the endpoints are now only set through whatever is found during instance discovery at the ".well-known/openid-configuration" endpoint.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/GenericAuthority.java
Avery-Dunn: I named it 'GenericAuthority' to match the style in .NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Instance/GenericAuthority.cs

However, I noticed that .NET's authority classes are not public (?), and agree that it's better for the public APIs in one library to match than it is have this bit of SDK consistency. In the latest commit it's been changed to 'OidcAuthority'

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/ClientCredentialsIT.java
Avery-Dunn: In the latest commit I added an integration test for auth code: all CCA, asserts that the authority and authorization endpoints are being set as expected, gets a token with the auth code flow, and does the silent flow to ensure it is cached correctly.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
Avery-Dunn: It turns out these changes weren't needed at all: this `createAuthority` method is only used in a few places that had to handle the special multicloud/instance aware scenario, which I don't think is relevant when using the new OidcAuthority API

I reverted the logic back to what it was in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
Avery-Dunn: This change was unnecessary and revert in the latest commit (more context in your other comment in this file: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/851#discussion_r1705299331)

We have existing unit tests covering some of these methods with various authority types (https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AuthorityTest.java), however none of them broke because they all succeeded before they would have gotten to the new code.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
```diff
@@ -0,0 +1,33 @@
...
+
+    private static URL transformAuthority(URL originalAuthority) throws MalformedURLException {
+        String transformedAuthority = originalAuthority.toString();
+        transformedAuthority += WELL_KNOWN_OPENID_CONFIGURATION;
+
+        return new URL(transformedAuthority);
+    }
+
+    void setAuthorityProperties(AadInstanceDiscoveryResponse instanceDiscoveryResponse) {
+        this.authorizationEndpoint = instanceDiscoveryResponse.authorizationEndpoint();
```
bgavrilMS, line 28: I think you are conflating instance discovery and oidc discovery. These are different http calls, with different responses etc. Instance Discovery is an AAD invention. OIDC discovery is based on the OIDC protocol. 

For AAD, B2C, ADFS and CIAM, MSAL4J chose not skip OIDC discovery and hardcode the endpoints.
For CIAM CUD / OIDC authority, this is no longer possible. OIDC discoevry needs to happen.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
```diff
@@ -0,0 +1,33 @@
...
+    static final String WELL_KNOWN_OPENID_CONFIGURATION = ".well-known/openid-configuration";
+    private static final String AUTHORITY_FORMAT = "https://%s/%s/";
+
+    OidcAuthority(URL authorityUrl) throws MalformedURLException {
+        super(transformAuthority(authorityUrl), AuthorityType.OIDC);
+
+        this.authority = String.format(AUTHORITY_FORMAT, host, tenant);
+    }
+
+    private static URL transformAuthority(URL originalAuthority) throws MalformedURLException {
```
bgavrilMS: This name is confusing. THe authority url suffers no transformation. You are just computing the oidc discovery URL

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
```diff
@@ -29,4 +29,13 @@ class AadInstanceDiscoveryResponse {
 
     @JsonProperty("correlation_id")
     private String correlationId;
+
+    @JsonProperty("authorization_endpoint")
+    private String authorizationEndpoint;
```
bgavrilMS: I recommend you create a new response class for OIDC discovery instead of reusing the Instance Discovery class.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
bgavrilMS: I recommend you create more abstraction to model this complexity. For example you can model both through interfaces InstanceDiscovery and OidcDiscovery. MSAL would use both to compute urls. And provide 2 implementations for each: 

## Instance 
- NoInstanceDiscovery - which applies to everything but AAD. 
- AadInstanceDiscovery - which applies to AAD

## OIDC
- HardcodedOidcDIscovery - which applies to everything but OIDC
- OidcDiscovery - which applies to OIDC

Thoughts? (just an idea, you undertamnd the object model better)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CIAMAuthority.java
```diff
@@ -6,7 +6,7 @@
 import java.net.MalformedURLException;
 import java.net.URL;
 
-public class CIAMAuthority extends Authority{
+public class CIAMAuthority extends Authority {
```
bgavrilMS, line 9: Pls add a unit test to ensure that OIDC discovery doesn't happen twice for a Generic Authority.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -545,10 +558,21 @@ public T correlationId(String val) {
                         builder.httpClient)
         );
 
+        //In most cases we will retrieve various auth/token/etc. via instance discovery during a token request,
+        //  the format is well known enough to be stored in the codebase itself, or the instance discovery result is
+        //  provided at app creation so we can cache it and skip discovery.
+        //However, if the authority being set is a generic OIDC authority (rather than Azure AD-style) then we need to perform
+        //  instance discovery when the client app is created, as there are some app-level methods that require the endpoints from
+        //  the instance discovery metadata.
         if (aadAadInstanceDiscoveryResponse != null) {
```
bgavrilMS: It's not a great behavior to fire HTTP requests when creating the "app" object. But maybe add a bug for this to revisit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
Avery-Dunn: This was just something copied from another authority class, and in that one this method did more to the authority. In the latest commit this is now 'createOidcDiscoveryUrl' to be more clear about what it's doing.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
Avery-Dunn: In the latest commit I moved the OIDC discovery logic into some new classes, and restored the AAD instance discovery classes to what they were before these changes. Now the instance discovery classes only do their behavior when the authority type is AAD, and and the OIDC discovery classes only do their behavior when the authority type is OIDC.

However, what do you mean by "HardcodedOidcDiscovery - which applies to everything but OIDC"? Do all other authority types need to make a call to some hardcoded version of that ".well-known/openid-configuration" endpoint?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryResponse.java
Avery-Dunn: In the latest commit these fields are now in a new `OidcDiscoveryResponse` class.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
Avery-Dunn: I agree it's not ideal, however this is the easiest way to do it because of the `authenticationAuthority` field in AbstractApplicationBase which is used to reference the endpoints found in the OIDC discovery: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/424eea69c795e122171bd6335c253767042519a1/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java#L32

That field is used in a couple dozen places, such as:
- A method called in the confidential client app's constructor: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/424eea69c795e122171bd6335c253767042519a1/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java#L94
- A public helper method for creating the URL for an auth code request: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/424eea69c795e122171bd6335c253767042519a1/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java#L173
- Parameter classes for various token request flows

If the OIDC discovery doesn't happen before that field gets referenced, then all of the endpoints will be null. It's definitely something we can revisit and refactor later, but for now we either make the HTTP call in this one place during app creation, or add checks to a bunch of different places to make sure it get initialized.

bgavrilMS: Pls log a bug, it's a P3 as customers have not complained, and this is existing behavior. We should at least track this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -105,6 +105,22 @@ IHttpResponse executeHttpRequest(HttpRequest httpRequest,
         return httpResponse;
     }
 
+    IHttpResponse executeHttpRequest(HttpRequest httpRequest) {
+        IHttpResponse httpResponse;
+
+        try {
+            httpResponse = executeHttpRequestWithRetries(httpRequest, httpClient);
+        } catch (Exception e) {
+            throw new MsalClientException(e);
```
bgavrilMS, line 114: Is it really MsalClientException? Service errors should be MsalServiceException?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OidcAuthority.java
bgavrilMS: MSAL needs to figure out the token endpoiint, authorization endpoint, device_code enpoint of an authority. For performance reasons, to avoid an extra HTTP call, for AAD, B2C, CIAM, we will hardcode these and skip OIDC discovery. THis is already the behavior of MSAL 4j. 

For generic authorities, OIDC discovery (via HTTP call) needs to happen.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
bgavrilMS: Do you have tests covering this?

bgavrilMS (approved):

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
Avery-Dunn: Just created a task to track this: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/856

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
Avery-Dunn: This is the same exception type as the other versions of the `executeHttpRequest`, and after a quick look through the file history it's been that exception type since 2019.

It might be worth looking into exception types as an SDK consistency item, but to keep consistent with existing MSAL and avoiding breaking changes I'm leaving this as `MsalClientException`

Issue #848, opened 2024-07-31
Ensure, refreshOn is computed to half the token lifetime for Managed Identity Tokens


g2vinay:
### Summary:
Implement logic to compute and return the refreshOn value as half the token lifetime for long-lived Managed Identity tokens (those with a token expiry of >2 hours). This adjustment will align with Azure Identity standards and prevent intermittent token refresh failures for our Enterprise customers, such as Cosmos.

### Background:
The current implementation does not account for the unique requirements of long-lived Managed Identity tokens, which have a token expiry of more than 2 hours. As a result, our Enterprise customers have occasionally experienced intermittent token refresh failures. This proposed change aims to mitigate such issues by ensuring a more predictable and consistent token refresh process.

### Details:

Compute refreshOn Value:

For Managed Identity tokens with a token expiry of more than 2 hours, calculate the refreshOn value as half the token's lifetime. For example, if a token has a 4-hour lifetime, the refreshOn value should be set to 2 hours.
This approach ensures that the token is refreshed well before it expires, providing a buffer period to handle any potential refresh issues.

### Consistency with Azure Identity:

This adjustment aligns our token refresh strategy with Azure Identity standards, promoting consistency across our services and improving overall reliability.

### Enterprise Customer Impact:

By implementing this change, we can enhance the experience for our Enterprise customers, such as Cosmos, by reducing the likelihood of token refresh failures and ensuring seamless operations.
Implementation Plan:

### Code Modification:
Identify the section of the codebase responsible for calculating the refreshOn value for Managed Identity tokens.
Introduce logic to check the token expiry duration. If the duration is more than 2 hours , calculate refreshOn as half of the token lifetime.
If, refresh In is coming back as part of response, honor it.

### Testing:
Develop unit tests to validate the new refreshOn calculation logic.
Conduct integration testing to ensure that the feature works correctly in various scenarios, including different token lifetimes and renewal intervals.

### Documentation:
Update internal documentation to reflect the new logic for calculating the refreshOn value.

### Conclusion:
Adjusting the refreshOn calculation for long-lived Managed Identity tokens will significantly enhance the reliability and consistency of our token refresh process. This change will align our practices with Azure Identity standards and provide a better experience for our Enterprise customers, such as Cosmos, by minimizing the risk of intermittent token refresh failures.

bgavrilMS:
Agreed, this is a consistency item with other MSALs.

rayluo:
[...]

PR #858, opened 2024-08-27
Set refreshOn in MI flows to half of token lifetime

Avery-Dunn:
Adjusts the refreshOn calculation in MI flows to be half of a token's lifetime (if that lifetime is greater than two hours), see https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/848 for details.

Additionally, this PR removes the static caching of managedIdentitySource, since determining the source is cheap but caching it statically can cause issues in test environments. This same problem that was also found in MSAL .NET and this same fix was done in https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4840

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
@@ -13,6 +13,8 @@ class AcquireTokenByManagedIdentitySupplier extends AuthenticationResultSupplier
 
     private static final Logger LOG = LoggerFactory.getLogger(AcquireTokenByManagedIdentitySupplier.class);
 
+    private static final int TWO_HOURS = 2*3600;
+
     private ManagedIdentityParameters managedIdentityParameters;
 
     AcquireTokenByManagedIdentitySupplier(ManagedIdentityApplication managedIdentityApplication, MsalRequest msalRequest) {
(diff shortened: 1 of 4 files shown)


bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -93,15 +95,27 @@ private AuthenticationResult fetchNewAccessTokenAndSaveToCache(TokenRequestExecu
...
+                .metadata(metadata)
                 .build();
     }
+
+    private long calculateRefreshOn(long expiresOn){
+        long timestampSeconds = System.currentTimeMillis() / 1000;
+        long expiresIn = expiresOn - timestampSeconds;
+
+        //The refreshOn value should be half the value of the token lifetime, if the lifetime is greater than two hours
+        return expiresIn > TWO_HOURS ? (expiresIn / 2) + timestampSeconds : 0;
```
bgavrilMS, line 119: Is there a difference between 0 and NULL (absence of a value) for refreshOn? 

In other words, is there a danger that MSAL interprets refreshOn = 0 as "always refresh" if expiry < 2h ?

Note: In MSAL.NET the implementation differentiates between NULL and int (0 or non-0). Not suggesting you do the same, I think it's fair to say that when `refreshOn = 0` or some other sentinel value, then you ignore it. Just want to make sure this is really the case.

Avery-Dunn: If you look at this part of AcquireTokenSilentSupplier you'll see one of the places where this value gets used internally: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/310cc07d912acc6d4e3371e569ba660c916f98fc/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java#L138

In that if statement 'null' and '0' are treated the same, if refreshon is either value then we won't use it as a reason to refresh. Anywhere else it gets used should have a similar a 'refreshon > 0' check.

Thinking about it now this does seem... fragile. It might we worth refactoring to always use 'null' (since misusing it would result in an exception rather than looking valid), however we seem to have the same behavior for expiry times and other values and there's always a chance a customer relies on certain behavior.

bgavrilMS: Thanks for confirming!

Issue #881, opened 2024-11-22
[Bug] On behalf of supplier doesn't pass tenant id to silent supplier


q-benwillis:
### Library version used

1.17.2

### Java version

17.0.11

### Scenario

ConfidentialClient - web api (AcquireTokenOnBehalfOf)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

It seems that when the OBO supplier attempts to acquire the token silently from the cache it doesn't pass the Tenant ID in the SilentParameters which means that the token cannot be found in the cache.

I'm sending OnBehalfOfParameters with the Tenant ID:
```
OnBehalfOfParameters.builder(scopes, userAssertion).tenant(tenantId).build
```

This is all I'm seeing in the logs:
```
2024-11-22T11:56:56.127Z DEBUG 24 --- [app] [onPool-worker-3] c.m.a.m.AcquireTokenByOnBehalfOfSupplier : SkipCache set to false. Attempting cache lookup
2024-11-22T11:56:56.128Z DEBUG 24 --- [app] [onPool-worker-3] c.m.a.m.AcquireTokenByOnBehalfOfSupplier : Cache lookup failed: Token not found in the cache
```

However if I dig a little deeper I can see that the cache miss is because the cached token realm doesn't match the authority tenant [here](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/081341c8d0afe8ca5d024f60047d45c5f440a5c7/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java#L501). If I follow this back I can see that the authority tenant can come from the request parameters [here](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/081341c8d0afe8ca5d024f60047d45c5f440a5c7/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultSupplier.java#L32). However it looks like this isn't happening because the `OnBehalfOfSupplier` doesn't add the tenant to the `SilentParameters` [here](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/081341c8d0afe8ca5d024f60047d45c5f440a5c7/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByOnBehalfOfSupplier.java#L28).

### Relevant code snippets

```csharp
val userAssertion = new UserAssertion(???)
val scopes = Set(???)
val tenantId = ???

[...]

PR #886, opened 2024-12-09
Pass optional tenant override to internal silent call

Avery-Dunn:
Fixes an issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/881. Essentially, there is an API at the request level which overrides the tenant set at the application level. 

However, in some confidential flows we internally make a silent call with a new SilentParameters object which did not receive the tenant that may have been set with that API.

This PR fixes the two places where that API wasn't getting properly passed to that internal silent call, adds a new test for that behavior, and makes a few unrelated fixes/improvements to some integration tests.

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
@@ -123,49 +123,6 @@ class OnBehalfOfIT {
         assertNotEquals(result6.accessToken(), result2.accessToken());
     }
 
-    @Test
-    void acquireTokenWithOBO_TenantOverride() throws Exception {
-        cfg = new Config(AzureEnvironment.AZURE);
-        String accessToken = this.getAccessToken();
-
-        final String clientId = cfg.appProvider.getOboAppId();
-        final String password = cfg.appProvider.getOboAppPassword();
-
-        ConfidentialClientApplication cca =
-                ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromSecret(password)).
-                        authority(cfg.tenantSpecificAuthority()).
-                        build();
-
-        //This token should be cached with the tenant-specific authority set at the application level
-        IAuthenticationResult resultNoOverride = cca.acquireToken(OnBehalfOfParameters.builder(
-                                Collections.singleton(cfg.graphDefaultScope()),
-                                new UserAssertion(accessToken)).build()).
-                        get();
-
-        //This token should be cached with an 'organizations' authority set at the request level
-        IAuthenticationResult resultOrganizations = cca.acquireToken(OnBehalfOfParameters.builder(
-                                Collections.singleton(cfg.graphDefaultScope()),
-                                new UserAssertion(accessToken))
-                                .tenant("organizations")
-                                .build()).
-                        get();
-
-        //This token should come from the cache and match the token with the 'organizations' authority
-        IAuthenticationResult resultOrganizationsCached = cca.acquireToken(OnBehalfOfParameters.builder(
-                                Collections.singleton(cfg.graphDefaultScope()),
-                                new UserAssertion(accessToken))
-                                .tenant("organizations")
-                                .build()).
-                        get();
-
-        assertResultNotNull(resultNoOverride);
-        assertResultNotNull(resultOrganizations);
-        assertResultNotNull(resultOrganizationsCached);
-
-        assertNotEquals(resultNoOverride.accessToken(), resultOrganizations.accessToken());
-        assertEquals(resultOrganizations.accessToken(), resultOrganizationsCached.accessToken());
-    }
-
     private void assertResultNotNull(IAuthenticationResult result) {
         assertNotNull(result);
         assertNotNull(result.accessToken());
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
@@ -39,8 +39,7 @@ class OnBehalfOfIT {
                         new UserAssertion(accessToken)).build()).
                         get();
 
-        assertNotNull(result);
-        assertNotNull(result.accessToken());
+        assertResultNotNull(result);
     }
 
     @ParameterizedTest
@@ -63,8 +62,7 @@ class OnBehalfOfIT {
                         new UserAssertion(accessToken)).build()).
                         get();
 
-        assertNotNull(result1);
-        assertNotNull(result1.accessToken());
+        assertResultNotNull(result1);
 
         // Same scope and userAssertion, should return cached tokens
         IAuthenticationResult result2 =
@@ -82,8 +80,7 @@ class OnBehalfOfIT {
                         new UserAssertion(accessToken)).build()).
                         get();
 
-        assertNotNull(result3);
-        assertNotNull(result3.accessToken());
+        assertResultNotNull(result3);
         assertNotEquals(result2.accessToken(), result3.accessToken());
 
         // Scope 2, should return cached token
@@ -105,8 +102,7 @@ class OnBehalfOfIT {
                                 .build()).
                         get();
 
-        assertNotNull(result5);
-        assertNotNull(result5.accessToken());
+        assertResultNotNull(result5);
         assertNotEquals(result5.accessToken(), result4.accessToken());
         assertNotEquals(result5.accessToken(), result2.accessToken());
 
@@ -121,11 +117,58 @@ class OnBehalfOfIT {
                                 .build()).
                         get();
 
-        assertNotNull(result6);
-        assertNotNull(result6.accessToken());
+        assertResultNotNull(result6);
         assertNotEquals(result6.accessToken(), result5.accessToken());
         assertNotEquals(result6.accessToken(), result4.accessToken());
         assertNotEquals(result6.accessToken(), result2.accessToken());
+    }
+
+    @Test
+    void acquireTokenWithOBO_TenantOverride() throws Exception {
+        cfg = new Config(AzureEnvironment.AZURE);
+        String accessToken = this.getAccessToken();
+
+        final String clientId = cfg.appProvider.getOboAppId();
+        final String password = cfg.appProvider.getOboAppPassword();
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromSecret(password)).
+                        authority(cfg.tenantSpecificAuthority()).
+                        build();
+
+        //This token should be cached with the tenant-specific authority set at the application level
+        IAuthenticationResult resultNoOverride = cca.acquireToken(OnBehalfOfParameters.builder(
+                                Collections.singleton(cfg.graphDefaultScope()),
+                                new UserAssertion(accessToken)).build()).
+                        get();
+
+        //This token should be cached with an 'organizations' authority set at the request level
+        IAuthenticationResult resultOrganizations = cca.acquireToken(OnBehalfOfParameters.builder(
+                                Collections.singleton(cfg.graphDefaultScope()),
+                                new UserAssertion(accessToken))
+                                .tenant("organizations")
+                                .build()).
+                        get();
+
+        //This token should come from the cache and match the token with the 'organizations' authority
+        IAuthenticationResult resultOrganizationsCached = cca.acquireToken(OnBehalfOfParameters.builder(
+                                Collections.singleton(cfg.graphDefaultScope()),
+                                new UserAssertion(accessToken))
+                                .tenant("organizations")
+                                .build()).
+                        get();
+
+        assertResultNotNull(resultNoOverride);
+        assertResultNotNull(resultOrganizations);
+        assertResultNotNull(resultOrganizationsCached);
+
+        assertNotEquals(resultNoOverride.accessToken(), resultOrganizations.accessToken());
+        assertEquals(resultOrganizations.accessToken(), resultOrganizationsCached.accessToken());
+    }
+
+    private void assertResultNotNull(IAuthenticationResult result) {
+        assertNotNull(result);
+        assertNotNull(result.accessToken());
     }
 
     private String getAccessToken() throws Exception {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
@@ -26,6 +26,7 @@ class AcquireTokenByClientCredentialSupplier extends AuthenticationResultSupplie
                 SilentParameters parameters = SilentParameters
                         .builder(this.clientCredentialRequest.parameters.scopes())
                         .claims(this.clientCredentialRequest.parameters.claims())
+                        .tenant(this.clientCredentialRequest.parameters.tenant())
                         .build();
 
                 RequestContext context = new RequestContext(
--- a/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
+++ b/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
@@ -20,28 +20,11 @@ import static org.mockito.Mockito.times;
 @ExtendWith(MockitoExtension.class)
 class OnBehalfOfTests {
 
-    private String getSuccessfulResponse(String accessToken) {
-        return "{\"access_token\":\""+accessToken+"\",\"expires_in\": \""+ 60*60*1000 +"\",\"token_type\":" +
-                "\"Bearer\",\"client_id\":\"client_id\",\"Content-Type\":\"text/html; charset=utf-8\"}";
-    }
-
-    private HttpResponse expectedResponse(int statusCode, String response) {
-        Map<String, List<String>> headers = new HashMap<String, List<String>>();
-        headers.put("Content-Type", Collections.singletonList("application/json"));
-
-        HttpResponse httpResponse = new HttpResponse();
-        httpResponse.statusCode(statusCode);
-        httpResponse.body(response);
-        httpResponse.addHeaders(headers);
-
-        return httpResponse;
-    }
-
     @Test
     void OnBehalfOf_InternalCacheLookup_Success() throws Exception {
         DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
 
-        when(httpClientMock.send(any(HttpRequest.class))).thenReturn(expectedResponse(200, getSuccessfulResponse("token")));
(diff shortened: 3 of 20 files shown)


msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
```diff
@@ -121,13 +117,60 @@ void acquireTokenWithOBO_testCache(String environment) throws Exception {
...
+                                new UserAssertion(accessToken))
+                                .tenant("organizations")
+                                .build()).
+                        get();
+
+        assertResultNotNull(resultNoOverride);
+        assertResultNotNull(resultOrganizations);
+        assertResultNotNull(resultOrganizationsCached);
+
+        assertNotEquals(resultNoOverride.accessToken(), resultOrganizations.accessToken());
```
bgavrilMS: nit: you really should get that improvement of exposing a `fromCache` flag in the auth result going. This string comparison is not great. ESTS-R for example caches tokens.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
```diff
@@ -121,13 +117,60 @@ void acquireTokenWithOBO_testCache(String environment) throws Exception {
...
+        IAuthenticationResult resultNoOverride = cca.acquireToken(OnBehalfOfParameters.builder(
+                                Collections.singleton(cfg.graphDefaultScope()),
+                                new UserAssertion(accessToken)).build()).
+                        get();
+
+        //This token should be cached with an 'organizations' authority set at the request level
+        IAuthenticationResult resultOrganizations = cca.acquireToken(OnBehalfOfParameters.builder(
+                                Collections.singleton(cfg.graphDefaultScope()),
+                                new UserAssertion(accessToken))
+                                .tenant("organizations")
```
bgavrilMS: I am not sure I agree with this test. Afaik MSAL will use the tenant ID from the response, not from the request, so it's not that clear what this test does. It's also showcases 2 anti-patterns:

1. go from a tenant specific authority on the CCA to the `lmo/organizations` authority on the request. Usually it's the other way around.
1. OBO with "/common" or "/organizations". This is wrong, OBO should always be tenanted. It should use the client token's `tid` claim. Otherwise you introduce a bug where guest users can't login. See https://learn.microsoft.com/en-us/entra/msal/dotnet/acquiring-tokens/web-apps-apis/on-behalf-of-flow#important-note-on-on-behalf-of-obo-flow-with-guest-users


My recommendation is to rely on unit tests for this one: 

- app authority is { "lmo/organizations", "lmo/T1", "lmo/T2" }
- req tenant is {T1, T2}
- flow { OBO, client_creds, silent}

So there are quite a few unit tests (3x2x3), but mostly repetitive.

Then Assert on the correct MSAL behavior i.e. token from cache or from the correctly tenanted token endpoint

bgavrilMS (changes requested):
Needs a unit test

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/OnBehalfOfIT.java
Avery-Dunn: Agreed that unit tests would be better, originally I just did an integration test since there were ones to copy from but we didn't have any OBO unit tests.

In the latest commit I replaced this integration test with a couple of unit tests. It's not as many as you suggest, but they should show the correct behavior by checking for the correct number of HTTPClient calls, the token cache size, etc., depending on whether the tenant was set at the request level or not.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
+
+@ExtendWith(MockitoExtension.class)
+class OnBehalfOfTests {
+
+    private String getSuccessfulResponse(String accessToken) {
+        return "{\"access_token\":\""+accessToken+"\",\"expires_in\": \""+ 60*60*1000 +"\",\"token_type\":" +
+                "\"Bearer\",\"client_id\":\"client_id\",\"Content-Type\":\"text/html; charset=utf-8\"}";
```
bgavrilMS: 1. Consider adding this to a common file. It's the standard /token response for S2S
2. OBO receives id tokens, refresh tokens etc. The response should reflect that. 
3. I am not aware of the COntent-Type being part of the response content. It may be a header?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    void OnBehalfOf_InternalCacheLookup_Success() throws Exception {
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+
+        when(httpClientMock.send(any(HttpRequest.class))).thenReturn(expectedResponse(200, getSuccessfulResponse("token")));
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder("clientId", ClientCredentialFactory.createFromSecret("password"))
```
bgavrilMS, line 30: Consider using some constants for ClientID.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
+import static org.mockito.Mockito.times;
+
+@ExtendWith(MockitoExtension.class)
+class OnBehalfOfTests {
+
+    private String getSuccessfulResponse(String accessToken) {
```
bgavrilMS, line 23: nit: the access token is never relevant. not sure why you'd want this to be a param in this function. Just use "at" or smth

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+        return "{\"access_token\":\""+accessToken+"\",\"expires_in\": \""+ 60*60*1000 +"\",\"token_type\":" +
```
bgavrilMS: You also need to add a client_info, that is part of all user token protocols.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder("clientId", ClientCredentialFactory.createFromSecret("password"))
+                        .authority("https://login.microsoftonline.com/tenant")
+                        .instanceDiscovery(false)
+                        .validateAuthority(false)
+                        .httpClient(httpClientMock)
+                        .build();
+
+        when(httpClientMock.send(any(HttpRequest.class))).thenReturn(expectedResponse(200, getSuccessfulResponse("appTenantToken")));
```
bgavrilMS: THis is a great path forward for tests. Consider making a separate class with all this logic. It should the standard way of doing tests and little by little all tests should be re-written to use this.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    @Test
+    void OnBehalfOf_TenantOverride() throws Exception {
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder("clientId", ClientCredentialFactory.createFromSecret("password"))
+                        .authority("https://login.microsoftonline.com/tenant")
+                        .instanceDiscovery(false)
```
bgavrilMS, line 54: ok for now, but the http mock can easily handle this as well.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
```diff
@@ -0,0 +1,95 @@
...
+                        .validateAuthority(false)
+                        .httpClient(httpClientMock)
+                        .build();
+
+        OnBehalfOfParameters parameters = OnBehalfOfParameters.builder(Collections.singleton("scopes"), new UserAssertion(TestHelper.signedToken)).build();
+
+        IAuthenticationResult result = cca.acquireToken(parameters).get();
+        IAuthenticationResult result2 = cca.acquireToken(parameters).get();
+
+        //OBO flow should perform an internal cache lookup, so similar parameters should only cause one HTTP client call
```
bgavrilMS, line 42: So do you confirm that the lookup is done by assertion_hash in this case?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
```diff
@@ -27,4 +38,21 @@ static void deleteFileContent(Class<?> classInstance, String resource)
         fileWriter.write("");
         fileWriter.close();
     }
+
+    static String generateToken() {
```
bgavrilMS, line 50: This is fine. Other MSALs just create a json to represent the JWK. MSALs do not validate headers nor signatures, so only the payload matters.

You don't really need a "proper" token for OBO. MSAL never looks into it. 

You will need a proper id token though.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
```diff
@@ -26,6 +26,7 @@ AuthenticationResult execute() throws Exception {
                 SilentParameters parameters = SilentParameters
                         .builder(this.clientCredentialRequest.parameters.scopes())
                         .claims(this.clientCredentialRequest.parameters.claims())
+                        .tenant(this.clientCredentialRequest.parameters.tenant())
```
bgavrilMS, line 29: This is not tested.

bgavrilMS (changes requested):
More testing is needed.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
bgavrilMS: You should only return that response once, not all the time, i.e. assert that only 1 call to the token endpoint is made.

bgavrilMS (approved):

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: It might be more clear in the latest commit since I refactored some of the code, but the `verify(httpClientMock, times(1)).send(any());` later on was verifying exactly that.

The 'send' method is essentially the final step which sends the request to the endpoint, and 'httpClientMock' was passed into the confidential client app so that mocked HTTP client is getting used for all HTTP calls in the test. So, verifying the number of times 'send' was called is just a quick way of counting the number of calls to the token endpoint.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: This was mainly a quick fix for counting the number of times I made a call to the token endpoint vs. the cache: I wasn't testing instance discovery anyway, and if I did do instance discovery I'd need to both mock it and count an extra call in every `verify(httpClientMock, times(...)).send()` check

It's definitely better to mock it and will set that up in a future PR, but I wanted to keep this PR focused on the new OBO tests and essentially be a proof-of-concept for how to improve the unit tests.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: I moved all of the "expectedResponse" creation to the TestHelper class in the latest commit, however I kept the "when.(...).then(...)" this class. It's something that can apply to many other tests, but I need to look into the other tests more to figure out a common/flexible way of doing it.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: In the latest commit this has been moved to TestHelper, and now has client_info, access/id/refresh tokens, etc.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: Normally I would, but it's only used twice in these tests and isn't checked anyway. The next time I need to make unit tests for app creation I'll refactor this and other similar code into a helper method.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/OnBehalfOfTests.java
Avery-Dunn: This was just copied from some similar tests in ManagedIdentityTests, and what I was looking at in these new tests didn't need anything different in the response so I didn't look at it too closely.

In the latest commit I moved it to TestHelper and restructured it to give it a lot more flexibility: the string is now properly formatted with placeholders, and the placeholders are optionally filled in based on values set in a map.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
Avery-Dunn: In the latest commit I changed the name of the field that uses this to be more clear, but this is just used to create an assertion specifically for the OBO tests.

MSAL doesn't validate it, but internally we have some OAuth libraries that use the assertion to create an authorization grant, and as part of that it does some parsing, checks that it's signed, etc., and it was just quicker to create a 'real' one using that same library than to try to fake it or mock that part of MSAL.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
```diff
@@ -27,4 +46,55 @@ static void deleteFileContent(Class<?> classInstance, String resource)
...
+        long expiresIn = responseValues.containsKey("expires_in") ?
+                Long.parseLong(responseValues.get("expires_in")) :
+                3600;
+        long expiresOn = responseValues.containsKey("expires_on")
+                ? Long.parseLong(responseValues.get("expires_0n")) :
+                (System.currentTimeMillis() / 1000) + expiresIn;
+
+        return String.format(successfulResponseFormat,
+                responseValues.getOrDefault("access_token", "access_token"),
+                responseValues.getOrDefault("id_token", "id_token"),
```
bgavrilMS, line 79: Hmm, does this just work? MSAL does parse the id token to get details such as `tid` etc.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
```diff
@@ -3,19 +3,38 @@
...
 class TestHelper {
 
-    static String readResource(Class<?> classInstance, String resource) throws IOException, URISyntaxException {
-        return new String(
-                Files.readAllBytes(
-                        Paths.get(classInstance.getResource(resource).toURI())));
+    //Signed JWT which should be enough to pass the parsing/validation in the library, useful if a unit test needs an
+    // assertion but that is not the focus of the test
+    static String signedAssertion = generateToken();
+    private static final String successfulResponseFormat = "{\"access_token\":\"%s\",\"id_token\":\"%s\",\"refresh_token\":\"%s\"," +
```
bgavrilMS, line 27: nit: successfulUserTokenResponseFormat

bgavrilMS (approved):

bgavrilMS (changes requested):
Still requires a test for client_credentials.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
bgavrilMS: Ok, thanks for confirming. It may be worth adding a comment around this in code, as it's quite surprising.

Avery-Dunn: Yeah, here's an example where parsing happens: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/95b5efc9fc30514719e5b678c739351ddcf35fba/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java#L36

It's some very old cold (that line is from 2019, so pre-1.0), and seems like the AuthorizationGrant is used to make query parameters (?). I think reducing our usage of that package was something that was talked about years ago, and it might be worth looking into again.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
Avery-Dunn: Added tests in the latest commit, essentially just a copy/paste of the OBO tests but use the type of Parameters that would go through this code.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
Avery-Dunn: Yes it just works, though it probably shouldn't. As we talked about yesterday we rely on that nimbus library, and you can see here we pass the values into their "OIDCTokens" method: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/95b5efc9fc30514719e5b678c739351ddcf35fba/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenResponse.java#L30

That doesn't through any error when we pass in something like "id_token". Instead, the instance of TokenResponse that gets made has a getOIDCTokens() method which has two relevant fields:
 - `idTokenString`, which is the actual string we give it ("id_token" in this case) 
 - `idToken`, which is JWT object that is simply null if we pass in something like "id_token"

And then later on in the library it seems like having a 'null' for `idToken` means that everything acts like it's a confidential client flow that had no ID token and nothing breaks. Based on what we talked about before I just created an issue about using those packages less: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/889

bgavrilMS (approved):

PR #888, opened 2024-12-17
Improve integration tests

Avery-Dunn:
Contains a number of fixes and improvements to the integration tests:
- Use 'common' authority instead of 'organizations' for ADFS tests
  - Fixes an issue caused by a change to a test app, and is also what is used in other MSALs
- Update queries to use newer labs resources
- Remove some unused code/constants/imports/etc., and refactor some common code into a new IntegrationTestHelper class
  - Just added some low-hanging fruit for now, more refactoring to come in future PRs when it makes sense

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
@@ -1,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+import java.net.MalformedURLException;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+class IntegrationTestHelper {
+
+    static PublicClientApplication createPublicApp(String appID, String authority) {
+        try {
+            return PublicClientApplication.builder(
+                            appID).
+                    authority(authority).
+                    build();
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    static void assertTokenResultNotNull(IAuthenticationResult result, boolean checkAccessToken, boolean checkIDToken) {
+        assertNotNull(result);
+        if (checkAccessToken) assertNotNull(result.accessToken());
+        if (checkIDToken) assertNotNull(result.idToken());
+    }
+}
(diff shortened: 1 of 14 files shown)


msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
```diff
@@ -0,0 +1,28 @@
...
+                    build();
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    static void assertTokenResultNotNull(IAuthenticationResult result, boolean checkAccessToken, boolean checkIDToken) {
+        assertNotNull(result);
+        if (checkAccessToken) assertNotNull(result.accessToken());
+        if (checkIDToken) assertNotNull(result.idToken());
```
bgavrilMS: You can add a few more asserts here. If the Id token is not null, the Account should also not be null.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
```diff
@@ -0,0 +1,28 @@
...
+                    authority(authority).
+                    build();
+        } catch (MalformedURLException e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    static void assertTokenResultNotNull(IAuthenticationResult result, boolean checkAccessToken, boolean checkIDToken) {
+        assertNotNull(result);
+        if (checkAccessToken) assertNotNull(result.accessToken());
```
bgavrilMS: I am not aware of any case where the access token is null?

bgavrilMS (approved):

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
Avery-Dunn: Me neither, I just added it because there were a few tests where we just checked if the result or the ID token was null but not the access token.

However, I just used a 'true' value for both token options in all the tests that now use this, so in the latest commit I changed this helper method to be `assertAccessAndIdTokensNotNull` and it will always check both tokens.

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/IntegrationTestHelper.java
Avery-Dunn: This was just meant to replace the "result & tokens aren't null" check the happens in a lot of tests and I wanted to keep the changes limited on the first PR. The tests that care about the account still have their own asserts, and the next time I work with those tests I'll add another helper method for checking various account info.

Issue #891, opened 2024-12-23
[Bug] Acquire token interactively with secure redirect URL


ErRijcken:
### Library version used

1.16.2

### Java version

17

### Scenario

PublicClient (AcquireTokenInteractive, AcquireTokenByUsernamePassword)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

We have a Java client where we would like to use MSAL4J for fetching a token (for authentication and authorization).
This worked well using `http://localhost:<port>` as a redirect URL, but our Windows people advised us that the security of this setup was not optimal (we have an on-premise authority which is configured by our people). They have talked to their MS contact, who recommended them to configure the redirect URL as `http://localhost:<port>/<randomString>/`.
I have been unable to make my MSAL4J setup work with this. Based on debugging the browser, it seems like the call to our authority does work, but the library doesn't properly treat the response.

### Identity provider

Azure Active Directory Federation Services (ADFS)

ErRijcken:
Ok, after a deep dive into the code (I'm by no means an expert), it seems like adding the random string in the redirect URL is not supported (see the if-block in  AuthorizationResponseHandler#handle).
In my mind, it should not be hard to modify this to allow a specific redirect path. I will try to make the change.

rayluo:
FWIW, I don't think adding a random string path would work; and adding a specific/static path can work but that has nothing to do with security, but still [useful as a feature to differentiate different redirect URIs](https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/593#issuecomment-1755890166).

ErRijcken:
I agree that the security-gain is probably negligible, some extra obscurity at most. I still agree that it would be a useful change, I made the change and created a pull request: #892. I have no experience in this library, but the code change seemed rather small. I'm not sure if I should close this issue and leave the rest of the discussion in the pull request?

PR #892, opened 2024-12-24
Use path of the redirect URI

ErRijcken:
Use the path as given by the redirect URI to allow for more specific redirect URIs. See issue #891.


ErRijcken:
@microsoft-github-policy-service agree

Issue #863, opened 2024-09-24
[Bug] IClientCredential public interface is broken by the SHA256 changes


bgavrilMS:
### Library version used

1.17.1

### Java version

8

### Scenario

ConfidentialClient - web site (AcquireTokenByAuthCode), ConfidentialClient - web api (AcquireTokenOnBehalfOf), ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

1. Use a version of MSAL < 1.16.2
2. Provide a custom implementation of `IClientCertificate` interface. Implementation of `publicCertificateHash` returns `base64(cert_x5t)` as per docs
3. Acquire a token in confidential client, e.g. an app token. This works fine
4. Update to MSAL 1.16.2 or later

Expected: token 
Actual: ESTS error complaining about the signature of the client assertion

Root cause: 

- With 1.16.2 MSAL uses `x5t#sha256` in the header, breaking existing custom implementations of `IClientCertificate` which return the `x5t`

Possible solutions: 

1. Deprecate the interface and create a new one, IClientCertificate2 instead (kind of ugly, but this is not really expected to be used by many). The problem is that for ADFS, which still requires x5t, it's kind of ugly.

2. Add a new method to the interface "publicCertificateHash256" with a [default implementation](https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html) of returning null. Then MSAL shall call `publicCertificateHash256` and if it returns null, it will call the older `publicCertificateHash`. Depending on this, the assertion will use x5t or x5t#s256 

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

Avery-Dunn:
Fixed as part of https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/898 and released in version 1.19.0

PR #840, opened 2024-07-14
Use SHA256 thumbprints in non-ADFS cert flows

Avery-Dunn:
Updates the internal thumbprint method for certificate flows from SHA-1 to SHA-256, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/760

ADFS scenarios must still use SHA-1 until SHA-256 is supported server-side.

PR #898, opened 2025-01-10
Fix regression issue in IClientCertificate caused by SHA256 changes

Avery-Dunn:
Fix for the issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/863

As part of https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/840, the internal implementation of the `IClientCertificate` interface's `publicCertificateHash()` method was changed to use SHA-256, and a new `publicCertificateHashSha1()` method was added to the internal implementation to handle scenarios that still needed SHA-1. 

However, this meant that anyone who implemented their own `IClientCredential` instead of using our `ClientCredentialFactory` would have a `publicCertificateHash()` method that still used SHA-1, causing issues because our library expected that method to produce a SHA-256 hash.

This PR solves that issue by adding a new `publicCertificateHash256` method to the public interface to handle SHA-256 scenarios, and once again has `publicCertificateHash` handling the SHA-1 scenarios. 

The new interface method is given a default implementation so that it doesn't break any customer's implementations, and if it's not implemented then the library falls back to `publicCertificateHash` and SHA-1. Anyone who was always using our `ClientCredentialFactory ` should not encounter any issues from either the original changes or the ones in this PR.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
@@ -101,12 +101,7 @@ public class ConfidentialClientApplication extends AbstractClientApplicationBase
         } else if (clientCredential instanceof ClientCertificate) {
             this.clientCertAuthentication = true;
             this.clientCertificate = (ClientCertificate) clientCredential;
-            if (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS) {
-                //When this was added, ADFS did not support SHA256 hashes for client certificates
-                clientAuthentication = buildValidClientCertificateAuthority(true);
-            } else  {
-                clientAuthentication = buildValidClientCertificateAuthority(false);
-            }
+            clientAuthentication = buildValidClientCertificateAuthority();
         } else if (clientCredential instanceof ClientAssertion) {
             clientAuthentication = createClientAuthFromClientAssertion((ClientAssertion) clientCredential);
         } else {
@@ -120,19 +115,17 @@ public class ConfidentialClientApplication extends AbstractClientApplicationBase
             final Date currentDateTime = new Date(System.currentTimeMillis());
             final Date expirationTime = ((PrivateKeyJWT) clientAuthentication).getJWTAuthenticationClaimsSet().getExpirationTime();
             if (expirationTime.before(currentDateTime)) {
-                if (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS) {
-                    clientAuthentication = buildValidClientCertificateAuthority(true);
-                } else  {
-                    clientAuthentication = buildValidClientCertificateAuthority(false);
-                }
+                clientAuthentication = buildValidClientCertificateAuthority();
             }
         }
         return clientAuthentication;
     }
 
-    //The library originally used SHA-1 for thumbprints as other algorithms were not supported server-side,
-    //  and while support for SHA-256 has been added certain flows still only allow SHA-1
-    private ClientAuthentication buildValidClientCertificateAuthority(boolean useSha1) {
+    private ClientAuthentication buildValidClientCertificateAuthority() {
+        //The library originally used SHA-1 for thumbprints as other algorithms were not supported server-side.
+        //When this was written support for SHA-256 had been added, however ADFS scenarios still only allowed SHA-1.
+        boolean useSha1 = Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS;
+
         ClientAssertion clientAssertion = JwtHelper.buildJwt(
                 clientId(),
                 clientCertificate,
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
@@ -102,9 +102,10 @@ public class ConfidentialClientApplication extends AbstractClientApplicationBase
             this.clientCertAuthentication = true;
             this.clientCertificate = (ClientCertificate) clientCredential;
             if (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS) {
-                clientAuthentication = buildValidClientCertificateAuthoritySha1();
+                //When this was added, ADFS did not support SHA256 hashes for client certificates
+                clientAuthentication = buildValidClientCertificateAuthority(true);
             } else  {
-                clientAuthentication = buildValidClientCertificateAuthority();
+                clientAuthentication = buildValidClientCertificateAuthority(false);
             }
         } else if (clientCredential instanceof ClientAssertion) {
             clientAuthentication = createClientAuthFromClientAssertion((ClientAssertion) clientCredential);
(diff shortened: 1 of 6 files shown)


msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClientCertificateTest.java
```diff
@@ -36,4 +41,35 @@ void testGetClient() {
         final ClientCertificate kc = ClientCertificate.create(key, null);
         assertNotNull(kc);
     }
+
+    @Test
+    void testIClientCertificateInterface_Sha256AndSha1() throws NoSuchAlgorithmException, CertificateException {
```
bgavrilMS: Please create a better E2E (not necessarily integration) test that uses MSAL public API. To assert, you can intercept the HTTP call to the /token endpoint, get the client_assertion parameter and observe that it has x5t#s256 which is correctly computed.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
```diff
@@ -97,4 +105,34 @@ static HttpResponse expectedResponse(int statusCode, String response) {
 
         return httpResponse;
     }
+
+    static void setPrivateKeyAndCert() {
```
bgavrilMS, line 140: Can you use a static certificate instead? Like in MISE? That way you can assert on the the well known hash too.

bgavrilMS (changes requested):
Insufficient testing.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java
Avery-Dunn: The variables are already static, or are you saying to just create one locally and add it to the repo?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -102,9 +102,10 @@ private void initClientAuthentication(IClientCredential clientCredential) {
             this.clientCertAuthentication = true;
             this.clientCertificate = (ClientCertificate) clientCredential;
             if (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS) {
-                clientAuthentication = buildValidClientCertificateAuthoritySha1();
+                //When this was added, ADFS did not support SHA256 hashes for client certificates
+                clientAuthentication = buildValidClientCertificateAuthority(true);
             } else  {
-                clientAuthentication = buildValidClientCertificateAuthority();
+                clientAuthentication = buildValidClientCertificateAuthority(false);
```
Robbie-Microsoft: You could eliminate the if statement and call `buildValidClientCertificateAuthority` only once:

```typescript
const useSha1: boolean = (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS);
clientAuthentication = buildValidClientCertificateAuthority(useSha1)
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -119,32 +120,25 @@ protected ClientAuthentication clientAuthentication() {
             final Date currentDateTime = new Date(System.currentTimeMillis());
             final Date expirationTime = ((PrivateKeyJWT) clientAuthentication).getJWTAuthenticationClaimsSet().getExpirationTime();
             if (expirationTime.before(currentDateTime)) {
-                //The asserted private jwt with the client certificate can expire so rebuild it when the
-                clientAuthentication = buildValidClientCertificateAuthority();
+                if (Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS) {
+                    clientAuthentication = buildValidClientCertificateAuthority(true);
+                } else  {
+                    clientAuthentication = buildValidClientCertificateAuthority(false);
+                }
```
Robbie-Microsoft: See above comment about eliminating if statement

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -119,32 +120,25 @@ protected ClientAuthentication clientAuthentication() {
...
-                clientId(),
-                clientCertificate,
-                this.authenticationAuthority.selfSignedJwtAudience(),
-                sendX5c,
-                false);
-        return createClientAuthFromClientAssertion(clientAssertion);
-    }
-
     //The library originally used SHA-1 for thumbprints as other algorithms were not supported server-side,
     //  and while support for SHA-256 has been added certain flows still only allow SHA-1
```
Robbie-Microsoft: super duper nit: add a `,` after `added`. In my mind, it makes sense to indicate a pause to break up the sentence - please feel free to ignore this if you want

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
```diff
@@ -56,11 +56,12 @@ static ClientAssertion buildJwt(String clientId, final ClientCertificate credent
                 builder.x509CertChain(certs);
             }
 
-            //SHA-256 is preferred, however certain flows still require SHA-1 due to what is supported server-side
-            if (useSha1) {
-                builder.x509CertThumbprint(new Base64URL(credential.publicCertificateHashSha1()));
+            //SHA-256 is preferred, however certain flows still require SHA-1 due to what is supported server-side. If SHA-256
+            // is not supported or the IClientCredential.publicCertificateHash256() method is not implemented, the library will default to SHA-1.
+            if (useSha1 || credential.publicCertificateHash256() == null) {
```
Robbie-Microsoft: You're using `credential.publicCertificateHash256()` twice - once in the if check and once in the else. Can you set it to a const before the if check?

(and add a type, below is typescript)
```typescript
const publicCertificateHash256: hash | null = credential.publicCertificateHash256();
```

Robbie-Microsoft:
What is the "default" hash algorithm that you're implementing in this PR? Sha1 or Sha256? Your description states (as I interpret it, based on the last paragraph) that Sha256 should be the default, but your code revolves around checks for `if Sha1, then do this, else Sha256` instead of `if Sha256, then do this, else Sha1`.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClientCertificateTest.java
Avery-Dunn: In the latest commit I added a more thorough test: everything up until the call to the token endpoint is real, and then the mocked response will only be successful if 'x5t#s256' is one of the headers in the assertion.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
Avery-Dunn: In the latest commit it's now only set once.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: In the latest commit I fixed the grammar and adjusted the wording a bit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: Good point, in the latest commit I moved this logic down into the `buildValidClientCertificateAuthority` method. Now that method does the ADFS authority check itself, with a comment explaining why it's done.

Avery-Dunn:
> What is the "default" hash algorithm that you're implementing in this PR? Sha1 or Sha256? Your description states (as I interpret it, based on the last paragraph) that Sha256 should be the default, but your code revolves around checks for `if Sha1, then do this, else Sha256` instead of `if Sha256, then do this, else Sha1`.

To clarify, 'default' here refers a default method in an interface. If you add a new method to a public interface, it needs a default implementation or else anything that implements the interface would also need to change to implement the new method (which would be a breaking change): https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

So technically the `publicCertificateHash()` method that returns a SHA-1 hash can be considered the 'default' hashing method, because the library will work fine even if the new `publicCertificateHash256()` API is not given a 'real' implementation and just returns null. However, the vast majority of customers will use the library's implementation of the `IClientCertificate` interface, which uses SHA-256 for everything except ADFS.

(as a bit of background, we found this issue because a customer was using the library's implementation of `IClientCertificate.publicCertificateHash()` in their own token calls. The library's implementation is package-private but the interface is public, so when that method started producing a SHA-256 hash instead of SHA-1 it was a breaking change for their very niche use case)

bgavrilMS (approved):

PR #722, opened 2023-10-05
Static token cache for MSI

Avery-Dunn:
Adds a static token cache field to ManagedIdentityApplication so that the cache is shared among all ManagedIdentityApplication instances, as well as a TokenSource and AuthenticationResultMetadata classes to provide information about where tokens are coming from.

This is similar to what is being done in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/708, however this is a simpler implementation to help ensure the behavior is correct for upcoming beta testing.

PR #901, opened 2025-01-17
Add CacheRefreshReason to telemetry and metadata

Avery-Dunn:
This PR finishes the work discussed for https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/697. Token source metadata was added in PR https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/722 and merged in as part of the initial MSI support, however PR https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/708 was made in parallel with some additional changes that were never merged in:
- Add a public enum CacheRefreshReason, to provide the refresh reason in a way that's consistent with MSAL .NET's [CacheRefreshReason](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs)
- Update AuthenticationResultMetadata to include the reason a token was refreshed, along with the existing data for the source of the token
- Clean up the logic around setting the refresh reason in server-side telemetry

There are unit tests to cover various refresh reasons, which also fills in a testing gap for token expiry behavior (mainly around the 5-minute buffer for refreshing tokens).


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultMetadata.java
```diff
@@ -20,6 +20,19 @@
...
+     */
     private TokenSource tokenSource;
+
+    /**
+     * When the token should be proactively refreshed. May be null or 0 if proactive refresh is not used
+     */
     private Long refreshOn;
+
+    /**
+     * Specifies the reason for refreshing a token, see {@link CacheRefreshReason} for possible values.Will be {@link CacheRefreshReason#NOT_APPLICABLE} if the token was not refreshed
```
bgavrilMS:
```suggestion
     * Specifies the reason for refreshing the access token, see {@link CacheRefreshReason} for possible values. Will be {@link CacheRefreshReason#NOT_APPLICABLE} if the token was returned from the cache or if the API used to fetch the token does not attempt to read the cache.
```

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -116,6 +119,79 @@ void confidentialAppAcquireTokenSilently_claimsSkipCache() throws Throwable {
...
     }
 
+    @Test
+    void testTokenRefreshReasons() throws Exception {
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder("clientId", ClientCredentialFactory.createFromSecret("password"))
+                        .authority("https://login.microsoftonline.com/tenant/")
+                        .instanceDiscovery(false)
```
bgavrilMS, line 129: nit: won't block the PR for this, but in general you do not want to over-customize MSAL. The Http mocking can reply with instance discovery json.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -116,6 +119,79 @@ void confidentialAppAcquireTokenSilently_claimsSkipCache() throws Throwable {
...
+                        .validateAuthority(false)
+                        .httpClient(httpClientMock)
+                        .build();
+
+        HashMap<String, String> responseParameters = new HashMap<>();
+
+        //Acquire a token that expired an hour ago
+        responseParameters.put("access_token", "expiredToken");
+        responseParameters.put("id_token", TestHelper.createIdToken(new HashMap<>()));
```
bgavrilMS, line 138: Client Credentails doesn't respond with id token.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -116,6 +119,79 @@ void confidentialAppAcquireTokenSilently_claimsSkipCache() throws Throwable {
...
+        responseParameters.put("expires_in", "-3600");
```
bgavrilMS: nit: Every test out there should be mocking the HTTP req / res. I believe you alraedy built some helpers for this? Can you not reuse the code ?

bgavrilMS: I'd argue that MSAL should never accept to put a token with expires_in in the cache...

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
```diff
@@ -0,0 +1,36 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * Specifies the reason for fetching the access token from the identity provider when using {@link AbstractClientApplicationBase#acquireTokenSilently(SilentParameters)}
+ */
+public enum CacheRefreshReason {
```
bgavrilMS, line 9: What about claims?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -113,15 +119,15 @@ private boolean shouldRefresh(SilentParameters parameters, AuthenticationResult
...
+            setCacheTelemetry(CacheRefreshReason.FORCE_REFRESH);
             log.debug("Refreshing access token because forceRefresh parameter is true.");
             return true;
         }
 
         //If the request contains claims then the token should be refreshed, to ensure that the returned token has the correct claims
         //  Note: these are the types of claims found in (for example) a claims challenge, and do not include client capabilities
         if (parameters.claims() != null) {
-            setCacheTelemetry(CacheTelemetry.REFRESH_FORCE_REFRESH.telemetryValue);
+            setCacheTelemetry(CacheRefreshReason.FORCE_REFRESH);
```
bgavrilMS: Can we have a separate entry for Claims please?

bgavrilMS (changes requested):
Let's please get a Claims entry as reason.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
Avery-Dunn: Should there actually be a separate enum entry, or should it be like MSAL .NET and have the name reference both force refresh and claims? https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7d17705533eab68d504ee2f179364922367dfba7/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs#L18

It makes sense for them to be separate, however I think both of those reasons map to the same telemetry value.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
Avery-Dunn: In the latest commit I renamed the `FORCE_REFRESH` reason to `FORCE_REFRESH_OR_CLAIMS`, to better match MSAL .NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/7d17705533eab68d504ee2f179364922367dfba7/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs#L18

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
Avery-Dunn: Writing that test case definitely felt wrong, and I agree the library shouldn't allow it. For the most part we just extract various values from the token response, and don't check whether or not the values make sense: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java

I saw that MSAL .NET does do some checks and other MSALs probably do some as well, so I'll create an SDK consistency issue to add in some validation.

Avery-Dunn: There is a helper and these tests are using it: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/15da510dc38a61472f36717785e9e95effe19ff2/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/TestHelper.java#L79

I suppose I could make additional helpers for various specific scenarios (like `getExpiredToken()`, `getNearlyExpiredToken()`, `getValidToken()`, etc.), but to me it seems better to let tests adjust the few parameters that are important for the test rather than have a separate helper method for each scenario.

However, there wasn't an existing helper for creating the requests, the mock just returned the same response for any HttpRequest object. In the latest commit I added `TestHelper.createTokenRequestMock()` to contain the logic for creating the mocked token requests/responses, which should make the tests easier to read and better control the most common mock.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
Avery-Dunn: Good point, it's valid as far the library is concerned but probably not a good practice. In the latest commit this test uses the OBO flow, which is more appropriate for a refresh test.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
Avery-Dunn: Since these unit tests weren't covering anything related to instance discovery I figured it'd be easier to just disable these extra HTTP calls, but I agree that's bad practice. I'll create an issue to improve the HTTP mocking since it is such a core part of the unit tests.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
bgavrilMS: Let's pls do the right thing from the start. Separate Force_Refresh and Claims. We need to do this in .NET too, please open a bug there.

Avery-Dunn: In the latest commit these are now separate entries in the enum.

Do they also have separate telemetry values? Or is this more for developer clarity about why the token was refreshed?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
```diff
@@ -116,6 +118,81 @@ void confidentialAppAcquireTokenSilently_claimsSkipCache() throws Throwable {
...
+        responseParameters.put("access_token", "claimsToken");
+        TestHelper.createTokenRequestMock(httpClientMock, TestHelper.getSuccessfulTokenResponse(responseParameters), 200);
+
+        silentParameters = SilentParameters.builder(Collections.singleton("someScopes"), result.account()).claims(new ClaimsRequest()).build();
+        result = cca.acquireTokenSilently(silentParameters).get();
+
+        assertRefreshedToken(result, "claimsToken", CacheRefreshReason.CLAIMS, cca.tokenCache.accessTokens.size());
+    }
+
+    //Asserts that there is one expected token in the cache, and that it was refreshed with the expected reason
```
gladjohn, line 206: can you also add a test for proactive token renewal?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
```diff
@@ -0,0 +1,40 @@
...
+    /**
+     * Token did not need to be refreshed, or was retrieved in a non-silent call
+     */
+    NOT_APPLICABLE(0),
+    /**
+     * Silent call was made with the force refresh option
+     */
+    FORCE_REFRESH(1),
```
gladjohn, line 17: If telemetry logs 1, how do we distinguish between FORCE_REFRESH and CLAIMS

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenSilentSupplier.java
```diff
@@ -113,15 +119,15 @@ private boolean shouldRefresh(SilentParameters parameters, AuthenticationResult
...
             log.debug("Refreshing access token because forceRefresh parameter is true.");
             return true;
         }
 
         //If the request contains claims then the token should be refreshed, to ensure that the returned token has the correct claims
         //  Note: these are the types of claims found in (for example) a claims challenge, and do not include client capabilities
         if (parameters.claims() != null) {
-            setCacheTelemetry(CacheTelemetry.REFRESH_FORCE_REFRESH.telemetryValue);
+            setCacheTelemetry(CacheRefreshReason.CLAIMS);
             log.debug("Refreshing access token because the claims parameter is not null.");
```
gladjohn, line 131:
```suggestion
            log.debug("Refreshing access token because the claims parameter is not null.");
```
log.debug("Refreshing access token due to reason: " + CacheRefreshReason.CLAIMS);

gladjohn:
```suggestion
            log.debug("Refreshing access token due to reason: " + CacheRefreshReason.CLAIMS);
```

Avery-Dunn: In the latest commit I improved these logs, now they use proper String formatting and include the enum values like you suggested.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CacheRefreshReason.java
Avery-Dunn: FORCE_REFRESH and CLAIMS are mapped to the same telemetry value in .NET and (I think) other MSALs: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/Cache/CacheRefreshReason.cs

That is also the enum that [the telemetry schema docs](https://eng.ms/docs/microsoft-security/identity/entra-developer-application-platform/id4s-identity-for-services/authn-middleware-sdk-microsoft-identity-service-essentials/microsoft-identity-service-essentials/articles/using-mise/telemetry-schema) point to as an example of valid values. 

However, in a different comment Bogdan suggested splitting them in order to make the code/logs more clear.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AcquireTokenSilentlyTest.java
Avery-Dunn: Added in the latest commit.

bgavrilMS (approved):

gladjohn (approved):

gladjohn:
let's you merge with open comments?

PR #902, opened 2025-01-21
Add unit tests for refresh buffer

Avery-Dunn:
Fills in a testing gap related to the 5-minute buffer for refreshing tokens. These two new tests show that expired tokens are refreshed if past their expiration time, and that this behavior is the same if the expiration time is less than 5 minutes away.


Avery-Dunn:
Closing, it makes more sense just to add these tests to https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/901

Issue #899, opened 2025-01-14
[Engineering task] Fix codeql issues


neha-bhargava:
https://liquid.microsoft.com/codeql/issues/77d1b4fa-c17d-4ec0-88b0-271ecd626c71?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058
https://liquid.microsoft.com/codeql/issues/853058b9-6ee1-4faf-997c-fd0aae5665a8?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058
https://liquid.microsoft.com/codeql/issues/624d83c5-820a-40aa-baa8-f85cc37d9d4f?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058
https://liquid.microsoft.com/codeql/issues/186ffe1e-8d20-4409-bff3-93e5c16146dc?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058
https://liquid.microsoft.com/codeql/issues/1df1f695-41e1-4845-a9fe-de6e154adb89?copilot_promptid=E91B0CE9-0C1B-4AC2-8A46-33F49B67E058
https://liquid.microsoft.com/codeql/issues/b93ffa6c-c2ec-4131-b528-3c40749bebaf

PR #904, opened 2025-02-05
Add CodeQL suppressions for false positives

Avery-Dunn:
Fixes the CodeQL issues from https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/899:
- Three issues in the samples folder: suppressed since they are only meant to help manual testing during development, and the code is not included in any released package
- Two issues `AcquireTokenByInteractiveFlowSupplier`: the issue was String concatenation with user input, however the URL is validated earlier in the interactive flow
- One issue in `DefaultHttpClientManagedIdentity`: the issue was setting a `HostnameVerifier` that accepts all host names, however the `TrustManager` that uses the verifier only accepts a specific certificate thumbprint (this code was copied from the Azure Identity Java SDK and I discussed this behavior with them)

### DIFF
--- a/msal4j-sdk/src/samples/msal-web-sample/src/main/java/com/microsoft/azure/msalwebsample/CookieHelper.java
+++ b/msal4j-sdk/src/samples/msal-web-sample/src/main/java/com/microsoft/azure/msalwebsample/CookieHelper.java
@@ -32,6 +32,7 @@ public class CookieHelper {
         Cookie stateCookie = new Cookie(MSAL_WEB_APP_STATE_COOKIE, "");
         stateCookie.setMaxAge(0);
 
+        // CodeQL [java/insecure-cookie]: Suppressing CodeQL warning since this is just a sample
         httpResponse.addCookie(stateCookie);
 
         Cookie nonceCookie = new Cookie(MSAL_WEB_APP_NONCE_COOKIE, "");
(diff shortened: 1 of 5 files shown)


bgavrilMS (approved):

msal4j-sdk/src/samples/msal-web-sample/src/main/java/com/microsoft/azure/msalwebsample/CookieHelper.java
```diff
@@ -32,6 +32,7 @@ static void removeStateNonceCookies(HttpServletResponse httpResponse){
         Cookie stateCookie = new Cookie(MSAL_WEB_APP_STATE_COOKIE, "");
         stateCookie.setMaxAge(0);
 
+        // CodeQL [java/insecure-cookie]: Suppressing CodeQL warning since this is just a sample
```
bgavrilMS, line 35: @Avery-Dunn - can we revisit this? Our samples should not use insecure practices.

Avery-Dunn: This is just a simpler version of the "real" sample meant to help with manual E2E testing rather than teach something, however it turns out the same code is in the real sample: https://github.com/Azure-Samples/ms-identity-msal-java-samples/blob/main/1-server-side/msal-b2c-web-sample/src/main/java/com/microsoft/azure/msalwebsample/CookieHelper.java#L30

I've just created an issue about auditing the samples to find these sorts of insecure practices: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/911

Issue #909, opened 2025-02-12
Reduce usage of external dependencies


Avery-Dunn:
The Azure Identity SDK project is aiming to minimize the number of external dependencies, as they can lead to security risks and compatibility issues  for customers.

The specific ask for MSAL Java is to evaluate and remove these dependencies:
- [x] com.nimbusds.oauth-oidc-sdk
  - Only used for some very specific OAuth behaviors. Risk of regression and must be thoroughly tested
  - https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/889
- [x] net.minidev.json-smart
  - Azure SDK team suggested an alternative package for JSON parsing: [azure-json](https://github.com/Azure/azure-sdk-for-java/tree/main/sdk/serialization/azure-json)
  - Unfortunately the main places we use this are due to it being used in com.nimbusds.oauth-oidc-sdk, so this cannot be easily removed without also removing com.nimbusds.oauth-oidc-sdk
- [ ] org.slf4j.slf4j-api & org.slf4j.slf4j-simple
  - Azure SDK team suggested reflective approach to loading the SLF4j types if they are on the class path, example [here](https://github.com/billwert/azure-sdk-for-java/blob/eg-new-shared-events-package/sdk/clientcore/core/src/main/java/io/clientcore/core/instrumentation/logging/ClientLogger.java#L50)
- [x] org.projectlombok.lombok
  - Can simply be replaced by the same code it generates
- [x] com.fasterxml.jackson.core
  - https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/441
  - Available in azure-core https://mvnrepository.com/artifact/com.azure/azure-core
  - Can be replaced by https://mvnrepository.com/artifact/com.azure/azure-json

Avery-Dunn:
Done as of version 1.23.0: https://github.com/AzureAD/microsoft-authentication-library-for-java/releases/tag/v1.23.0

PR #918, opened 2025-03-04
Replace jackson-databind with azure-json in HTTP response classes

Avery-Dunn:
This PR begins the process of replacing the com.fasterxml.jackson.core dependency with azure-json, #909

There are a lot of added and deleted lines, but for the most part the changes are:
- Adjust classes to implement azure-json's [JsonSerializable](https://learn.microsoft.com/en-us/java/api/com.azure.json.jsonserializable?view=azure-java-stable) interface
- Remove jackson-databind's JSON annotations from those classes
- Add a new method to JsonHelper which converts a JSON string to a class which implements the JsonSerializable interface

To help keep the PR focused and lower the risk of a breaking change in behavior, this first PR covers various package-private classes and methods. Follow-up PRs will finish the job of removing com.fasterxml.jackson.core, as well as use RevAPI to ensure no changes are made to the API

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
@@ -6,6 +6,7 @@ package com.microsoft.aad.msal4j;
 import com.azure.json.JsonProviders;
 import com.azure.json.JsonReader;
 import com.azure.json.JsonSerializable;
+import com.azure.json.ReadValueCallback;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationFeature;
@@ -40,9 +41,9 @@ class JsonHelper {
     }
 
     //This method is used to convert a JSON string to an object which implements the JsonSerializable interface from com.azure.json
-    static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObject(String jsonResponse, Class<T> tClass) {
+    static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObject(String jsonResponse, ReadValueCallback<JsonReader, T> readFunction) {
         try (JsonReader jsonReader = JsonProviders.createReader(jsonResponse)) {
-            return (T) tClass.getDeclaredMethod("fromJson", JsonReader.class).invoke(null, jsonReader);
+            return readFunction.read(jsonReader);
         } catch (IOException e) {
             throw new MsalClientException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
         } catch (Exception e) {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
@@ -3,6 +3,9 @@
 
 package com.microsoft.aad.msal4j;
 
+import com.azure.json.JsonProviders;
+import com.azure.json.JsonReader;
+import com.azure.json.JsonSerializable;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonParser;
 import com.fasterxml.jackson.databind.DeserializationFeature;
@@ -33,6 +36,17 @@ class JsonHelper {
             return mapper.readValue(json, tClass);
         } catch (IOException e) {
             throw new MsalClientException(e);
+        }
+    }
+
+    //This method is used to convert a JSON string to an object which implements the JsonSerializable interface from com.azure.json
+    static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObject(String jsonResponse, Class<T> tClass) {
+        try (JsonReader jsonReader = JsonProviders.createReader(jsonResponse)) {
+            return (T) tClass.getDeclaredMethod("fromJson", JsonReader.class).invoke(null, jsonReader);
+        } catch (IOException e) {
+            throw new MsalClientException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
+        } catch (Exception e) {
+            throw new MsalClientException("Error parsing JSON response: " + e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
         }
     }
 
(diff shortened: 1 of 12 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -36,6 +39,17 @@ static <T> T convertJsonToObject(final String json, final Class<T> tClass) {
         }
     }
 
+    //This method is used to convert a JSON string to an object which implements the JsonSerializable interface from com.azure.json
+    static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObject(String jsonResponse, Class<T> tClass) {
+        try (JsonReader jsonReader = JsonProviders.createReader(jsonResponse)) {
+            return (T) tClass.getDeclaredMethod("fromJson", JsonReader.class).invoke(null, jsonReader);
```
bgavrilMS: Is all this reflection ok for apps consuming our libs? In .NET at least, reflection is frawned upon because it stops the AOT linker from doing its job. It's also slow.

 JSON libraries prefer to use a "source generation" approach which avoids reflection and is faster.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
```diff
@@ -3,42 +3,67 @@
...
-     * @param token the token string.
-     * @param expiresOn the expiration time.
-     */
-    @JsonCreator
-    private ManagedIdentityResponse(
-            @JsonProperty(value = "access_token") String token,
-            @JsonProperty(value = "expires_on") String expiresOn) {
-        this.accessToken = token;
-        this.expiresOn =  expiresOn;
+    public static ManagedIdentityResponse fromJson(JsonReader jsonReader) throws IOException {
```
bgavrilMS, line 27: Jackson seems more elegant / readable with the attribute approach...

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
Avery-Dunn: It's definitely more elegant, there's a reason it's by far the most popular JSON-related library on Maven: https://mvnrepository.com/open-source/json-libraries

But to get rid of 3rd-party dependencies there's not much choice, azure-json is at least used by most other Azure products: https://mvnrepository.com/artifact/com.azure/azure-json/usages

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: The first version I had simply had the same try/catch clause in a lot of places, and to avoid duplicate code I tried this approach.

However, ease of debugging is a concern (that's part of the reason we're even doing this), and since this JSON parsing could happen on every request it might slow them down. I'll do some more research/testing to see if there's a better approach (sending a lambda?), and if not I'll go back to directly calling that "fromJson" method in various places.

alzimmermsft: You should be able to remove reflection by making a few changes here:

1. Change the method parameters to:

```java
static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObject(String jsonResponse, ReadValueCallback<JsonReader, T> readFunction)
```

You know at calling time what `T` will be.

2. Change the try-with-resources block to:

```java
try (JsonReader jsonReader = JsonProviders.createReader(jsonResponse)) {
    return readFunction.read(jsonReader);
}
```

3. Change the caller to:

```java
JsonHelper.convertJsonStringToJsonSerializableObject(instanceDiscoveryJson, AadInstanceDiscoveryResponse::fromJson);
```

```java
JsonHelper.convertJsonStringToJsonSerializableObject(response.body(), ManagedIdentityResponse::fromJson);
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
alzimmermsft: Jackson Databind is much cleaner, and in some cases clearer, but it does require a lot of reflection to be used by Jackson. This code here is effectively equivalent to what you'd see if you were using Jackson Core which doesn't have all the reflection handling.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/UserDiscoveryResponse.java
```diff
@@ -39,4 +33,50 @@ boolean isAccountManaged() {
...
+
+    public static UserDiscoveryResponse fromJson(JsonReader jsonReader) throws IOException {
+        UserDiscoveryResponse response = new UserDiscoveryResponse();
+        return jsonReader.readObject(reader -> {
+            while (reader.nextToken() != JsonToken.END_OBJECT) {
+                String fieldName = reader.getFieldName();
+                reader.nextToken();
+                switch (fieldName) {
+                    case "ver":
+                        response.version = Float.parseFloat(reader.getString());
```
alzimmermsft, line 45: `JsonReader` has `getFloat`, I'd recommend using that instead. Unless you need special handling if `ver` is actually a String JSON node.

Avery-Dunn: I originally used `getFloat`, however it gave me the following error: `Current token (VALUE_STRING) not numeric, can not use numeric value accessors at [Source: (String)"{"ver":"1.0", ...`

I assumed it was because the JSON we were receiving was a string of "1.0", though at least according to the specs shouldn't it be able to parse a number or a string? https://learn.microsoft.com/en-us/java/api/com.azure.json.jsonreader?view=azure-java-stable#com-azure-json-jsonreader-getfloat()

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: Thanks for the suggestion, I made these changes in the latest commit.

Definitely easier to read, and just from a couple quick tests runs about twice as fast (0-2 milliseconds instead of 2-4)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/UserDiscoveryResponse.java
alzimmermsft: `JsonReader` uses a very literal interpretation of the JSON spec where it won't perform implicit conversions to numbers or booleans if the underlying value is a JSON string.

PR #926, opened 2025-03-27
Remove usage of com.nimbusds.oauth2 from grant-related classes

Avery-Dunn:
This PR removes the com.nimbusds.oauth2 dependency from various classes that deal with different grant types, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909

The dependency was mainly used by these classes to create maps of different parameters needed for each OAuth grant type, so the changes are mostly the same for each file:
- Remove `com.nimbusds.oauth2.sdk.AuthorizationGrant` and various similar imports
- Replace usages of Nimbus's `AuthorizationGrant` subclasses with our existing `OAuthAuthorizationGrant` class
- Replace Nimbus's `AuthorizationGrant.toParameters()` method with our own `Map<String, List<String>>` creation
- Add a new `GrantConstants` file for standardized OAuth parameter names and values
- Refactor older code and remove unused code


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -15,74 +13,54 @@
...
-    String addCommonScopes(String scopes) {
-        Set<String> allScopes = new HashSet<>(
-                Arrays.asList(COMMON_SCOPES_PARAM.split(SCOPES_DELIMITER)));
-
-        if (!StringHelper.isBlank(scopes)) {
-            allScopes.addAll(Arrays.asList(scopes.split(SCOPES_DELIMITER)));
-        }
-        return String.join(SCOPES_DELIMITER, allScopes);
+    OAuthAuthorizationGrant(Map<String, List<String>> params, Set<String> scopesSet, ClaimsRequest claims) {
+        this(params, scopesSet != null ? String.join(" ", scopesSet) : null, claims);
```
bgavrilMS: question: Is the idea here that this constructor will have typed parameters for all "common" parameters and a property bag for the grant specified parameters? 

suggestion: add a comment to the ctor for this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
```diff
@@ -77,55 +77,56 @@ private boolean IsUiRequiredCacheSupported() {
     private OAuthAuthorizationGrant processPasswordGrant(
             OAuthAuthorizationGrant authGrant) throws Exception {
 
-        if (!(authGrant.getAuthorizationGrant() instanceof ResourceOwnerPasswordCredentialsGrant)) {
+        if (!(authGrant.getParameters().get(GrantConstants.GRANT_TYPE_PARAMETER).get(0).equals(GrantConstants.PASSWORD))) {
```
bgavrilMS: issue: see my comments below - the grant should be a strongly typed parameter. It will make the code easier to read and more performant.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java
```diff
@@ -30,20 +29,21 @@ class OnBehalfOfRequest extends MsalRequest {
...
         }
 
-        return new OAuthAuthorizationGrant(jWTBearerGrant, String.join(SCOPES_DELIMITER, parameters.scopes()), params);
+        try {
+            params.put(GrantConstants.ASSERTION_PARAMETER, Collections.singletonList(SignedJWT.parse(parameters.userAssertion().getAssertion()).getParsedString()));
+        } catch (ParseException e) {
+            throw new RuntimeException(e);
+        }
+
+        return new OAuthAuthorizationGrant(params, String.join(SCOPES_DELIMITER, parameters.scopes()), null);
```
bgavrilMS: issue:  that the second parameter of that ctor is a List<string>. but here you add a string (which becomes a list of char?)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java
```diff
@@ -30,20 +29,21 @@ class OnBehalfOfRequest extends MsalRequest {
...
+        params.put(GrantConstants.GRANT_TYPE_PARAMETER, Collections.singletonList(GrantConstants.JWT_BEARER));
         params.put("requested_token_use", Collections.singletonList("on_behalf_of"));
+
         if (parameters.claims() != null) {
             params.put("claims", Collections.singletonList(parameters.claims().formatAsJSONString()));
         }
 
-        return new OAuthAuthorizationGrant(jWTBearerGrant, String.join(SCOPES_DELIMITER, parameters.scopes()), params);
+        try {
+            params.put(GrantConstants.ASSERTION_PARAMETER, Collections.singletonList(SignedJWT.parse(parameters.userAssertion().getAssertion()).getParsedString()));
```
bgavrilMS: issue: We should never parse tokens. Can we get a bug to remove this?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/RefreshTokenRequest.java
```diff
@@ -31,11 +32,13 @@ class RefreshTokenRequest extends MsalRequest {
         this.parentSilentRequest = silentRequest;
     }
 
-    private static AbstractMsalAuthorizationGrant createAuthenticationGrant(
-            RefreshTokenParameters parameters) {
+    private static AbstractMsalAuthorizationGrant createAuthenticationGrant(RefreshTokenParameters parameters) {
+        Map<String, List<String>> params = new LinkedHashMap<>();
+
+        params.put(GrantConstants.GRANT_TYPE_PARAMETER, Collections.singletonList(GrantConstants.REFRESH_TOKEN));
+        params.put("refresh_token", Collections.singletonList(parameters.refreshToken()));
```
bgavrilMS, line 39: TODO: move "refresh_token" to constants since everything else is a constant

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -15,74 +13,54 @@
...
-
-    String addCommonScopes(String scopes) {
-        Set<String> allScopes = new HashSet<>(
-                Arrays.asList(COMMON_SCOPES_PARAM.split(SCOPES_DELIMITER)));
-
-        if (!StringHelper.isBlank(scopes)) {
-            allScopes.addAll(Arrays.asList(scopes.split(SCOPES_DELIMITER)));
-        }
-        return String.join(SCOPES_DELIMITER, allScopes);
+    OAuthAuthorizationGrant(Map<String, List<String>> params, Set<String> scopesSet, ClaimsRequest claims) {
```
bgavrilMS: issue: I haven't seen a single instance where the param value is a list with more than 1 value. They are all 1 values i.e. String? This creates unnecessary allocations of List

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -15,74 +13,54 @@
 
 class OAuthAuthorizationGrant extends AbstractMsalAuthorizationGrant {
```
bgavrilMS, line 13: q: What is the point of the abstract class?

bgavrilMS, line 13: q: who adds client_info = 1 ? Do we have E2E tests that will fail if client_info=1 isn't added ?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -15,74 +13,54 @@
...
```
bgavrilMS, line 14: q: do we have test coverage for extra query paramters?

bgavrilMS (changes requested):
See comments marked with issue

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
Avery-Dunn: It's existed since before I joined so I don't know the actual history of it. However, it appears to be copying the style of the `com.nimbusds.oauth2.sdk.AuthorizationGrant`, it specifies that the grant needs a `toParameters()` method and adds some MSAL-specific behavior (mainly just scopes?)

There are a few other classes that extend it, for example `DeviceCodeAuthorizationGrant` it has its own `toParameters()` that creates the parameters map in the same way that I did for the files in this PR: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DeviceCodeAuthorizationGrant.java

Avery-Dunn: It's added in the `toParameters()` method: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/a990b4a75981d4ebd5351e978df4fb3a4c547b41/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java#L72

Looks like that line hasn't been changed since mid-2019, and other classes that extend `AbstractMsalAuthorizationGrant` do the same.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
Avery-Dunn: There are a few tests that use junk extra query parameters, but I don't think any test uses 'real' parameters that would be evaluated in an E2E. Is there an example of that sort of test in .NET?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
Avery-Dunn: In the latest commit I refactored this class's constructors and added a comment to clarify:
- `params` are grant-specific parameters, such as the specific value of `grant_type` or the `username` parameter only used in ROPC
- `scopes` are the scopes they want to request, and are added to a default set of common scopes
- `claims` are just all the additional claims that might be needed

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java
Avery-Dunn: In the latest commit I reduced the constructors in `OAuthAuthorizationGrant` to just one, and refactored the code in several classes so that they aren't swapping back and forth between sets, lists, strings, etc.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OnBehalfOfRequest.java
Avery-Dunn: Removed in the latest commit.

When it was using the dependency's `JWTBearerGrant` class the constructor needed a `SignedJWT` object created from the assertion string, and I just copied that without thinking that the assertion string was all we needed.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
Avery-Dunn: All of them at least have `grant_type`, and some have a few more. For example, `AuthorizationCodeRequest` also has `code`, `redirect_uri`, and `code_verifier`

However, I agree that it's for the most part an unnecessary. I believe the reason for it is because a lot of Nimbus's APIs use `Map<String, List<String>>` for parameters, even though the `List` often has just one element

The next PR will remove the dependency from the classes that handle the HTTP calls where these parameters maps are actually used, and that will include refactoring to simplify these maps

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: What other comment are you referring to? Also, by "strongly typed parameter" do you mean:
- An enum with all the grant types?
- A field in OAuthAuthorizationGrant (so it'd be just `authGrant.grantType()`)?
- Specific classes for each grant, like the `ResourceOwnerPasswordCredentialsGrant` that used to be on this line?

billwert: Agreed with @bgavrilMS. `getParametrs().get(...).get(N)` is a confusing pattern to keep reading. `OAuthAuthorizationGrant` should get helper methods for common things. Maybe you need this path for something user-defined (can such a thing exist?) or something, but the common cases should be obvious.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
```diff
@@ -15,74 +12,43 @@
...
-    }
-
-    OAuthAuthorizationGrant(AuthorizationGrant grant, Map<String, List<String>> params) {
-        this.grant = grant;
-
-        if (params != null) {
-            this.params.putAll(params);
+            this.params.put("claims", Collections.singletonList(claims.formatAsJSONString()));
         }
     }
```
billwert: It seems like nothing can modify values in this class after construction. This doesn't need to reconstruct the output map every time. Setup the extra values into the `params` field at construction.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
```diff
@@ -77,55 +77,56 @@ private boolean IsUiRequiredCacheSupported() {
...
                     msalRequest.requestContext(),
                     this.clientApplication.serviceBundle(),
                     this.clientApplication.logPii());
 
-            AuthorizationGrant updatedGrant = getSAMLAuthorizationGrant(response);
+            Map<String, List<String>> params = getSAMLAuthGrantParameters(response);
+            params.putAll(authGrant.toParameters());
 
-            authGrant = new OAuthAuthorizationGrant(updatedGrant, authGrant.getParameters());
+            authGrant = new OAuthAuthorizationGrant(params, null, null);
```
billwert: Maybe some constructor overloads if `null, null` is a commonish case.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
```diff
@@ -77,55 +77,56 @@ private boolean IsUiRequiredCacheSupported() {
...
         }
         return authGrant;
     }
 
-    private AuthorizationGrant getSAMLAuthorizationGrant(WSTrustResponse response) throws UnsupportedEncodingException {
-        AuthorizationGrant updatedGrant;
+    private Map<String, List<String>> getSAMLAuthGrantParameters(WSTrustResponse response) {
```
billwert, line 104: Now maybe I see why you went with the "expose a map everywhere" strategy. Maybe OAuthGrantParameters is too specific a case, and we need a more generic "paramter bag" type?

billwert (changes requested):
High level looks good. Two themes:
1) Replace `Map<String, List<String>>` as the representation of params with some type with good helpers. 
2) `OAuthAuthorizationGrant` seems immutable, so it doesn't need to recreate the params and add claims etc every time. Just do it at construction.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
bgavrilMS: No, and we don't need an E2E test. We just need to ensure that the extra query parameters are appended to the URLs that MSAL hits, in particular the token endpoint.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationCodeRequest.java
```diff
@@ -17,19 +17,19 @@ class AuthorizationCodeRequest extends MsalRequest {
...
         if (parameters.codeVerifier() != null) {
-            authorizationGrant = new AuthorizationCodeGrant(
-                    new AuthorizationCode(parameters.authorizationCode()),
-                    parameters.redirectUri(),
-                    new CodeVerifier(parameters.codeVerifier()));
-
-        } else {
-            authorizationGrant = new AuthorizationCodeGrant(
-                    new AuthorizationCode(parameters.authorizationCode()), parameters.redirectUri());
+            params.put("code_verifier", Collections.singletonList(parameters.codeVerifier()));
```
bgavrilMS, line 30: I am not sure how much of PKCE was handled by nimbus, so please make sure we have enough tests that validate that auth_code uses PKCE. Both desktop and web app scenarios.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: Yeah, pretty much every nimbus API used `Map<String, List<String>>` for parameters. The next PR will replace the final util imports that still use that map format and use something simpler, this PR was trying to just do a 1:1 replacement to lower the risk of behavioral changes.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: This spot and one other in a test were the only places that used a double null, however another also had a null for the third parameter (claims) but not the second (scopes).

In the latest commit I added a constructor that omits the claims parameter, as that's the one that's most likely to be null in a flow.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthAuthorizationGrant.java
Avery-Dunn: Good point, in the latest commit the "client_info" parameter is added to the map by the constructor, and the toParameters() method simply returns an unmodifiable version of the existing map.

changelog.txt
```diff
@@ -1,3 +1,11 @@
+Version 1.20.0
```
bgavrilMS, line 5: bad merge? These are not part of this PR?

Avery-Dunn: Not a bad merge, just a lack of a one: this PR is pointing to the "nimbus-removal" feature branch, and two other PRs got merged into that feature branch before this PR.

I just pulled current feature branch into this PR's branch, and now it seems to have the correct file list again.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: In the latest commit there is now a helper method in `OAuthAuthorizationGrant` to get specific values for keys defined in `GrantConstants`.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
```diff
@@ -39,8 +39,7 @@ AuthenticationResult execute() throws Exception {
         }
 
         if (authGrant instanceof OAuthAuthorizationGrant) {
-            msalRequest.msalAuthorizationGrant =
-                    processPasswordGrant((OAuthAuthorizationGrant) authGrant);
+            processPasswordGrant((OAuthAuthorizationGrant) authGrant);
```
bgavrilMS, line 42: It'a bit strange that the authorization grant needs to know about ROPC. Seems like a violation of Single Responsability Principle?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
```diff
@@ -74,58 +73,52 @@ private boolean IsUiRequiredCacheSupported() {
...
-            OAuthAuthorizationGrant authGrant) throws Exception {
-
-        if (!(authGrant.getAuthorizationGrant() instanceof ResourceOwnerPasswordCredentialsGrant)) {
-            return authGrant;
-        }
+    private void processPasswordGrant(OAuthAuthorizationGrant authGrant) throws Exception {
 
-        if (msalRequest.application().authenticationAuthority.authorityType != AuthorityType.AAD) {
-            return authGrant;
+        //Additional processing is only needed if it's a password grant with a non-AAD authority
```
bgavrilMS: User Real discovery needs to be done only for AAD authority. For non-AAD authority, perform the basic grant only.

bgavrilMS: I think you got the condition correct, but the comment wrong?

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: It's bad practice, though it matches the old behavior. I tried to keep the changes simple to make the PRs easier to review, however it'd definitely be better for various grants to contain their own specific behavior.

I've just created an issue to track this and improve the structure of grant related classes: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/944

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByAuthorizationGrantSupplier.java
Avery-Dunn: Yep, the if statement correctly exits the method if it's not-AAD, however when writing the comment I just said "non-AAD" because of the `!= AuthorityType.AAD` part

Comment fixed in the latest commit

PR #927, opened 2025-04-01
Remove usage of com.nimbusds.oauth2's HTTP classes

Avery-Dunn:
This PR removes usage of several com.nimbusds.oauth2 classes, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909

Mainly HTTPRequest, HTTPResponse, and ClientAuthentication were removed, along with some related minor utils. These were essentially just used as a helper for making HTTP requests: managing query parameters, creating assertions for different credentials, etc.

To replace them, this PR does the following:
- Refactors code in `ConfidentialClientApplication` related to credentials and assertions, it now mainly deals with assertion/secret Strings, some logic was moved to other classes, and some methods that now had duplicate or unused code
- Adjusts OAuthHttpRequest and HttpResponse to remove Nimbus classes and add the very minor features we were using from those classes
- Refactors code in `TokenRequestExecutor` related to credentials and assertions, simplifying and clarifying the behavior around query parameters and assertions/secrets that are needed in each scenario


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
```diff
@@ -29,7 +27,6 @@ public abstract class AbstractClientApplicationBase extends AbstractApplicationB
     private String applicationName;
     private String applicationVersion;
     private AadInstanceDiscoveryResponse aadAadInstanceDiscoveryResponse;
-    protected abstract ClientAuthentication clientAuthentication();
```
billwert, line 32: We don't need a replacement for this?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthHttpRequest.java
```diff
@@ -55,39 +54,31 @@ private Map<String, String> configureHttpHeaders() {
...
 
-        final HTTPResponse response = new HTTPResponse(httpResponse.statusCode());
+        final HttpResponse response = new HttpResponse();
+        response.statusCode(httpResponse.statusCode());
 
         final String location = HttpUtils.headerValue(httpResponse.headers(), "Location");
         if (!StringHelper.isBlank(location)) {
             try {
-                response.setLocation(new URI(location));
+                response.addHeader("Location", new URI(location).toString());
```
billwert, line 73: Maybe worth it to have an enum/constants for well known headers.

billwert (approved):

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractClientApplicationBase.java
Avery-Dunn: Long story short, this was mainly used as a parameter or return type of various nimbus methods. It's utility behavior was fully delegated to other classes:
- It represented a client assertion/secret/cert, which some existing MSAL classes did as well
- It was used to add query parameters to an HTTP request, but `TokenRequestExecutor.addJWTBearerAssertionParams()` now handles that
- It was passed into some nimbus APIs that are no longer used

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthHttpRequest.java
Avery-Dunn: I was planning to do this in a follow up PR that removes various nimbus utility methods, since we'll then have the full picture of what constants and enums would be useful

PR #929, opened 2025-04-01
Add APIs to add client capabilities and claims

neha-bhargava:
Add APIs to the managed identity app and request to add client capabilities and claims. Bypass the cache when claims are passed.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
@@ -103,15 +103,16 @@ public class ManagedIdentityApplication extends AbstractApplicationBase implemen
         }
 
         /**
-         * Sets client capabilities for the application.
+         * Informs the token issuer that the application is able to perform complex authentication actions.
+         * For example, "cp1" means that the application is able to perform conditional access evaluation,
+         * because the application has been setup to parse WWW-Authenticate headers associated with a 401 response from the protected APIs,
+         * and to retry the request with claims API.
          * 
-         * @param clientCapabilities List of client capabilities to be requested
-         * @return instance of Builder of ManagedIdentityApplication
+         * @param clientCapabilities a list of capabilities (e.g., ["cp1"]) recognized by the token service.
+         * @return instance of Builder of ManagedIdentityApplication.
          */
         public Builder clientCapabilities(List<String> clientCapabilities) {
-            if (clientCapabilities != null) {
-                this.clientCapabilities = clientCapabilities;
-            }
+            this.clientCapabilities = clientCapabilities;
             return self();
         }
 
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
@@ -5,6 +5,9 @@ package com.microsoft.aad.msal4j;
 
 import org.slf4j.LoggerFactory;
 
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 /**
@@ -56,6 +59,7 @@ public class ManagedIdentityApplication extends AbstractApplicationBase implemen
     public ManagedIdentityId getManagedIdentityId() {
         return this.managedIdentityId;
     }
+    
     @Override
     public CompletableFuture<IAuthenticationResult> acquireTokenForManagedIdentity(ManagedIdentityParameters managedIdentityParameters)
             throws Exception {
@@ -86,6 +90,7 @@ public class ManagedIdentityApplication extends AbstractApplicationBase implemen
 
         private String resource;
         private ManagedIdentityId managedIdentityId;
+        private List<String> clientCapabilities;
 
         private Builder(ManagedIdentityId managedIdentityId) {
             super(managedIdentityId.getIdType() == ManagedIdentityIdType.SYSTEM_ASSIGNED ?
@@ -99,9 +104,21 @@ public class ManagedIdentityApplication extends AbstractApplicationBase implemen
             return self();
         }
 
+        /**
+         * Sets client capabilities for the application.
+         * 
+         * @param clientCapabilities List of client capabilities to be requested
+         * @return instance of Builder of ManagedIdentityApplication
+         */
+        public Builder clientCapabilities(List<String> clientCapabilities) {
+            if (clientCapabilities != null) {
+                this.clientCapabilities = clientCapabilities;
+            }
+            return self();
+        }
+
         @Override
         public ManagedIdentityApplication build() {
-
             return new ManagedIdentityApplication(this);
         }
 
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
@@ -5,9 +5,7 @@ package com.microsoft.aad.msal4j;
 
 import org.slf4j.LoggerFactory;
 
-import java.util.HashSet;
 import java.util.List;
-import java.util.Set;
 import java.util.concurrent.CompletableFuture;
 
 /**
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
@@ -3,7 +3,7 @@
 
 package com.microsoft.aad.msal4j;
 
-import java.util.List;
+import java.util.Map;
 import java.util.Set;
 
 /**
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
@@ -4,7 +4,6 @@
 package com.microsoft.aad.msal4j;
 
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 
 /**
@@ -30,7 +29,7 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
 
     @Override
     public ClaimsRequest claims() {
-        return (claims != null) ? ClaimsRequest.fromJsonString(claims) : null;
+        return (claims != null) ? ClaimsRequest.formatAsClaimsRequest(claims) : null;
     }
 
     @Override
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
@@ -30,8 +30,7 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
     @Override
     public ClaimsRequest claims() {
         if (claims == null || claims.isEmpty()) {
-            throw new MsalClientException("Claims cannot be null or empty",
-                                         AuthenticationErrorCode.INVALID_JSON);
+            return null;
         }
 
         try {
@@ -108,6 +107,8 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
          * @return this builder instance
          */
         public ManagedIdentityParametersBuilder claims(String claims) {
+            ParameterValidationUtils.validateNotBlank("claims", claims);
+
             this.claims = claims;
             return this;
         }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
@@ -3,6 +3,7 @@
 
 package com.microsoft.aad.msal4j;
 
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
@@ -14,10 +15,12 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
 
     String resource;
     boolean forceRefresh;
+    String claims;
 
-    private ManagedIdentityParameters(String resource, boolean forceRefresh) {
+    private ManagedIdentityParameters(String resource, boolean forceRefresh, String claims) {
         this.resource = resource;
         this.forceRefresh = forceRefresh;
+        this.claims = claims;
     }
 
     @Override
@@ -27,7 +30,7 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
 
     @Override
     public ClaimsRequest claims() {
-        return null;
+        return (claims != null) ? ClaimsRequest.fromJsonString(claims) : null;
     }
 
     @Override
@@ -69,6 +72,7 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
     public static class ManagedIdentityParametersBuilder {
         private String resource;
         private boolean forceRefresh;
+        private String claims;
 
         ManagedIdentityParametersBuilder() {
         }
@@ -83,8 +87,13 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
             return this;
         }
 
+        public ManagedIdentityParametersBuilder claims(String claims) {
+            this.claims = claims;
+            return this;
+        }
+
         public ManagedIdentityParameters build() {
-            return new ManagedIdentityParameters(this.resource, this.forceRefresh);
+            return new ManagedIdentityParameters(this.resource, this.forceRefresh, this.claims);
         }
 
         public String toString() {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
@@ -29,7 +29,18 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
 
     @Override
     public ClaimsRequest claims() {
-        return (claims != null) ? ClaimsRequest.formatAsClaimsRequest(claims) : null;
+        if (claims == null || claims.isEmpty()) {
+            throw new MsalClientException("Claims cannot be null or empty",
+                                         AuthenticationErrorCode.INVALID_JSON);
+        }
+
+        try {
+            return ClaimsRequest.formatAsClaimsRequest(claims);
+        } catch (Exception ex) {
+            // Log the exception if the claims JSON is invalid
+            throw new MsalClientException("Failed to parse claims JSON: " + ex.getMessage(),
+                                         AuthenticationErrorCode.INVALID_JSON);
+        }
     }
 
     @Override
@@ -86,6 +97,16 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
             return this;
         }
 
+        /**
+         * Instructs the SDK to bypass any token caches and to request new tokens with an additional claims challenge.
+         * The claims challenge string is opaque to applications and should not be parsed.
+         * The claims challenge string is issued either by the STS as part of an error response or by the resource,
+         * as part of an HTTP 401 response, in the WWW-Authenticate header.
+         * For more details see https://learn.microsoft.com/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=dotnet
+         *
+         * @param claims a valid JSON string representing additional claims
+         * @return this builder instance
+         */
         public ManagedIdentityParametersBuilder claims(String claims) {
             this.claims = claims;
             return this;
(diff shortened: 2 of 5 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -99,9 +102,21 @@ public Builder resource(String resource) {
             return self();
         }
 
+        /**
+         * Sets client capabilities for the application.
+         * 
+         * @param clientCapabilities List of client capabilities to be requested
```
gladjohn:
```suggestion
         * @param clientCapabilities a list of capabilities (e.g., ["cp1"]) recognized by the token service
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -99,9 +102,21 @@ public Builder resource(String resource) {
...
+         * @return instance of Builder of ManagedIdentityApplication
```
gladjohn:
```suggestion
         * @return this Builder instance
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -99,9 +102,21 @@ public Builder resource(String resource) {
             return self();
         }
 
+        /**
+         * Sets client capabilities for the application.
```
gladjohn:
```suggestion
         * Sets the client capabilities requested during authentication.
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -83,8 +86,13 @@ public ManagedIdentityParametersBuilder forceRefresh(boolean forceRefresh) {
             return this;
         }
 
+        public ManagedIdentityParametersBuilder claims(String claims) {
```
gladjohn, line 109:
```suggestion
        /**
       * Optional claims in JSON format to be included in the token request.
       * Example: {"access_token":{"nbf":{"essential":true, "value":"1743543930"}}}
       *
       * @param claims a valid JSON string representing additional claims
       * @return this builder instance
       */
        public ManagedIdentityParametersBuilder claims(String claims) {
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -83,8 +86,13 @@ public ManagedIdentityParametersBuilder forceRefresh(boolean forceRefresh) {
...
+            this.claims = claims;
```
gladjohn, line 112: we should consider clarifying in code or Javadoc what happens if claims is an empty string, malformed JSON, or null.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -27,7 +29,7 @@ public Set<String> scopes() {
 
     @Override
     public ClaimsRequest claims() {
-        return null;
+        return (claims != null) ? ClaimsRequest.formatAsClaimsRequest(claims) : null;
```
gladjohn: would logging be helpful here. i.e. claims are present? not the actual claims itself

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -579,6 +580,38 @@ void managedIdentity_SharedCache(ManagedIdentitySourceType source, String endpoi
         verify(httpClientMock, times(1)).send(any());
     }
 
+    // managedIdentityTest_WithClaims: Tests that acquiring a token with claims works correctly
```
gladjohn, line 621: The new test method managedIdentityTest_WithClaims is a great start. We might also want to check:

- **Multiple capabilities:** If we pass multiple items in clientCapabilities, do we confirm they are all being sent in the request?

- **Claims edge cases:** If we pass an empty or invalid JSON, does the code handle it gracefully?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -579,6 +580,38 @@ void managedIdentity_SharedCache(ManagedIdentitySourceType source, String endpoi
...
+                .httpClient(httpClientMock)
+                .build();
+
+        // Clear caching to avoid cross test pollution.
+        miApp.tokenCache().accessTokens.clear();
+
+        String claimsJson = "{\"default\":\"claim\"}";
+        IAuthenticationResult result = miApp.acquireTokenForManagedIdentity(
+            ManagedIdentityParameters.builder(resource)
+                .claims(claimsJson)
```
gladjohn: do we already have tests that checks the final JSON structure explicitly when both claims and capabilities are set? if not we should consider adding it

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -579,6 +580,38 @@ void managedIdentity_SharedCache(ManagedIdentitySourceType source, String endpoi
...
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+        if (source == SERVICE_FABRIC) {
+            ServiceFabricManagedIdentitySource.setHttpClient(httpClientMock);
+        }
+
+        when(httpClientMock.send(expectedRequest(source, resource))).thenReturn(expectedResponse(200, getSuccessfulResponse(resource)));
+
+        miApp = ManagedIdentityApplication
+                .builder(ManagedIdentityId.systemAssigned())
+                .clientCapabilities(singletonList("cp1"))
```
gladjohn: would be good to test `null` and empty string `""`

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -579,6 +580,38 @@ void managedIdentity_SharedCache(ManagedIdentitySourceType source, String endpoi
         verify(httpClientMock, times(1)).send(any());
     }
 
+    // managedIdentityTest_WithClaims: Tests that acquiring a token with claims works correctly
+    @ParameterizedTest
+    @MethodSource("com.microsoft.aad.msal4j.ManagedIdentityTestDataProvider#createDataError")
+    void managedIdentityTest_WithClaims(ManagedIdentitySourceType source, String endpoint) throws Exception {
```
gladjohn, line 624: The test suggests just claims but even capability is being used.  maybe we can consider changing the test name

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
gladjohn: we should also consider adding a test without capability. i.e. just use the claims api

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -579,6 +580,38 @@ void managedIdentity_SharedCache(ManagedIdentitySourceType source, String endpoi
...
+        // Clear caching to avoid cross test pollution.
+        miApp.tokenCache().accessTokens.clear();
+
+        String claimsJson = "{\"default\":\"claim\"}";
+        IAuthenticationResult result = miApp.acquireTokenForManagedIdentity(
+            ManagedIdentityParameters.builder(resource)
+                .claims(claimsJson)
+                .build()).get();
+
+        assertNotNull(result.accessToken());
```
gladjohn, line 649: are we able to check if the token source is IDP when claims are passed?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
gladjohn, line 624: praise: Excellent new test coverage for the claims JSON scenario!

gladjohn:
The new features for client capabilities and claims looks great. Adding tests for edge cases (like empty or malformed JSON) will strengthen the code coverage further. Excellent job!!!

gladjohn:
nit: would be also good to have a tracking item in the PR description.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
gladjohn: reason: app developers can pass claims alone. they need not register for capability, and they should still get a new token.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: Not applying this change since this is consistent with rest of the comments

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
neha-bhargava: Is this the format in which the claims are expected? `{"accessToken":{"nbf":{"essential":true}}}`

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: In this PR I am only exposing the APIs. The rest of the work will be in another PR. CC: @bgavrilMS

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
gladjohn: it will be like this `{"access_token":{"nbf":{"essential":true, "value":"1743543930"}}}`

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: I am adding the APIs in this PR. I am not yet passing this to the endpoints

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
gladjohn: > In this PR I am only exposing the APIs. The rest of the work will be in another PR. CC: @bgavrilMS

**issue (blocking):** Exposing a public API without the corresponding functionality can be risky.

**discussion:** If this API is merged now, consumers might assume it’s ready. We typically prefer to finish the feature (and its tests) in a private or feature branch, then merge once the functionality is complete. Otherwise, we risk confusion or partial breakage for downstream users.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
```diff
@@ -99,9 +102,21 @@ public Builder resource(String resource) {
...
+         * 
+         * @param clientCapabilities List of client capabilities to be requested
+         * @return instance of Builder of ManagedIdentityApplication
+         */
+        public Builder clientCapabilities(List<String> clientCapabilities) {
+            if (clientCapabilities != null) {
```
Avery-Dunn: nit: Since the builder's `clientCapabilities` isn't set to a default value it'll be null anyway, so a null check here isn't necessary but also won't cause any problems.

gladjohn (changes requested):
when claims are present, we need to bypass MSALs cache

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
neha-bhargava: Yup, AI generated code I missed to look at.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
bgavrilMS: Didn't we add the APIs to MSAL.NET in the exact same way @gladjohn ?

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/c776ab9a8e263787c535463b9518d1521fbd59a9/src/client/Microsoft.Identity.Client/ApiConfig/AcquireTokenForManagedIdentityParameterBuilder.cs#L74

At a minimum, we should bypass MSAL's cache. This is still a value add, even  though we cannot bypass MSI caches.

And add a comment that the APIs will not work (i.e. no-op) on some sources.

Avery-Dunn: I don't think merging the APIs without the functionality is a big problem since the APIs related to managed identity won't be used as much as public/confidential, some comments like Bogdan suggested should be enough to let customers know what the situation is.

We could also keep it separate until fully implemented by merging this PR and follow ups into their own branch instead of dev, like I'm doing with some of the partial dependency removals: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/927

gladjohn: > Didn't we add the APIs to MSAL.NET in the exact same way @gladjohn ?

Yes, but we also have added the functionality to bypass cache. 

> At a minimum, we should bypass MSAL's cache. This is still a value add, even though we cannot bypass MSI caches.

agree @bgavrilMS 

@Avery-Dunn – I’d strongly recommend we bypass MSAL’s cache whenever claims are present.

**Reasoning:**
If token revocation occurs for a system-assigned or user-assigned managed identity, bypassing the cache helps us avoid serving a stale token. Instead, we always fetch a new token from the Managed Identity Resource Provider (MIRP). Even if the provider doesn’t fully support token revocation parameters, they can shorten token lifespans or handle partial revocation flows. That way, we reduce the risk of inadvertently returning invalid or expired tokens.

**Next Steps**

- Bypass MSAL’s cache in these scenarios.

This strikes a balance between making the APIs available and ensuring we don’t break consumer expectations regarding token freshness.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
neha-bhargava: I will throw here instead.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: I will add code to bypass cache when claims are present. That is also something that Bogdan suggested. Thanks everyone for your suggestions and perspectives. My goal here is to keep PRs short and since the whole token revocation with App service new api version cannot be merged, have a PR that can be merged and a small PR with changes that cannot be merged. Short PRs are easy to review and expedites the process. 

Like @Avery-Dunn mentioned can be done as a feature branch as well instead of dev.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: I will add more tests. Thanks

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityApplication.java
bgavrilMS: I find Neha's comment simpler to read.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
bgavrilMS: The claims string is opaque to the client. We should not give examples, particularly we should not suggest that json is the format.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
bgavrilMS: See above comment - claims are opaque concept. It's a string. Do not perform any validation.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
bgavrilMS: I suggest all MSALs add a comment like:

"Instructs the SDK to bypass any token caches and to request new tokens with an additional claims challenge. The claims challenge string is opaque to applications and should not be parsed. The claims challenge string is issued either by the STS as part of an error response or by the resource, as part of an HTTP 401 response, in the WWW-Authenticate header. For more details see https://learn.microsoft.com/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=dotnet"

and for CC

"Informs the token issuer that the application is able to perform complex authentication actions. For example, "cp1" means that the application is able to perform conditional access evaluation, because the application has been setup to parse WWW-Authenticate headers associated with a 401 response from the protected APIs, and to retry the request with claims API."

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
bgavrilMS: @gladjohn - did you document all the test cases developers are required to write? Is there a test plan associated with the spec?

gladjohn: We do not have test case document, but [MSAL .NET unit tests](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Unit/ManagedIdentityTests/ManagedIdentityTests.cs#L298) should be a good place to start.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
neha-bhargava: Updated

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: Updated tests.

gladjohn (dismissed):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -27,7 +29,18 @@ public Set<String> scopes() {
 
     @Override
     public ClaimsRequest claims() {
-        return null;
+        if (claims == null || claims.isEmpty()) {
```
Avery-Dunn, line 32: suggestion: If the null/empty check was move to the builder, then the rest of the library could just ignore it and wouldn't have to throw an exception

neha-bhargava: Yeah noticed it when the tests started failing. Always takes a bit to switch to Java style from .Net

Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
rayluo: > I suggest all MSALs add a comment like:
> 
> For MI
>
> "Instructs the SDK to bypass any token caches and to request new tokens with an additional claims challenge. The claims challenge string is opaque to applications and should not be parsed. The claims challenge string is issued either by the STS as part of an error response or by the resource, as part of an HTTP 401 response, in the WWW-Authenticate header. For more details see https://learn.microsoft.com/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=dotnet"
> 
> and for CC
> 
> "Informs the token issuer that the application is able to perform complex authentication actions. For example, "cp1" means that the application is able to perform conditional access evaluation, because the application has been setup to parse WWW-Authenticate headers associated with a 401 response from the protected APIs, and to retry the request with claims API."

Speaking of claims challenge format, I remember that back then when MSALs were developing that feature for CCA, there was no end-to-end test environment. So, we implemented based on the specs, [which seems to be expecting a string representation of a json blob without base64 encoding](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/4.70.2/src/client/Microsoft.Identity.Client/Internal/ClaimsHelper.cs#L41-L48). Will it work with the base64 encoded sample in [the quoted document](https://learn.microsoft.com/en-us/entra/identity-platform/app-resilience-continuous-access-evaluation?tabs=dotnet#handling-cae-in-your-application)?

Ideally, we shall document the expected claims format accordingly, even if our current implementation in MSI does not (yet?) parse the claim. Otherwise, when/if we decide to parse it in the future, a wrong format input would crash and become a breaking change.

Issue #930, opened 2025-04-03
Pass token source for managed identity requests


neha-bhargava:
Token source needs to be passed to result as IDP. And update the tests to check the token source

PR #931, opened 2025-04-03
Pass token source and update tests

neha-bhargava:
#930

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
@@ -13,7 +13,7 @@ class AcquireTokenByManagedIdentitySupplier extends AuthenticationResultSupplier
 
     private static final Logger LOG = LoggerFactory.getLogger(AcquireTokenByManagedIdentitySupplier.class);
 
-    private static final int TWO_HOURS = 2*3600;
+    private static final int TWO_HOURS = 2 * 3600;
 
     private ManagedIdentityParameters managedIdentityParameters;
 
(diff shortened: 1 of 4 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/SilentRequestHelper.java
```diff
@@ -0,0 +1,55 @@
...
+
+class SilentRequestHelper {
+
+    private static final int ACCESS_TOKEN_EXPIRE_BUFFER_IN_SEC = 5 * 60;
+
+    private SilentRequestHelper() {
+        // Utility class
+    }
+
+    static CacheRefreshReason NeedsRefresh(SilentParameters parameters, AuthenticationResult cachedResult, Logger log) {
```
Avery-Dunn: nit: Method names generally start with a lower-case letter in Java: https://www.oracle.com/java/technologies/javase/codeconventions-namingconventions.html

Also, since it's returning a `CacheRefreshReason` instead of a true/false, something like `getRefreshReason` or `determineRefreshReason` might be a better name.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -63,23 +65,37 @@ AuthenticationResult execute() throws Exception {
...
+                        LOG);
+
+                // If the token does not need a refresh, return the cached token
+                // Else refresh the token if it is either expired, proactively refreshable, or if the claims are passed.
+                if (cacheRefreshReason == CacheRefreshReason.NOT_APPLICABLE) {
+                    LOG.debug("Returning token from cache");
+                    result.metadata().tokenSource(TokenSource.CACHE);
+                    return result;
+                } else {
+                    LOG.debug(String.format("Refreshing access token. Cache refresh reason: %s", cacheRefreshReason));
```
Avery-Dunn: suggestion: When I first saw this I was a bit confused about where the access token would be refreshed, until I realized it'd be taken care of by the return down near line 95.

For better clarity about the code flow and returns. I'd suggest:
- Add an additional `return fetchNewAccessTokenAndSaveToCache()` here in the else
- Rather than have a negative if statement like `if (!managedIdentityParameters.forceRefresh)` contain all this other logic, cut the execution short with a ` if (managedIdentityParameters.forceRefresh) { return fetchNewAccessToken... }` at the top and have all the other logic come after that 
  - Like it says in the `SilentRequestHelper.needsRefresh()`, if `forceRefresh` is true then there's no need to check any other refresh option

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/SilentRequestHelper.java
neha-bhargava: Updated

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: I updated it, checking the force refresh in the beginning since force refresh do not need a cache lookup.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -37,49 +37,66 @@ AuthenticationResult execute() throws Exception {
...
+                    parameters,
+                    this.clientApplication,
+                    context,
+                    null);
+
+            AcquireTokenSilentSupplier supplier = new AcquireTokenSilentSupplier(
+                    this.clientApplication,
+                    silentRequest);
+
+            AuthenticationResult result = supplier.execute();
```
bgavrilMS, line 72: So much ceremony to get a token :)

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
neha-bhargava: All this definitely needs a refactoring. But not in this PR :)

Avery-Dunn (approved):

PR #936, opened 2025-04-10
Implement retryable call to withstand Connection reset, Connection refused, 503 Service not available when MSAL4J is used behind proxy

JozefDropco:
…issues/935 implement retryablecall to avoid socketexception


JozefDropco:
@microsoft-github-policy-service agree

PR #942, opened 2025-04-23
Represent query parameters with Map<String, String>

Avery-Dunn:
This PR refactors and simplifies how URL query parameters and handled in the library, made possible by the removal of dependences both here and in other PRs for https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909

Although the PR affects a lot of files, most shouldn't need much scrutiny: nearly all the changes are simply replacing `Map<String, List<String>>` with `Map<String, String>` and removing usage of `Collections.singletonList()`

However, there are a few files which need a more thorough review: 
- `StringHelper` : Several new helper methods for converting a `Map<String, String>` into a properly formatted URL query string
- `AuthorizationRequestUrlParameters`: All of the Maps are set internally, but unfortunately one public API in this class exposed its package-private Map field. The public API was untouched, but a new method was added to `StringHelper` to convert its internal Map back to the old format
- `HelperAndUtilityTests`: Unit tests for the new methods in `StringHelper`

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
@@ -125,7 +125,7 @@ final class StringHelper {
     }
 
     //Much of the library once used Map<String, List<String>> for query parameters due to reliance on a specific dependency.
-    // Although Map<String, String> is now used internally, some public APIs still return Map<String, List<String>>
+    // Although Map<String, String> is now used internally, some public APIs in AuthorizationRequestUrlParameters still return Map<String, List<String>>
     static Map<String, List<String>> convertToMultiValueMap(Map<String, String> singleValueMap) {
         Map<String, List<String>> multiValueMap = new HashMap<>();
 
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
@@ -3,17 +3,20 @@
 
 package com.microsoft.aad.msal4j;
 
+import java.io.UnsupportedEncodingException;
+import java.net.URLDecoder;
+import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
 import java.security.MessageDigest;
 import java.security.NoSuchAlgorithmException;
-import java.util.Base64;
+import java.util.*;
 
 final class StringHelper {
 
     static String EMPTY_STRING = "";
 
     static boolean isBlank(final String str) {
-        return str == null || str.trim().length() == 0;
+        return str == null || str.trim().isEmpty();
     }
 
     static String createBase64EncodedSha256Hash(String stringToHash) {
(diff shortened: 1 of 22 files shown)


rayluo:
Can you provide more context for this change? For example, is it because the removal of some dependencies makes handling `Map<String, List<String>>` difficult or impossible?

The thing is, the original `Map<String, List<String>>` data structure is conceptually correct, because it is [possible to have same-name query parameter with multiple values](https://stackoverflow.com/q/11621477/728675). Granted, we might not see those duplication in OAuth2 protocols frequently, but if it would ever come up, we will need another major refactor. If this codebase has already paid the price and did it the `Map<String, List<String>>` way, why do we bother simplify it?

Avery-Dunn:
> Can you provide more context for this change? For example, is it because the removal of some dependencies makes handling `Map<String, List<String>>` difficult or impossible?

It's kind of the opposite @rayluo , a dependency required us to use `Map<String, List<String>>` and removing the dependency made using `Map<String, String>` possible. 

For context, some previous PRs removed a bunch of `com.nimbusds` imports which used `Map<String, List<String>>` for parameters, and some comments (https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/926#discussion_r2016543637) pointed out we were making a lot unnecessary `List` creations since we only passed in one parameter. Plus retrieving a single parameter value was complicated, since it needed several null checks and an index size check to avoid exceptions.

After removing the rest of the imports, it became clear that MSAL Java never used more than a single query parameter internally, and we didn't have a Public API for that format to come from the customer. If we have a future scenario that requires it, it shouldn't be too hard to undo all the changes in this PR, or add a new helper method that does the opposite of this PR's new `StringHelper.convertToMultiValueMap()`

rayluo:
> It's kind of the opposite @rayluo , a dependency required us to use `Map<String, List<String>>` and removing the dependency made using `Map<String, String>` possible.

I mean that dependency's design choice was right. That said, I agree with you that we can afford simplifying it here, because it is not part of our public API so we still have flexibility to refactor if/when the time comes.

Consider this conversation resolved.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
```diff
@@ -42,6 +45,96 @@ static private String createSha256Hash(String stringToHash, boolean base64Encode
...
+
+                params.put(key, value);
+            }
+
+            return params;
+        }
+    }
+
+    //Much of the library once used Map<String, List<String>> for query parameters due to reliance on a specific dependency.
+    // Although Map<String, String> is now used internally, some public APIs still return Map<String, List<String>>
```
neha-bhargava: This is not needed for all the cases? When you say some public APIs which ones?

neha-bhargava (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
Avery-Dunn: Just `AuthorizationRequestUrlParameters` returned query parameters as a `Map<String, List<String>>` in a public API, and in the latest commit I added that info the comment.

PR #943, opened 2025-04-23
Token revocation for service fabric

neha-bhargava:
Started this PR with AI generated code.
Updated manually after updating the tests to validate the query params


4gust:
This PR missed the updated App Service new API version.
Gladwin's update on spec should be able to resolve this.
Rest of it looks good

msal4j-sdk/pom.xml
```diff
@@ -157,6 +157,12 @@
             <version>2.14.0</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.xml.bind</groupId>
+            <artifactId>jakarta.xml.bind-api</artifactId>
```
Avery-Dunn, line 162: Is this new dependency used anywhere? Doesn't seem like there are new imports with it, and we're trying to remove dependencies anyway.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRevocationUtil.java
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Converts a token to its SHA256 hash string representation.
+     * This is used in token revocation scenarios to identify tokens without transmitting the original token.
+     *
+     * @param token The token to hash
+     * @return The SHA256 hash of the token as a lowercase hex string
+     * @throws MsalClientException If the SHA-256 algorithm is not available
+     */
+    public static String convertTokenToSHA256HashString(String token) {
```
Avery-Dunn: I don't think we need this new class, unless there's going to be more utils just for token revocation. For hashing, there's an existing method in StringHelper that creates a SHA256 hash: https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/54f3d44e13e2e3aaa2743d24fd0835fc132db6a0/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java#L23

msal4j-sdk/pom.xml
Avery-Dunn: We're trying to remove reliance on dependencies as much as possible: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909. Seems like this is only used in the new TokenRevocationUtil class, and shouldn't be needed if that class is removed (see my comment there).

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRevocationUtil.java
neha-bhargava: Removed

msal4j-sdk/pom.xml
neha-bhargava: Removed

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
```diff
@@ -57,6 +57,7 @@ public ManagedIdentityResponse getManagedIdentityResponse(
             ManagedIdentityParameters parameters) {
 
         createManagedIdentityRequest(parameters.resource);
+        managedIdentityRequest.addTokenRevocationParametersToQuery(parameters);
```
gladjohn, line 60: if this will go into each of the sources then the cae query params will be added only to those sources, right? in this case it will be added only to SF. Do we still need a CAE list? Ok to have a double check but wondering if it is really needed.

gladjohn (approved):
Looks good Neha. Just one clarification comment.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ServiceFabricManagedIdentitySource.java
neha-bhargava: In Java, the createManagedIdentityRequest is either created from ServiceFabric or AbstractManagedIdentity class. I am using this method in both the locations and as we add more sources it can go there. I like this way of having the list of supporting sources, I am also using it to add parameters for supported sources in the mocking.

gladjohn: thanks @neha-bhargava, yes if AbstractManagedIdentity handles revocation scenarios directly then we need the source check to allow only token revocation supported sources to add the cae capable query params.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Constants.java
```diff
@@ -23,4 +26,13 @@ final class Constants {
     public static final String MSI_ENDPOINT = "MSI_ENDPOINT";
     public static final String IDENTITY_SERVER_THUMBPRINT = "IDENTITY_SERVER_THUMBPRINT";
 
+    // Constants for token revocation and client capabilities
+    public static final String TOKEN_HASH_CLAIM = "token_sha256_to_refresh";
+    public static final String CLIENT_CAPABILITY_REQUEST_PARAM = "xms_cc";
+    
+    // Only Service Fabric and App Service managed identity environments support token revocation
+    public static final Set<ManagedIdentitySourceType> TOKEN_REVOCATION_SUPPORTED_ENVIRONMENTS = new HashSet<ManagedIdentitySourceType>() {{
+        add(ManagedIdentitySourceType.SERVICE_FABRIC);
```
Avery-Dunn, line 35: The comment says that SF and app service support revocation, so should ManagedIdentitySourceType.APP_SERVICE be added here too?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -15,7 +15,7 @@ public class ManagedIdentityParameters implements IAcquireTokenParameters {
     String resource;
     boolean forceRefresh;
     String claims;
-
+    String revokedTokenHash;
     private ManagedIdentityParameters(String resource, boolean forceRefresh, String claims) {
```
Avery-Dunn: nit: very minor, just adding a blank line between all the fields and the first method

```suggestion
    String revokedTokenHash;
    
    private ManagedIdentityParameters(String resource, boolean forceRefresh, String claims) {

File: msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java [hunk 6]
```diff
@@ -41,6 +41,39 @@ static private String createSha256Hash(String stringToHash, boolean base64Encode
...
 
+    /**
+     * Creates a SHA-256 hash of the input string and returns it as a lowercase hex string.
+     * This is used for token revocation and other scenarios requiring hex hash representation.
+     *
+     * @param stringToHash The string to hash
+     * @return The SHA-256 hash of the string as a lowercase hex string
+     * @throws MsalClientException If the SHA-256 algorithm is not available
+     */
+    public static String createSha256HashHexString(String stringToHash) {
```
Avery-Dunn: nit: nothing outside the package can access StringHelper methods, so no need to be public 

```suggestion
    static String createSha256HashHexString(String stringToHash) {
```

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -669,6 +685,46 @@ void managedIdentityTest_WithClaims(ManagedIdentitySourceType source, String end
...
+
+        // Second call, get the token from the cache without passing the claims.
+        result = miApp.acquireTokenForManagedIdentity(
+                ManagedIdentityParameters.builder(resource)
+                        .build()).get();
+
+        assertNotNull(result.accessToken());
+        assertEquals(TokenSource.CACHE, result.metadata().tokenSource());
+
+        verify(httpClientMock, times(2)).send(any());
```
Avery-Dunn: Seems like this test is failing because httpClientMock.send() is only being called once. 

Since the second acquireTokenForManagedIdentity() result is expected to come from the cache, should this expect the httpClientMock.send() method to be called only once? Or is there something else that should try to use httpClientMock.send() during that second token call?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Constants.java
neha-bhargava: I am only adding service fabric here since app service will need version update to pass these and that version is not available yet.

neha-bhargava: I am updating the comment for now to only have service fabric here

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
neha-bhargava: Done

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
neha-bhargava: Missed. Thanks for pointing it out.

Avery-Dunn (approved):

PR #945, opened 2025-04-29
Remove final usages of com.nimbusds from the packaged codebase

Avery-Dunn:
Removes the final usages of com.nimbusds from the main library, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909

This PR makes the following changes:
- Adds new methods to `JsonHelper` to parse encoded tokens into JSON-formatted strings, and convert JSON strings into `Map<String, Object>` representing claims
- Replaces several nimbus imports with the new helper methods
- Removes some leftover imports, such as various nimbus exception types that are no longer thrown during tests
- Changes the scope of the nimbus dependency to 'test', which means customers will no longer bring nimbus into their projects because of MSAL
  - A couple tests still use nimbus, though the final JSON changes in the next PR will likely solve that

This PR also removes the `CustomJWTAuthentication` class, which was added in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/458. This was originally used as a workaround for some unnecessary validation that nimbus was doing, so was no longer needed.

However, it was unfortunately and unnecessarily added as a public class that extended a nimbus class. Therefore it is impossible to remove nimbus without this potentially breaking change, but thankfully there's no reason for a customer to have been using it so hopefully the impact is minimal (and this will be part of a beta release to help confirm).


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -40,15 +41,77 @@ static <T> T convertJsonToObject(final String json, final Class<T> tClass) {
...
-            idTokenJson = new String(Base64.getUrlDecoder().decode(token.split("\\.")[1]), StandardCharsets.UTF_8);
+            return new String(Base64.getUrlDecoder().decode(token.split("\\.")[1]), StandardCharsets.UTF_8);
         } catch (ArrayIndexOutOfBoundsException e) {
             throw new MsalClientException("Error parsing ID token, missing payload section.",
                     AuthenticationErrorCode.INVALID_JWT);
         }
+    }
+
+    //Converts a generic JSON string to a Map<String, Object> with relevant types
+    static Map<String, Object> parseJsonToMap(String jsonString) {
```
bgavrilMS, line 63: This code path doesn't have sufficient tests. You should parse a few id tokens that cover all the cases (objects, numbers, booleans, null) with the old way of parsing (manually). And then programatically check a few id tokens (or one complex "id token" with all cases) and see that the parsed version matches the old one.

bgavrilMS (changes requested):
Insufficient tests

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -76,12 +67,7 @@ private ITenantProfile getTenantProfile() {
             return null;
         }
 
-        try {
-            return new TenantProfile(JWTParser.parse(idToken).getJWTClaimsSet().getClaims(),
-                    getAccount().environment());
-        } catch (ParseException e) {
-            throw new MsalClientException("Cached JWT could not be parsed: " + e.getMessage(), AuthenticationErrorCode.INVALID_JWT);
```
neha-bhargava, line 83: With this change this error message changes which I think is more explanatory of what failed. Catch and create error message here instead?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -40,15 +41,77 @@ static <T> T convertJsonToObject(final String json, final Class<T> tClass) {
     }
 
     static IdToken createIdTokenFromEncodedTokenString(String token) {
-        String idTokenJson;
+        return JsonHelper.convertJsonToObject(getTokenPayloadClaims(token), IdToken.class);
```
neha-bhargava, line 49: Would a log help here in a try catch block to clarify that parsing id token failed?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
billwert, line 63: Did you try `JsonReader.readMap(ReadValueCallback<JsonReader, T> valueReaderFunc)`? May be able to simplify your logic here some.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -40,15 +41,77 @@ static <T> T convertJsonToObject(final String json, final Class<T> tClass) {
...
+
+        try (JsonReader jsonReader = JsonProviders.createReader(jsonString)) {
+            jsonReader.nextToken();
+            return parseJsonObject(jsonReader);
+        } catch (IOException e) {
+            throw new MsalJsonParsingException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
+        }
+    }
+
+    private static List<Object> parseJsonArray(JsonReader jsonReader) throws IOException {
```
billwert: Same here - `readArray` might help you out?

billwert (changes requested):
I think we have helper methods that can at least reduce some of your testing burden (for malformed shapes, etc.) Let us know if you have questions there.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: Unfortunately, JsonReader.readMap does not seem to produce the same Map as the previous nimbus-based methods. For example, nimbus converted unix timestamps to Date objects, and some fields that could be arrays or single strings were always treated as arrays. So for backwards compatibility we need to match the old nimbus behavior.

In the latest commit I added new tests and a new handleSpecialFields() method to JsonHelper to correctly handle and better demonstrate this behavior.

Avery-Dunn: New tests were added in the latest commit, using the new parsing style as well as the old nimbus-based methods to parse claims out of an encoded token that has strings, numbers, nested JSONs, and null values.

Those tests also helped find a few cases where the new parsing style differed from the old, and new methods were added to JsonHelper to handle those special cases.

msal4j-sdk/pom.xml
```diff
@@ -58,7 +59,7 @@
             <groupId>org.projectlombok</groupId>
             <artifactId>lombok</artifactId>
             <version>1.18.36</version>
-            <scope>provided</scope>
+            <scope>test</scope>
```
billwert, line 62: When can this be removed?

billwert (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
billwert: Fair enough. You could maybe save some headache around managing state (keeping track of if you have progressed the parser, etc) by just having a callback that handled those special cases, but this is fine.

msal4j-sdk/pom.xml
Avery-Dunn: There will likely be one final PR removing all the leftover dependency references, I just wanted to keep the number of changed files in each PR as low as possible for easier review and avoid duplicate changes/complicated merging

Some of the infrastructure used by tests to get tokens/users/etc. use lombok and other dependencies, but it doesn't really affect the tests themselves

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: In the latest commit it now just uses readArray(JsonReader::readUntyped) rather than this method.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
Avery-Dunn: In the latest commit there is now better logging and exception handling in the JsonHelper methods that this code uses.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: In the latest commit there is now better logging and exception handling to help identify what failed.

PR #948, opened 2025-05-07
Replace com.fasterxml.jackson with com.azure.json

Avery-Dunn:
Replace final usages of com.fasterxml.jackson with com.azure.json, as per https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909

This PR makes changes to a lot of files, as they were unfortunately very interconnected and could not be split up into to multiple PRs like previous dependency removals.

However, most of the files received the same set of changes:
- Adjust classes to implement azure-json's [JsonSerializable](https://learn.microsoft.com/en-us/java/api/com.azure.json.jsonserializable?view=azure-java-stable) interface, mainly by adding two methods:
  - `toJson()`, which creates a JsonWriter that can be used to create a JSON-formatted String
  - `fromJson()`, which creates an instance of that class from a JsonReader (which is give a JSON-formatted string)
- Remove jackson-databind's JSON annotations from those classes

The following files are the exception, and have more bespoke changes that need a more thorough review:
 - `JsonHelper`: Received new methods and refactoring in order to better provide JSON helper methods to the rest of the codebase
 - `TokenCache`: Complete overhaul of the serialization/deserialization code
 
 Finally, there were a few files that only received small changes to use the new JsonHelper methods.
 
 
For testing, many integration tests deal with the cache and received some small changes to work with the new helper methods and exact serialization results, and new unit tests were added to `CacheTests` specifically to ensure the serialization/deserialization changes in `TokenCache` were covered
- One issue that became clear during this refactoring is that some tests used junk data in their persistent caches which was not used in anyway by the library (i.e., random field names), but was nonetheless added to the in-memory cache. Under the new system this type of junk data is no longer added to the in-memory cache, however this should not be a breaking change since the library would not have used it anyway

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
@@ -7,41 +7,22 @@ import com.azure.json.JsonProviders;
 import com.azure.json.JsonReader;
 import com.azure.json.JsonSerializable;
 import com.azure.json.JsonToken;
+import com.azure.json.JsonWriter;
 import com.azure.json.ReadValueCallback;
-import com.fasterxml.jackson.annotation.JsonInclude;
-import com.fasterxml.jackson.core.JsonParser;
-import com.fasterxml.jackson.databind.DeserializationFeature;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.node.ObjectNode;
-import com.fasterxml.jackson.databind.JsonNode;
 
+import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.io.StringWriter;
 import java.nio.charset.StandardCharsets;
 import java.util.*;
 
 class JsonHelper {
-    static ObjectMapper mapper;
-
-    static {
-        mapper = new ObjectMapper();
-        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
-        mapper.configure(JsonParser.Feature.ALLOW_UNQUOTED_CONTROL_CHARS, true);
-        mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
-    }
 
     private JsonHelper() {
     }
 
-    static <T> T convertJsonToObject(final String json, final Class<T> tClass) {
-        try {
-            return mapper.readValue(json, tClass);
-        } catch (Exception e) {
-            throw new MsalJsonParsingException(e.getMessage(), AuthenticationErrorCode.INVALID_JSON);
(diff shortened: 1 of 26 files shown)


bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -41,19 +41,10 @@ public TokenCache(ITokenCacheAccessAspect tokenCacheAccessAspect) {
...
-    @JsonProperty("IdToken")
     Map<String, IdTokenCacheEntity> idTokens = new LinkedHashMap<>();
-
-    @JsonProperty("Account")
     Map<String, AccountCacheEntity> accounts = new LinkedHashMap<>();
-
-    @JsonProperty("AppMetadata")
     Map<String, AppMetadataCacheEntity> appMetadata = new LinkedHashMap<>();
 
     transient ITokenCacheAccessAspect tokenCacheAccessAspect;
```
billwert: azure-json does not observe `transient` - you can remove usages.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
```diff
@@ -123,90 +97,102 @@ static <T extends JsonSerializable<T>> T convertJsonStringToJsonSerializableObje
...
+            String key = entry.getKey();
+            Object value = entry.getValue();
 
-            String fieldName = fieldNames.next();
-            JsonNode jsonNode = mainNode.get(fieldName);
-
-            if (jsonNode != null && jsonNode.isObject()) {
-                mergeJSONNode(jsonNode, addNode.get(fieldName));
+            if (mainMap.containsKey(key) && mainMap.get(key) instanceof Map && value instanceof Map) {
+                mergeJsonMaps((Map<String, Object>) mainMap.get(key), (Map<String, Object>) value);
```
billwert, line 172: Are all the inputs to this function controlled? Could malformed / malicious json be a problem here?

billwert (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JsonHelper.java
Avery-Dunn: Other than the recursive call, `mergeJsonMaps` is only used in the `mergeJSONString`, which converts the two json strings to a map using JsonReader so I don't think malformed json would be an issue.

Malicious json could be, as this is used for claims and client capabilities, which are ultimately strings that come from the customer. However, the values in those jsons aren't something we can control or really verify, so I don't think there's much more we can do beyond JsonReader's parsing.

PR #955, opened 2025-05-14
Add file based Azure Arc detection

Avery-Dunn:
Fixes the issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/846, and is similar to the fix in MSAL .NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4856

This PR adds file-based detection of Azure Arc, in addition to the current detection based on environment variables. If the files are found but the identity endpoint variable is not set, we default to a known endpoint.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
@@ -26,6 +26,7 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
     private static final String FILE_EXTENSION = ".key";
     private static final int MAX_FILE_SIZE_BYTES = 4096;
     private static final String WWW_AUTHENTICATE_HEADER = "WWW-Authenticate";
+    private static final String FALLBACK_IDENTITY_ENDPOINT = "http://127.0.0.1:40342/metadata/identity/oauth2/token";
 
     private final URI MSI_ENDPOINT;
 
@@ -33,6 +34,12 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
     {
         IEnvironmentVariables environmentVariables = getEnvironmentVariables();
         String identityEndpoint = environmentVariables.getEnvironmentVariable(Constants.IDENTITY_ENDPOINT);
+
+        if (StringHelper.isNullOrBlank(identityEndpoint)) {
+            LOG.info("[Managed Identity] Azure Arc was detected through file based detection but the environment variables were not found. Defaulting to known azure arc endpoint.");
+            identityEndpoint = FALLBACK_IDENTITY_ENDPOINT;
+        }
+
         String imdsEndpoint = environmentVariables.getEnvironmentVariable(Constants.IMDS_ENDPOINT);
 
         URI validatedUri = validateAndGetUri(identityEndpoint, imdsEndpoint);
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
@@ -6,11 +6,15 @@ package com.microsoft.aad.msal4j;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import java.io.File;
+
 /**
  * Class to initialize a managed identity and identify the service.
  */
 class ManagedIdentityClient {
     private static final Logger LOG = LoggerFactory.getLogger(ManagedIdentityClient.class);
+    private static final String WINDOWS_HIMDS_FILEPATH  = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
+    private static final String LINUX_HIMDS_FILEPATH = "/opt/azcmagent/bin/himds";
 
     static ManagedIdentitySourceType getManagedIdentitySource() {
         IEnvironmentVariables environmentVariables = AbstractManagedIdentitySource.getEnvironmentVariables();
@@ -24,8 +28,7 @@ class ManagedIdentityClient {
             }
         } else if (!StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.MSI_ENDPOINT))) {
             return ManagedIdentitySourceType.CLOUD_SHELL;
-        } else if (!StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IDENTITY_ENDPOINT)) &&
-                !StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IMDS_ENDPOINT))) {
+        } else if (validateAzureArcEnvironment(environmentVariables)) {
             return ManagedIdentitySourceType.AZURE_ARC;
         } else {
             return ManagedIdentitySourceType.DEFAULT_TO_IMDS;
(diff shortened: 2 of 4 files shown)


neha-bhargava (approved):
The changes looks good, were you able to test it?

gladjohn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
```diff
@@ -65,4 +68,33 @@ private static AbstractManagedIdentitySource createManagedIdentitySource(MsalReq
                 return new IMDSManagedIdentitySource(msalRequest, serviceBundle);
         }
     }
+
+    static boolean validateAzureArcEnvironment(IEnvironmentVariables environmentVariables) {
+        if (!StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IDENTITY_ENDPOINT)) &&
+                !StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IMDS_ENDPOINT))) {
+            LOG.debug("[Managed Identity] Azure Arc managed identity is available through environment variables.");
```
bgavrilMS: I'd bump these log messages to info

bgavrilMS (changes requested):
Needs at least a test for whatever OS the CI is using.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
Avery-Dunn: Done in latest commit.

Avery-Dunn:
> Needs at least a test for whatever OS the CI is using.

@bgavrilMS : I originally didn't add tests because it is very difficult to modify static/final variables in a test and our mocking framework doesn't allow mocks of low-level Java features like file manipulation, and the PR over in .NET didn't have them so I figured y'all ran into similar issues too.

However, in the latest commit I removed the 'final' modifier from the final paths and added some package-private setters, allowing the tests to create temp directories to test the new behavior. We have some similar used-for-tests-only setters for a few other fields that otherwise could be private, and this is basically the only way to have meaningful tests for this file detection.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
```diff
@@ -26,13 +26,20 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
     private static final String FILE_EXTENSION = ".key";
     private static final int MAX_FILE_SIZE_BYTES = 4096;
     private static final String WWW_AUTHENTICATE_HEADER = "WWW-Authenticate";
+    private static final String FALLBACK_IDENTITY_ENDPOINT = "http://127.0.0.1:40342/metadata/identity/oauth2/token";
```
Robbie-Microsoft, line 29:
```suggestion
    private static final String FALLBACK_AZURE_ARC_IDENTITY_ENDPOINT = "http://127.0.0.1:40342/metadata/identity/oauth2/token";
```

Robbie-Microsoft: it's a nit, you can choose to ignore

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
```diff
@@ -26,13 +26,20 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
...
     {
         IEnvironmentVariables environmentVariables = getEnvironmentVariables();
         String identityEndpoint = environmentVariables.getEnvironmentVariable(Constants.IDENTITY_ENDPOINT);
+
+        if (StringHelper.isNullOrBlank(identityEndpoint)) {
+            LOG.info("[Managed Identity] Azure Arc was detected through file based detection but the environment variables were not found. Defaulting to known azure arc endpoint.");
+            identityEndpoint = FALLBACK_IDENTITY_ENDPOINT;
+        }
+
         String imdsEndpoint = environmentVariables.getEnvironmentVariable(Constants.IMDS_ENDPOINT);
```
Robbie-Microsoft, line 43: Don't you also need to check that `imdsEndpoint` doesn't exist before using `FALLBACK_IDENTITY_ENDPOINT`?

Robbie-Microsoft: and if neither exists, also set `imdsEndpoint = "N/A: himds executable exists";`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
```diff
@@ -6,11 +6,15 @@
...
 
+import java.io.File;
+
 /**
  * Class to initialize a managed identity and identify the service.
  */
 class ManagedIdentityClient {
     private static final Logger LOG = LoggerFactory.getLogger(ManagedIdentityClient.class);
+    private static String WINDOWS_HIMDS_FILEPATH  = "%Programfiles%\\AzureConnectedMachineAgent\\himds.exe";
+    private static String LINUX_HIMDS_FILEPATH = "/opt/azcmagent/bin/himds";
```
Robbie-Microsoft, line 17: nit: should you define a custom object instead? This is what we do in msal-node:

```
type FilePathMap = {
    win32: string;
    linux: string;
};

export const SUPPORTED_AZURE_ARC_PLATFORMS: FilePathMap = {
    win32: `${process.env["ProgramData"]}\\AzureConnectedMachineAgent\\Tokens\\`,
    linux: "/var/opt/azcmagent/tokens/",
};

export const AZURE_ARC_FILE_DETECTION: FilePathMap = {
    win32: `${process.env["ProgramFiles"]}\\AzureConnectedMachineAgent\\himds.exe`,
    linux: "/opt/azcmagent/bin/himds",
};
```

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityClient.java
```diff
@@ -65,4 +72,44 @@ private static AbstractManagedIdentitySource createManagedIdentitySource(MsalReq
...
+    static boolean validateAzureArcEnvironment(IEnvironmentVariables environmentVariables) {
+        if (!StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IDENTITY_ENDPOINT)) &&
+                !StringHelper.isNullOrBlank(environmentVariables.getEnvironmentVariable(Constants.IMDS_ENDPOINT))) {
+            LOG.info("[Managed Identity] Azure Arc managed identity is available through environment variables.");
+            return true;
+        }
+
+        String osName = System.getProperty("os.name").toLowerCase();
+
+        if (osName.contains("windows")) {
```
Robbie-Microsoft, line 85: Can "windows" and "linux" be saved as constants in a constants file? Surely they are used elsewhere, or will be? Otherwise, if you implement my suggestion above, you could get the file path via:

```
// get the expected Windows or Linux file path of the himds executable
const fileDetectionPath: string =
    AZURE_ARC_FILE_DETECTION[process.platform as keyof FilePathMap];
```

Robbie-Microsoft:
You might find msal-node's implementation of this feature interesting. It can be found in [this file](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/client/ManagedIdentitySources/AzureArc.ts).

Avery-Dunn:
Closing, will come back to this if there's a customer requesting this support.

PR #959, opened 2025-05-29
Add versioning and support doc

Avery-Dunn:
Adds RELEASES.md, similar to https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/RELEASES.md

This doc describes our versioning system and support lifecycle.


rayluo (approved):
Approving with suggestion.

LGTM in this commit.

All these files in different repo tend to become out of sync over time. We may consider just replace its content with a link to MSAL .Net's releases.md and make it the single source of truth.

PR #960, opened 2025-05-30
Refactor retry logic and add special behavior for IMDS

Avery-Dunn:
This PR refactors HTTP retry logic by splitting it into classes that implement a new interface, similar to what was done in MSAL Node and .NET:
https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/7614
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5231

The following changes were made:
- A new `IRetryPolicy` interface, defining methods to determine if a certain HTTP response should be retried, the number of retries to perform, and how long to wait between retries
- Three new classes that implement `IRetryPolicy`:
  - `IMDSRetryPolicy`, for behavior specific to IMDS
  - `ManagedIdentityRetryPolicy`, for all other managed identity scenarios
  - `DefaultRetryPolicy`, for all other scenarios
- Minor changes to internal APIs and method calls to use the new classes

For testing:
 - Much of the retry behavior remains the same and is already covered by unit tests, such as those found in `RequestThrottlingTest` and `ManagedIdentityTests.managedIdentityTest_Retry()`
 - The exception is the different retry behaviors in `IMDSRetryPolicy` based on status codes, which is now covered by the new unit test in `ManagedIdentityTests`
 - In addition, each retry policy class was given package-private methods to adjust the delays between retries, to significantly speed up the unit tests
 
 Most of the "files changed" list in this PR are due to changes in two other PRs:
 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960
 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/964


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+ * Default retry policy for most MSAL Java flows
+ */
+class DefaultRetryPolicy implements IRetryPolicy {
+    private static final int RETRY_NUM = 1;
+    private static int RETRY_DELAY_MS = 1000;
+
+    @Override
+    public boolean isRetryable(IHttpResponse httpResponse) {
+        return httpResponse.statusCode() >= 500 &&
+                httpResponse.statusCode() < 600 &&
```
Robbie-Microsoft: There should already be constants defined for these numbers? See in msal-node: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-common/src/utils/Constants.ts#L62

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private int currentRetryCount;
+    private int lastStatusCode;
+    
+    @Override
+    public boolean isRetryable(IHttpResponse httpResponse) {
+        currentRetryCount++;
+        lastStatusCode = httpResponse.statusCode();
+
+        return (lastStatusCode >= 500 && lastStatusCode < 600) ||
```
Robbie-Microsoft: see other comment about constants

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    //Package-private methods to allow much quicker testing. The delay values should be treated as constants in any non-test scenario.
+    static void setRetryDelayMs(int retryDelayMs) {
+        RETRY_DELAY_MS = retryDelayMs;
+    }
+
+    static void resetToDefaults() {
+        RETRY_DELAY_MS = 1000;
+    }
```
Robbie-Microsoft: I think you should consider setting up a retry policy factory like in .net: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5308

Additionally, set up a test retry policy factory.

This would eliminate the need for static variables.

What are your thoughts?

Robbie-Microsoft (changes requested):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
Avery-Dunn: In the latest commit there's a new `HttpStauts` enum with a variety of HTTP status codes and some helper methods. To keep this PR simple I only used them to the RetryPolicy classes, but they will be used in other classes as they get expanded/refactored.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
Avery-Dunn: That pattern is definitely better, and I've just spent some time trying to match what you're doing in .NET

Unfortunately, I don't think I can do it cleanly with how the HTTP-related classes currently work: there's no good place to inject the custom retry behavior, so it still needs a static config object (bad) or changing the constants through reflection (worse).

I think the best way forward is to match some of .NET's HTTP behavior around HttpManager and refactoring usage of HTTP classes to be more consistent, but that'll be a bigger refactor in another PR rather than this one.

Robbie-Microsoft: Should you open a new issue on GitHub to track the refactor? (genuinely curious)

Avery-Dunn: Yep, just made the issue to track it: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/961

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public boolean isRetryable(IHttpResponse httpResponse) {
+        currentRetryCount++;
+        lastStatusCode = httpResponse.statusCode();
+
+        return HttpStatus.isServerError(lastStatusCode) ||
+                lastStatusCode == HttpStatus.NOT_FOUND.getCode() ||
+                lastStatusCode == HttpStatus.REQUEST_TIMEOUT.getCode() ||
+                lastStatusCode == HttpStatus.GONE.getCode() ||
+                lastStatusCode == HttpStatus.TOO_MANY_REQUESTS.getCode();
```
Robbie-Microsoft: Not including `isServerError`, you should consider adding the other http status codes to an array, like in msal-node:

```
const HTTP_STATUS_400_CODES_FOR_EXPONENTIAL_STRATEGY: Array<number> = [
    HttpStatus.NOT_FOUND,
    HttpStatus.REQUEST_TIMEOUT,
    HttpStatus.GONE,
    HttpStatus.TOO_MANY_REQUESTS,
];
```

This is just a suggestion, you can ignore it.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRetryPolicy.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+class ManagedIdentityRetryPolicy implements IRetryPolicy {
+    private static final int RETRY_NUM = 3;
+    private static int RETRY_DELAY_MS = 1000;
+
+    @Override
+    public boolean isRetryable(IHttpResponse httpResponse) {
+        int statusCode = httpResponse.statusCode();
+
+        return statusCode == HttpStatus.NOT_FOUND.getCode() ||
```
Robbie-Microsoft: see other comments about creating array for retriable codes. You can ignore this if you'd like.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -574,6 +590,47 @@ void managedIdentityTest_Retry(ManagedIdentitySourceType source, String endpoint
...
+
+                //A 500 status code should trigger the exponential retry policy, with a total of 4 attempts (1 original + 3 retries)
+                verify(httpClientMock, times(4)).send(any());
+
+                return;
+            }
+
+            fail("MsalServiceException is expected but not thrown.");
+        }
+
```
Robbie-Microsoft, line 663: I think you should consider adding more tests. You can ignore this comment if these tests already exist:

- returns a 500 error response from the network request, just the first time
- returns a 500 error response from the network request permanently
- returns a 500 error response from the network request, just the first time, with a retry-after header of 3 seconds
- returns a 500 error response from the network request, just the first time, with a retry-after header of 3 seconds (extrapolated from an http-date)
- makes three acquireToken calls on the same managed identity application (which returns a 500 error response from the network request permanently) to ensure that retry policy lifetime is per request
- ensures that a retry does not happen when the http status code from a failed network response is not included in the retry policy
- ensures that a retry does not happen when the http status code from a failed network response is included in the retry policy, but the retry policy has been disabled

You can see these tests in msal-node here: https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/test/client/ManagedIdentitySources/DefaultManagedIdentityRetryPolicy.spec.ts

You should also consider running the more basic tests for both user-assigned and system-assigned, but you can stick with just system-assigned for the majority of them.

Avery-Dunn: Going through the list in order:
- Added a new test in the latest commit
- This is essentially the behavior of the existing tests, every HTTP call returns a 500
- `RequestThrottlingTest` has tests for this sort of throttling behavior. That behavior is separate from the new retry policies, and is the same for MI and non-MI scenarios
- This might be an SDK consistency item, currently MSAL Java only follows the retry-after header
- Added a new test in the latest commit
- Added a new test in the latest commit
- This might be an SDK consistency item, currently MSAL Java doesn't let the retry policy be disabled

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
Avery-Dunn: It's not as clean as node due to the (rather old) Java version that we support, however in the latest commit I did something similar: put all the appropriate codes in a static set, then just check if the set contains the response's status code.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+    public boolean isRetryable(IHttpResponse httpResponse) {
+        currentRetryCount++;
+        lastStatusCode = httpResponse.statusCode();
+
+        return HttpStatus.isServerError(lastStatusCode) || RETRYABLE_STATUS_CODES.contains(lastStatusCode);
+    }
+
+    @Override
+    public int getMaxRetryCount(IHttpResponse httpResponse) {
+        return (httpResponse.statusCode() == 410) ? LINEAR_RETRY_NUM : EXPONENTIAL_RETRY_NUM;
```
Robbie-Microsoft: Use HttpStatus.Gone.getCode() instead of 410?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    @Override
+    public int getRetryDelayMs(IHttpResponse httpResponse) {
+        // Use exponential backoff for non-410 status codes
+        if (lastStatusCode == 410) {
```
Robbie-Microsoft: Use HttpStatus.Gone.getCode() instead of 410?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    //Package-private methods to allow much quicker testing. The delay values should be treated as constants in any non-test scenario.
+    static void setRetryDelayMs(int retryDelayMs) {
+        LINEAR_RETRY_DELAY_MS = retryDelayMs;
+        EXPONENTIAL_RETRY_DELAY_MS = retryDelayMs;
+    }
+
+    static void resetToDefaults() {
+        LINEAR_RETRY_DELAY_MS = 10000;
```
Robbie-Microsoft: I would maybe define a non-static constant for 10000 (in this file) to use here and line 14 instead of using 10000 twice.

Robbie-Microsoft: Same comment for other places you do this as well

Robbie-Microsoft: What are your thoughts on this?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -574,6 +587,144 @@ void managedIdentityTest_Retry(ManagedIdentitySourceType source, String endpoint
...
+            DefaultHttpClientManagedIdentity httpClientMock = mock(DefaultHttpClientManagedIdentity.class);
+
+            miApp = ManagedIdentityApplication
+                    .builder(ManagedIdentityId.systemAssigned())
+                    .httpClient(httpClientMock)
+                    .build();
+
+            // IMDS has different retry logic for certain status codes, such as 410
+            when(httpClientMock.send(expectedRequest(IMDS, ManagedIdentityTestConstants.RESOURCE))).thenReturn(expectedResponse(410, ManagedIdentityTestConstants.MSI_ERROR_RESPONSE_500));
```
Robbie-Microsoft: Should you use HttpStatus.Gone.GetCode() here instead of 410 and other places in the tests where you are looking at http status codes?

Avery-Dunn: Fixed in the latest commit, here and elsewhere

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
Avery-Dunn: In the latest commit, the all-uppercase variables are now true constants, and there are new variables to represent the 'current' values being used by the policy.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
Avery-Dunn: Fixed in the latest commit, here and elsewhere

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
Avery-Dunn: Fixed in the latest commit, here and elsewhere

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
```diff
@@ -0,0 +1,66 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+//IMDS uses a different try policy than other MI flows, see https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/imds_retry_based_on_errors.md
+class IMDSRetryPolicy extends ManagedIdentityRetryPolicy {
+    private static final int LINEAR_RETRY_NUM = 7;
+    private static final int LINEAR_RETRY_DELAY_MS = 10000; // 10 seconds
+    private static final int EXPONENTIAL_RETRY_NUM = 3;
+    private static final int EXPONENTIAL_RETRY_DELAY_MS = 1000; // 1 second
```
Robbie-Microsoft, line 16: I don't think these should be static anymore since they're no longer referenced outside of this file

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,28 @@
...
+// Licensed under the MIT License.
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * Default retry policy for most MSAL Java flows
+ */
+class DefaultRetryPolicy implements IRetryPolicy {
+    private static final int RETRY_NUM = 1;
+    private static final int RETRY_DELAY_MS = 1000;
```
Robbie-Microsoft, line 11: should these be static?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -18,18 +18,18 @@ class HttpHelper implements IHttpHelper {
 
     private static final Logger log = LoggerFactory.getLogger(HttpHelper.class);
     public static final String RETRY_AFTER_HEADER = "Retry-After";
-    public static final int RETRY_NUM = 2;
-    public static final int RETRY_DELAY_MS = 1000;
 
     public static final int HTTP_STATUS_200 = 200;
     public static final int HTTP_STATUS_400 = 400;
     public static final int HTTP_STATUS_429 = 429;
     public static final int HTTP_STATUS_500 = 500;
```
Robbie-Microsoft: Can these be deleted since you have the HttpStatus enum now? Or maybe perhaps you can make another PR for this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -0,0 +1,38 @@
...
+        return code;
+    }
+
+    String getDescription() {
+        return description;
+    }
+
+    //All 5xx errors
+    static boolean isServerError(int code) {
+        return code >= 500 && code < 600;
```
Robbie-Microsoft, line 27: very nitpicky, but should you reference these numbers via the HttpStatus enum in this file?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRetryPolicy.java
```diff
@@ -0,0 +1,54 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Retry policy for most Managed Identity scenarios
+ */
+class ManagedIdentityRetryPolicy implements IRetryPolicy {
+    private static final int RETRY_NUM = 3;
+    private static final int RETRY_DELAY_MS = 1000;
```
Robbie-Microsoft, line 16: should these be static?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
```diff
@@ -0,0 +1,28 @@
...
+    @Override
+    public int getMaxRetryCount(IHttpResponse httpResponse) {
+        return RETRY_NUM;
+    }
+
+    @Override
+    public int getRetryDelayMs(IHttpResponse httpResponse) {
+        return RETRY_DELAY_MS;
+    }
+}
```
Robbie-Microsoft, line 28: Do you need setRetryDelayMs and resetToDefaults here for testing purposes?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
Robbie-Microsoft: Nevermind, I see your other PR now that refactors Http Status constants

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Robbie-Microsoft: moving this feedback to your Http Status PR

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityRetryPolicy.java
Avery-Dunn: See https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960#discussion_r2126919833

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
Avery-Dunn: To keep this PR more focused I didn't want to refactor tests unrelated to the main managed identity/IMDS behavior changes, so other tests that now use the DefaultRetryPolicy don't need to set/reset the delays

(there's only a few and they have ad-hoc ways of running faster, if those tests get refactored in the future we can add the new methods in)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultRetryPolicy.java
Avery-Dunn: See https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960#discussion_r2126919833

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMDSRetryPolicy.java
Avery-Dunn: It's bad practice to static variables that can change, but good practice to have static constants that never change:
- `private final` means they are associated with an instance and would be initialized every time a RetryPolicy is created, `private static final` means they are initialized once and used for every RetryPolicy for (more-or-less) the lifetime of the application
- It's a pretty minor boost in the grand scheme of things, but since they will never change and will be used by every instance of RetryPolicy there's no need to initialize them every time.

Robbie-Microsoft (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IRetryPolicy.java
```diff
@@ -0,0 +1,25 @@
+package com.microsoft.aad.msal4j;
```
bgavrilMS, line 1: suggestion: All this HTTP and Retry files should go into a folder of their own?

Avery-Dunn: I agree that it would be good to group similar files together, but currently all of the source files are just in one big folder. I don't know why it was originally designed that way (the people who would did it are no longer here), and I've always felt it was a bad practice.

I've created an issue to track a refactor of the directory structure, at the very least there's quick value in properly organizing the non-public classes: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/965

Robbie-Microsoft (approved):

PR #963, opened 2025-06-04
Add API to disable internal retry behavior

Avery-Dunn:
Adds a public API to `AbstractApplicationBase` to disable internal retry behavior for HTTP requests, similar to what was done in MSAL Node and .NET:
https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/7603
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5252

Internally, `HttpHelper` was given a new constructor more easily and cleanly use this new field and the new `IRetryPolicy` introduce in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960, and some minor refactoring was done to related classes.

`ManagedIdentityTests` received a new parameterized test which uses the API with several `IRetryPolicy` implementations.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
@@ -32,6 +32,7 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
     private IHttpClient httpClient;
     private Integer connectTimeoutForDefaultHttpClient;
     private Integer readTimeoutForDefaultHttpClient;
+    private boolean disableInternalRetries;
     String tenant;
 
     //The following fields are set in only some applications and/or set internally by the library. To avoid excessive
@@ -150,6 +151,10 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
         return this.readTimeoutForDefaultHttpClient;
     }
 
+    boolean retriesDisabled() {
+        return this.disableInternalRetries;
+    }
+
     String tenant() {
         return this.tenant;
     }
@@ -190,6 +195,7 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
         Boolean onlySendFailureTelemetry = false;
         Integer connectTimeoutForDefaultHttpClient;
         Integer readTimeoutForDefaultHttpClient;
+        boolean disableInternalRetries;
         private String clientId;
         private Authority authenticationAuthority = createDefaultAADAuthority();
 
@@ -319,6 +325,17 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
             return self();
         }
 
+        /**
+         * Disables the library's internal retry logic for HTTP requests. Used alongside
+         *
+         * @param val timeout value in milliseconds
+         * @return instance of the Builder on which method was called
+         */
+        public T disableInternalRetries(boolean val) {
+            disableInternalRetries = val;
+            return self();
+        }
+
         T telemetryConsumer(Consumer<List<HashMap<String, String>>> val) {
             validateNotNull("telemetryConsumer", val);
 
@@ -356,5 +373,16 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
         readTimeoutForDefaultHttpClient = builder.readTimeoutForDefaultHttpClient;
         authenticationAuthority = builder.authenticationAuthority;
         clientId = builder.clientId;
+        disableInternalRetries = builder.disableInternalRetries;
+
+        if (builder.httpClient == null) {
+            httpClient = new DefaultHttpClient(
+                    builder.proxy,
+                    builder.sslSocketFactory,
+                    builder.connectTimeoutForDefaultHttpClient,
+                    builder.readTimeoutForDefaultHttpClient);
+        } else {
+            httpClient = builder.httpClient;
+        }
     }
 }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
@@ -326,13 +326,14 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
         }
 
         /**
-         * Disables the library's internal retry logic for HTTP requests. Used alongside
+         * The library has a number of policies for retrying HTTP calls in different scenarios.
+         * <p>
+         * This will disable all internal retry behavior, allowing customized retry behavior via your own implementation of {@link IHttpClient}
          *
-         * @param val timeout value in milliseconds
          * @return instance of the Builder on which method was called
          */
-        public T disableInternalRetries(boolean val) {
-            disableInternalRetries = val;
+        public T disableInternalRetries() {
+            disableInternalRetries = true;
             return self();
         }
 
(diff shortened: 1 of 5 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -319,6 +325,17 @@ public T readTimeoutForDefaultHttpClient(Integer val) {
...
+        /**
+         * Disables the library's internal retry logic for HTTP requests. Used alongside
+         *
+         * @param val timeout value in milliseconds
+         * @return instance of the Builder on which method was called
+         */
+        public T disableInternalRetries(boolean val) {
+            disableInternalRetries = val;
+            return self();
+        }
```
Robbie-Microsoft: I don't think this should accept a param? It's only function should be to disable internal retries, and that should happen once on the initial application configuration. I think this should just set disableInternalRetries to`true`.

Robbie-Microsoft: Can you pass it in through the application config instead?

Avery-Dunn: > Can you pass it in through the application config instead?

After looking into MSAL.NET a bit, it seems like they have an `ApplicationConfiguration` file that's separate from the builder, which in turn is separate from the application types it builds? MSAL Java doesn't have that sort of structure, all of the fields/helper methods shared by the different application types are in this `AbstractApplicationBase`, and the builders are within the classes they build.

> I don't think this should accept a param?

Good call, coincidentally pretty much every single builder method we currently have has at least one parameter, but that's not required. In the latest commit the method now just sets a value of 'true' rather than take in a parameter.

Robbie-Microsoft: I was thinking in terms of [msal-node's implementation](https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/dev/lib/msal-node/src/config/Configuration.ts#L102) - the user passes in a config with `disableInternalRetries` set to true, otherwise it's undefined. But if Java doesn't/can't do this, then your solution is good.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -32,6 +32,7 @@ public abstract class AbstractApplicationBase implements IApplicationBase {
     private IHttpClient httpClient;
     private Integer connectTimeoutForDefaultHttpClient;
     private Integer readTimeoutForDefaultHttpClient;
+    private boolean disableInternalRetries;
```
Robbie-Microsoft: should this be set to `false` initially?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -26,12 +26,19 @@ class HttpHelper implements IHttpHelper {
 
     private IHttpClient httpClient;
     private IRetryPolicy retryPolicy;
+    private boolean retriesDisabled;
```
Robbie-Microsoft: initially set to false?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -150,6 +151,10 @@ public Integer readTimeoutForDefaultHttpClient() {
         return this.readTimeoutForDefaultHttpClient;
     }
 
+    boolean retriesDisabled() {
```
Robbie-Microsoft: nit: I would rename this to "areRetriesDisabled" to indicate we are expecting a "yes" or "no" answer. Or maybe "isRetryDisabled".

You can ignore this feedback and resolve this conversation if you'd like.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
```diff
@@ -319,6 +325,18 @@ public T readTimeoutForDefaultHttpClient(Integer val) {
...
         }
 
+        /**
+         * The library has a number of policies for retrying HTTP calls in different scenarios.
+         * <p>
+         * This will disable all internal retry behavior, allowing customized retry behavior via your own implementation of {@link IHttpClient}
+         *
+         * @return instance of the Builder on which method was called
+         */
+        public T disableInternalRetries() {
```
neha-bhargava, line 335: Maybe rename it to disableMsalRetries. In MSAL.Net this is passed as a flag when passing HttpClientFactory. Would recommend to be consistent with that if possible. [WithHttpClientFactory](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9fd9500cca0125250ad52668793b6bc778a6b05f/src/client/Microsoft.Identity.Client/AppConfig/BaseAbstractApplicationBuilder.cs#L82). The name of the flag is no longer valid in MSAL.Net but similar.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
neha-bhargava: By default the boolean variables are false. So should be fine.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
neha-bhargava: +1 to this

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Can confirm, it is false by default.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
Avery-Dunn: Yep, booleans are initialized as false by default.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: This is the naming that @Robbie-Microsoft did for Node and .NET:
https://github.com/AzureAD/microsoft-authentication-library-for-js/blob/413ea5d4514ff409af2def53af1b4774c7343e11/lib/msal-node/src/config/Configuration.ts#L102
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/9fd9500cca0125250ad52668793b6bc778a6b05f/src/client/Microsoft.Identity.Client/AppConfig/ApplicationConfiguration.cs#L119

MSAL Java doesn't really have a separate configuration file like Node and .NET seem to, so this API would be the closest equivalent.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Good call. Couldn't find much official guidance, but I believe the naming conventions are: 
- A boolean variable should just be the name ("retryDisabled")
- Methods that return booleans should be the name with is/can/should/etc. at the start ("isRetryDisabled")

In the latest commit the method is now `isRetryDisabled`, and the field is `retryDisabled`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AbstractApplicationBase.java
Avery-Dunn: Yeah, seems like msal-node and .NET both have a configuration file separate from the builder, but for MSAL Java it's all in the same place.

(there's no technical/Java reason why we couldn't also split it up, it's just not how the public API was originally designed)

bgavrilMS (approved):

PR #964, opened 2025-06-04
Consistently use constants for HTTP status codes

Avery-Dunn:
The changes in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960 made it clear that the library is inconsistent when it comes to HTTP status codes: not all codes had a constant, and even when there was a constant it often wasn't used.

This PR adjusts the `HttpStatus` enum introduced in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/960 to be a simpler class of constants and helper methods, and refactors every status code reference to use these constants.

This is a separate PR in order to be much easier to review, each individual change is minor and they should not affect any existing behavior.

The status code constants follow the style in HttpURLConnection, a standard Java package that also has these sorts of constants: https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html

However, I didn't want to rely on that since we only need a small subset and none of the other code, plus this allows us to add custom helper methods (like the isServerError() that checks for 5xx errors)

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
@@ -3,35 +3,26 @@
 
 package com.microsoft.aad.msal4j;
 
-enum HttpStatus {
-    OK(200, "OK"),
-    FOUND(302, "Found"),
-    BAD_REQUEST(400, "Bad Request"),
-    NOT_FOUND(404, "Not Found"),
-    REQUEST_TIMEOUT(408, "Request Timeout"),
-    GONE(410, "Gone"),
-    TOO_MANY_REQUESTS(429, "Too Many Requests"),
-    INTERNAL_SERVER_ERROR(500, "Internal Server Error"),
-    SERVICE_UNAVAILABLE(503, "Service Unavailable"),
-    GATEWAY_TIMEOUT(504, "Gateway Timeout");
+class HttpStatus {
 
-    private final int code;
-    private final String description;
+    static final int HTTP_OK = 200;
+    static final int HTTP_FOUND = 302;
+    static final int HTTP_BAD_REQUEST = 400;
+    static final int HTTP_UNAUTHORIZED = 401;
+    static final int HTTP_NOT_FOUND = 404;
+    static final int HTTP_REQUEST_TIMEOUT = 408;
+    static final int HTTP_GONE = 410;
+    static final int HTTP_TOO_MANY_REQUESTS = 429;
+    static final int HTTP_INTERNAL_ERROR = 500;
+    static final int HTTP_UNAVAILABLE = 503;
+    static final int HTTP_GATEWAY_TIMEOUT = 504;
 
-    HttpStatus(int code, String description) {
-        this.code = code;
-        this.description = description;
-    }
-
-    int getCode() {
-        return code;
-    }
-
-    String getDescription() {
-        return description;
-    }
-
-    //All 5xx errors
+    /**
+     * Determines if the status code represents a server error (5xx).
+     *
+     * @param code The HTTP status code
+     * @return true if the status code is between 500 and 599, inclusive
+     */
     static boolean isServerError(int code) {
         return code >= 500 && code < 600;
     }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -235,8 +235,8 @@ class AadInstanceDiscoveryProvider {
 
         AadInstanceDiscoveryResponse response = JsonHelper.convertJsonStringToJsonSerializableObject(httpResponse.body(), AadInstanceDiscoveryResponse::fromJson);
 
-        if (httpResponse.statusCode() != HttpHelper.HTTP_STATUS_200) {
-            if (httpResponse.statusCode() == HttpHelper.HTTP_STATUS_400 && response.error().equals("invalid_instance")) {
+        if (httpResponse.statusCode() != HttpStatus.HTTP_OK) {
+            if (httpResponse.statusCode() == HttpStatus.HTTP_BAD_REQUEST && response.error().equals("invalid_instance")) {
                 // instance discovery failed due to an invalid authority, throw an exception.
                 throw MsalServiceExceptionFactory.fromHttpResponse(httpResponse);
             }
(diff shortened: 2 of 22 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -3,35 +3,26 @@
...
-
-    //All 5xx errors
+    /**
+     * Determines if the status code represents a server error (5xx).
+     *
+     * @param code The HTTP status code
+     * @return true if the status code is between 500 and 599, inclusive
+     */
     static boolean isServerError(int code) {
         return code >= 500 && code < 600;
```
Robbie-Microsoft, line 27: Can you reference these numbers via the constants you just defined?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -3,35 +3,26 @@
...
-    GONE(410, "Gone"),
-    TOO_MANY_REQUESTS(429, "Too Many Requests"),
-    INTERNAL_SERVER_ERROR(500, "Internal Server Error"),
-    SERVICE_UNAVAILABLE(503, "Service Unavailable"),
-    GATEWAY_TIMEOUT(504, "Gateway Timeout");
+class HttpStatus {
 
-    private final int code;
-    private final String description;
+    static final int HTTP_OK = 200;
```
Robbie-Microsoft, line 8: nit: Since you're referencing these via `HttpStatus.xxx`, I would delete "HTTP" from each constant name.

`HttpStatus.HTTP_BAD_REQUEST` seems redundant. I would prefer `HttpStatus.BAD_REQUEST`.

It's up to you though, that's just my perspective.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -3,35 +3,26 @@
...
+    static final int HTTP_FOUND = 302;
```
Robbie-Microsoft, line 9: REDIRECT

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -3,35 +3,26 @@
...
+    static final int HTTP_BAD_REQUEST = 400;
+    static final int HTTP_UNAUTHORIZED = 401;
+    static final int HTTP_NOT_FOUND = 404;
+    static final int HTTP_REQUEST_TIMEOUT = 408;
+    static final int HTTP_GONE = 410;
+    static final int HTTP_TOO_MANY_REQUESTS = 429;
+    static final int HTTP_INTERNAL_ERROR = 500;
+    static final int HTTP_UNAVAILABLE = 503;
```
Robbie-Microsoft, line 17: SERVICE_UNAVAILABLE

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
```diff
@@ -3,35 +3,26 @@
...
-    private final String description;
+    static final int HTTP_OK = 200;
+    static final int HTTP_FOUND = 302;
+    static final int HTTP_BAD_REQUEST = 400;
+    static final int HTTP_UNAUTHORIZED = 401;
+    static final int HTTP_NOT_FOUND = 404;
+    static final int HTTP_REQUEST_TIMEOUT = 408;
+    static final int HTTP_GONE = 410;
+    static final int HTTP_TOO_MANY_REQUESTS = 429;
+    static final int HTTP_INTERNAL_ERROR = 500;
```
Robbie-Microsoft, line 16: INTERNAL_SERVER_ERROR

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Avery-Dunn: This is just following the style in `HttpURLConnection`, a standard Java package that also has these sorts of constants: https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html

(thought I didn't want to rely on that since we only need a small subset and none of the other code)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Avery-Dunn: I would've, but I felt it would be clearer to leave these as numbers:
- It wouldn't make sense to have a 600 HTTP status code, and `code >= HTTP_INTERNAL_ERROR && code < 600` looked weird to me
- `>= 500` makes it obvious that the range starts at 500, but `=> HTTP_INTERNAL_ERROR` would make you have to check the constant to be sure

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Avery-Dunn: It was 'redirect' in the 1.0 HTTP spec, but it looks like the wording has been 'found' since 1.1 in 1999: 
https://www.rfc-editor.org/rfc/rfc2616#section-10.3.3
https://en.wikipedia.org/wiki/HTTP_302
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/302

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Avery-Dunn: Updated the PR description and mentioned in another comment, I'm just following the style in `HttpURLConnection`: https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Avery-Dunn: Updated the PR description and mentioned in another comment, I'm just following the style in `HttpURLConnection`: https://docs.oracle.com/javase/8/docs/api/java/net/HttpURLConnection.html

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Robbie-Microsoft: Well maybe you could add a code for 600 then :) I added one in msal-node just for the one check that you have here.

`MULTI_SIDED_ERROR: 600,`

It's your choice though. You can resolve this conversation.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Robbie-Microsoft: thanks for sharing that

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpStatus.java
Robbie-Microsoft: ok, you can disregard if you'd like

Robbie-Microsoft (approved):

PR #968, opened 2025-06-06
PoC: Loading test config from a central service

Avery-Dunn:
A draft PR in MSAL Python introduced a new concept of how to set up automated testing: https://github.com/AzureAD/microsoft-authentication-library-for-python/pull/825

Essentially, configuration for how to set up constants, applications, token calls, assertions, and more is all stored in a central service that every MSAL can access, allowing us to create new tests in every library simply by creating new configuration file that is processed by generic "runners" in each MSAL.

This PR achieves this by introducing a few new classes:
- `RunnerTest` : Essentially just unit tests that retrieve configuration files and executes helper methods based on the contents
  - Currently just one unit test which can process config files that cover managed identity behavior
- `RunnerHelper` : A set of helper methods to create MSAL objects based on the contents of a configuration file
  - Currently just makes the objects needed for the one test, but is easily expandable as we add more
- `RunnerJsonHelper` : A set of helper methods to parse the configuration file 
  - For simplicity, it is pretty specific to the current config structure, and in a real implementation it would be more generic
- `Shortcuts` : As the name implies, some shortcuts that were taken just to make this POC quicker and easier
  - For example: simple internal classes to represent JSON objects until the config structure is finalized
  
For this first-draft review:
- `RunnerTest` and `RunnerHelper` are pretty representative of what a final implementation would look like, as they mostly just need to be expanded to cover more scenarios
- `RunnerJsonHelper` will likely be replaced by a more generic version that can handle any config file and reuse behavior in the library's existing `JsonHelper`
- `Shortcuts` an essentially be ignored during a review, since none of it would be part of a 'real' implementation
- The names of classes, methods, key-value pairs, and even the structure of the config file can all change as needed since this is not meant to be a public API, so no need to focus much on that

### DIFF
--- a/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/RunnerTest.java
+++ b/msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/RunnerTest.java
@@ -1,27 +1,57 @@
 package com.microsoft.aad.msal4j;
 
-import org.junit.jupiter.api.Test;
-
+import com.fasterxml.jackson.databind.JsonNode;
 import com.microsoft.aad.msal4j.Shortcuts.TestConfig;
 import com.microsoft.aad.msal4j.Shortcuts.TestStep;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.util.Map;
+import java.util.stream.Stream;
 
 class RunnerTest {
+    private static final Logger LOG = LoggerFactory.getLogger(RunnerTest.class);
 
-    @Test
-    void testManagedIdentityWithJsonConfig() throws Exception {
-        //TODO: get test cases list from the server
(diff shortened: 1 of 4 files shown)


msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/RunnerTest.java
```diff
@@ -0,0 +1,59 @@
...
+        );
+    }
+
+    @ParameterizedTest
+    @MethodSource("managedIdentityTestsProvider")
+    void runManagedIdentityTest(String testCaseName) throws Exception {
+        LOG.info("==========Executing Test Case==========");
+
+        // Get all test configurations
+        Map<String, JsonNode> configs = RunnerHelper.getAllTestCaseConfigs("https://smile-test.azurewebsites.net/testcases.json");
```
bgavrilMS, line 34: This needs to be local or hosted in the lab.

Avery-Dunn:
Closing, a more finalized version of this POC has been made in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/970

Issue #971, opened 2025-07-08
Ensure all public APIs in the msal4j package are covered by Javadoc comments


Avery-Dunn:
Although most of our public APIs are documented, there are clearly gaps in many classes and methods: https://javadoc.io/doc/com.microsoft.azure/msal4j/1.21.0/index.html

To follow Java best-practices and better help our customers, all APIs should be covered by Javadoc comments. 

To keep the PRs focused and manageable, this issue only applies to the com.microsoft.aad.msal4j package. The other packages in this repo (the broker at `com.microsoft.aad.msal4jbrokers` and persistence extensions at `com.microsoft.aad.msal4jextensions`) will be covered elsewhere.

PR #973, opened 2025-07-14
Add/improve javadocs for interface classes

Avery-Dunn:
Begins the work described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/971

This PR fills documentation gaps and improves existing documentation to all interfaces in the library. The changes exclusively affect Javadoc comments, so there are no changes to the actual behavior of any API.

This PR is also a trial run of the new agent feature in IntelliJ's Copilot plugin:
- All of the changes were manually reviewed to ensure MSAL behavior was described accurately
- With few exceptions, the prompts produced accurate descriptions on the first try without being given any context about MSAL's actual behavior
  - For the most part, I simply told it to review all interfaces with missing/poor documentation, list them based on quality, then told it to add/improve a few filenames at a time
- The few mistakes it did make were easily corrected by a follow-up prompt describing what was wrong
  - As an example, it incorrectly implied that "scopes" API was just for Graph resources (likely because many of our tests use Graph), and I just had to tell it to give a more generic description

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAcquireTokenParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAcquireTokenParameters.java
@@ -7,16 +7,61 @@ import java.util.Map;
 import java.util.Set;
 
 /**
- * Parameters shared by all acquireToken methods
+ * Interface representing common parameters shared across all token acquisition methods in MSAL.
+ * <p>
+ * This interface defines the core parameters that are common to all token acquisition operations,
+ * regardless of the specific authentication flow being used (interactive, silent, username/password, etc.).
+ * Various parameter classes in the library implement this interface to provide a consistent way to
+ * configure token requests.
  */
 interface IAcquireTokenParameters {
+
+    /**
+     * Gets the set of scopes (permissions) requested for the access token.
+     * <p>
+     * Scopes represent the permissions that the application is requesting access to.
+     *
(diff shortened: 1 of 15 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAcquireTokenParameters.java
```diff
@@ -7,16 +7,61 @@
...
 /**
- * Parameters shared by all acquireToken methods
+ * Interface representing common parameters shared across all token acquisition methods in MSAL.
+ * <p>
+ * This interface defines the core parameters that are common to all token acquisition operations,
+ * regardless of the specific authentication flow being used (interactive, silent, username/password, etc.).
+ * Various parameter classes in the library implement this interface to provide a consistent way to
+ * configure token requests.
  */
 interface IAcquireTokenParameters {
```
bgavrilMS, line 17: nit: strange that you went with interface for this. new field -> broken interface.

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IAcquireTokenParameters.java
Avery-Dunn: It was before my time here so I don't know why it's an interface.

But it's not public, so if we did add something it wouldn't be a breaking change for customers since they could only see our implementations anyway (it would be the same as adding a new public API to any class).

PR #982, opened 2025-08-12
Remove third-party dependencies from msal4j package

Avery-Dunn:
This PR represents the work done for https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/909, and all of the changes were already approved in the various PRs linked to that issue.

As can be seen on Maven, once merged the msal4j package will no longer be reliant on any package except for Azure SDK's [azure-json](https://mvnrepository.com/artifact/com.azure/azure-json) and the very popular logging framework [slf4j-api](https://mvnrepository.com/artifact/org.slf4j/slf4j-api): https://mvnrepository.com/artifact/com.microsoft.azure/msal4j/1.30.0-beta

All of the behavior from other dependencies were replaced with our own implementations. Despite the large number of changes, as found in our own testing and testing by Azure SDK there are only two groups known breaking changes:
- Removal of protected APIs that returned or used `ClientAuthentication`, which was a class from com.nimbusds
  - We could not remove this package without removing this reference so a breaking change was unavoidable, however - Refactor JSON parsing to return a more appropriate `MsalClientException` or `MsalServiceException` when a JSON was blank, rather than a `JsonParsingException` or similar
  - No API promised JSON parse exceptions and the MSAL exceptions have more details about the underlying issue
  
Normally breaking changes to public/protected APIs would warrant a major version bump, however these changes should affect few customers (if any) and were closer to bugfixes than changes to the overall design as they should never have been publicly available:
 - The return values of APIs that referenced `ClientAuthentication` were never used by any other public msal4j API, and we could couldn't think of any scenario where another project would need the instance we produced
 - They were protected methods in very fundamental MSAL classes that customers are unlikely to extend, so few would have even known about them


changelog.txt
```diff
@@ -1,3 +1,9 @@
+Version 1.30.0-beta
+=============
+- Replace org.projectlombok dependencies with implementations of generated code (#946)
```
g2vinay, line 3: We should mention about breaking changes from exception changes

Avery-Dunn: Normally we just make a separate changelog/version update PR once everything is merged to clearly mark the cutoff for changes that we release, so I didn't include them here

But this is kind of a special case due to it already being released. I just set up this PR to make it easier to review the changelog, and will either merge into this branch or the main branch depending on which gets approved first: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/983

bgavrilMS (approved):

neha-bhargava (approved):

Issue #984, opened 2025-08-29
[Bug] client assetion is not within valid time range.


sudoapt-getclean:
### Library version used

v1.23.0

### Java version

23

### Scenario

ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

The app is in production, and I have upgraded to a new version of MSAL

### Issue description and reproduction steps

Hi,
1 hour afte we upgrade to msal v1.23.0 we started seeing this stacktrace:
```
Caused by: com.microsoft.aad.msal4j.MsalServiceException: AADSTS700024: Client assertion is not within its valid time range. Current time: 2025-08-29T05:47:32.1501147Z, assertion valid from 1970-01-01T00:00:00.0000000Z, expiry time of assertion 2025-08-29T00:30:05.0000000Z. Review the documentation at https://learn.microsoft.com/entra/identity-platform/certificate-credentials . Trace ID: 7a84b74b-1889-494f-b408-66e8b36c7400 Correlation ID: d75bd923-d9ca-4d18-b80e-b13851840fa9 Timestamp: 2025-08-29 05:47:32Z
	at com.microsoft.aad.msal4j.MsalServiceExceptionFactory.fromHttpResponse(MsalServiceExceptionFactory.java:38)
	at com.microsoft.aad.msal4j.TokenRequestExecutor.createAuthenticationResultFromOauthHttpResponse(TokenRequestExecutor.java:168)
	at com.microsoft.aad.msal4j.TokenRequestExecutor.executeTokenRequest(TokenRequestExecutor.java:35)
	at com.microsoft.aad.msal4j.AbstractApplicationBase.acquireTokenCommon(AbstractApplicationBase.java:75)
	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.execute(AcquireTokenByAuthorizationGrantSupplier.java:60)
	at com.microsoft.aad.msal4j.AcquireTokenByClientCredentialSupplier.acquireTokenByClientCredential(AcquireTokenByClientCredentialSupplier.java:87)
	at com.microsoft.aad.msal4j.AcquireTokenByClientCredentialSupplier.execute(AcquireTokenByClientCredentialSupplier.java:50)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:69)
[...]

Issue #977, opened 2025-07-28
[Feature Request]     Support refreshing Client Assertion


CodyPolton:
### MSAL client type

Confidential

### Problem Statement

Our company is starting to look at using SPIFFE/SPIRE to generate short lived tokens that we are using as our client assertion. We are currently needing to write our own implementation to retrieve a new Token when the original client assertion expires. 

### Proposed solution

It looks like this functionality has been added to the [dotnet version](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/releases/tag/4.25.0) as the WithClientAssertion function. Would be useful to have something similar as a callback function whenever the client assertion expires. 

### Alternatives

Currently we have had to move away from the msal4j library and handling making call to generate a token and caching it as well as handling when the client assertion expires, but seems like this functionality would be useful in the msal4j library.

CodyPolton:
Any timeline for this enhancement?

Avery-Dunn:
Sorry for the lack of timeline, however version 1.23.1 has been released containing a fix for this issue: https://github.com/AzureAD/microsoft-authentication-library-for-java/releases/tag/v1.23.1

[`ClientCredentialFactory.createFromCallback()`](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/dev/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialFactory.java#L101) was intended to match MSAL .NET's behavior of allowing a callback method to provide the assertion. A bug prevented the callback from being called more then once we configured at the app-level, but you could still set the assertion on each request when using [`ClientCredentialParameters.clientCredential()`](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/b8c025bacb28a6ad8e997f3d168d6d77bfed4325/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java#L154)

The changes in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/988 should resolve this problem by fixing the bug and generally improving assertion behavior, so that the callback is called every time the assertion is needed. Let use know if you still have any issues after updating to 1.23.1

Issue #879, opened 2024-11-21
[Feature Request] Properly asynchronous client assertions


q-benwillis:
### MSAL client type

Confidential

### Problem Statement

We use a client assertion as our client credential when building our confidential client application. This assertion comes from a file stored locally and gets refreshed every hour. At the moment the credential can only be created before the confidential client application is created, even when using the `createFromCallback` method meaning we need to create a whole new confidential client application everytime the token is refreshed.

### Proposed solution

Ideally we would like to build one confidential client application that refreshes this client credential on every acquire token request.

I can see that we can create a client credential from a callback using `createFromCallback` but this still gets executed when the confidential client is created. Ideally this callback should be executed when the `acquireToken` method is called.

### Alternatives

At the moment we need to either create a new confidential client for every request or somehow build a new confidential client every time the token is updated.

pdefreitas:
+1 for this one. The .NET library provides a builder with WithClientAssertion(<Func>). The builder with plain string is obsolete. Node has also implemented async callbacks:
https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/7014

Avery-Dunn:
Sorry for the slow response, however version 1.23.1 has been released containing a fix for this issue: https://github.com/AzureAD/microsoft-authentication-library-for-java/releases/tag/v1.23.1

The bug you mentioned prevented the callback from being called more then once we configured at the app-level, but you may have still be able to set the assertion on each request when using [`ClientCredentialParameters.clientCredential()`](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/b8c025bacb28a6ad8e997f3d168d6d77bfed4325/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java#L154)

The changes in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/988 should resolve this problem by fixing the bug and generally improving assertion behavior, so that the callback is called every time the assertion is needed. Let use know if you still have any issues after updating to 1.23.1

PR #986, opened 2025-09-02
Refactor client credential behavior to be per-request

Avery-Dunn:
Refactors behavior related to client credentials and also fixes the regression described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/984

Version 1.23.0 introduced a large refactor to remove usage of third-party dependencies, in particular those from `com.nimbusds`. Prior to 1.23.0 assertions created for `ClientCertificate` were given 10 minute expiry time and were only recreated if the old one expired, however the refresh behavior was lost during a refactor in PR https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/927

In addition, this PR fixes/improves assertion behavior to resolve the issues described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/879 and https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/977: when using the callback option for assertions the original design only called the callback when the `ClientAssertion` instance was first created, meaning an app-level assertion was only created once

This PR fixes that regression and refactors the behavior around client credentials to be more in line with other MSALs:
- Remove locally cached credential fields from `ConfidentialClientApplication`, and move logic for creating assertion from `ConfidentialClientApplication` to `ClientCertificate`
- Adjusts `ClientAssertion` and `ClientCredentialFactory` to properly call callbacks set at the app-level
- Refactor certain methods in `TokenRequestExecutor` to use credential fields/methods directly, properly creates assertions when tenants are set at the request-level, and generally improve some complicated logic
- Adjusts `ClientAssertion` and `ClientCredentialFactory` to properly call the app-level callback whenever the assertion is needed
- Update and expand tests in `ClientCertificateTest` and `ClientCredentialTest` to cover the new behavior

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -131,6 +131,40 @@ abstract class Authority {
         }
     }
 
+    /**
+     * Creates a new Authority instance with a different tenant.
+     * This is useful when overriding the tenant for a specific request.
+     *
+     * @param originalAuthority The original authority to base the new one on
+     * @param newTenant The new tenant to use in the authority URL
+     * @return A new Authority instance with the specified tenant
+     * @throws MalformedURLException If the new authority URL is invalid
+     * @throws NullPointerException If originalAuthority or newTenant is null
+     */
+    static Authority replaceTenant(Authority originalAuthority, String newTenant) throws MalformedURLException {
+        if (originalAuthority == null) {
+            throw new NullPointerException("originalAuthority");
+        }
+        if (StringHelper.isBlank(newTenant)) {
+            throw new NullPointerException("newTenant");
+        }
+
+        URL originalUrl = originalAuthority.canonicalAuthorityUrl();
+        String host = originalUrl.getHost();
+        String protocol = originalUrl.getProtocol();
+        int port = originalUrl.getPort();
+
+        // Build path with new tenant
+        String newAuthority = String.format("%s://%s%s/%s/",
+                protocol,
+                host,
+                (port == -1 ? "" : ":" + port),
+                newTenant);
+
+        // Create proper authority instance with the tenant-specific URL
+        return createAuthority(new URL(newAuthority));
+    }
+
     static String getTenant(URL authorityUrl, AuthorityType authorityType) {
         String[] segments = authorityUrl.getPath().substring(1).split("/");
         if (authorityType == AuthorityType.B2C) {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -133,7 +133,7 @@ abstract class Authority {
 
     /**
      * Creates a new Authority instance with a different tenant.
-     * This is useful when overriding the tenant for a specific request.
+     * This is useful when overriding the tenant at request level.
      *
      * @param originalAuthority The original authority to base the new one on
      * @param newTenant The new tenant to use in the authority URL
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -138,31 +138,12 @@ abstract class Authority {
      * @param originalAuthority The original authority to base the new one on
      * @param newTenant The new tenant to use in the authority URL
      * @return A new Authority instance with the specified tenant
-     * @throws MalformedURLException If the new authority URL is invalid
-     * @throws NullPointerException If originalAuthority or newTenant is null
      */
     static Authority replaceTenant(Authority originalAuthority, String newTenant) throws MalformedURLException {
-        if (originalAuthority == null) {
-            throw new NullPointerException("originalAuthority");
-        }
-        if (StringHelper.isBlank(newTenant)) {
-            throw new NullPointerException("newTenant");
-        }
+        String authorityString = originalAuthority.canonicalAuthorityUrl().toString();
+        authorityString = authorityString.replace(originalAuthority.tenant, newTenant);
 
-        URL originalUrl = originalAuthority.canonicalAuthorityUrl();
-        String host = originalUrl.getHost();
-        String protocol = originalUrl.getProtocol();
-        int port = originalUrl.getPort();
-
-        // Build path with new tenant
-        String newAuthority = String.format("%s://%s%s/%s/",
-                protocol,
-                host,
-                (port == -1 ? "" : ":" + port),
-                newTenant);
-
-        // Create proper authority instance with the tenant-specific URL
-        return createAuthority(new URL(newAuthority));
+        return createAuthority(new URL(authorityString));
     }
 
     static String getTenant(URL authorityUrl, AuthorityType authorityType) {
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
@@ -54,11 +54,16 @@ final class ClientAssertion implements IClientAssertion {
         if (assertionProvider != null) {
             try {
                 String generatedAssertion = assertionProvider.call();
+
                 if (StringHelper.isBlank(generatedAssertion)) {
-                    throw new MsalClientException("Assertion provider returned null or empty assertion",
-                            AuthenticationErrorCode.INVALID_JWT);
+                    throw new MsalClientException(
+                        "Assertion provider returned null or empty assertion",
+                        AuthenticationErrorCode.INVALID_JWT);
                 }
+
                 return generatedAssertion;
+            } catch (MsalClientException ex) {
+                throw ex;
             } catch (Exception ex) {
                 throw new MsalClientException(ex);
             }
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
@@ -4,21 +4,66 @@
 package com.microsoft.aad.msal4j;
 
 import java.util.Objects;
+import java.util.concurrent.Callable;
 
 final class ClientAssertion implements IClientAssertion {
 
     static final String ASSERTION_TYPE_JWT_BEARER = "urn:ietf:params:oauth:client-assertion-type:jwt-bearer";
     private final String assertion;
+    private final Callable<String> assertionProvider;
 
+    /**
+     * Constructor that accepts a static assertion string
+     *
+     * @param assertion The JWT assertion string to use
+     * @throws NullPointerException if assertion is null or empty
+     */
     ClientAssertion(final String assertion) {
         if (StringHelper.isBlank(assertion)) {
             throw new NullPointerException("assertion");
         }
 
         this.assertion = assertion;
+        this.assertionProvider = null;
     }
 
+    /**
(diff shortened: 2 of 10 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -93,23 +98,58 @@ private void initClientAuthentication(IClientCredential clientCredential) {
...
+        return this.getAssertionString(this.clientCredential);
+    }
+
+    /**
+     * Checks if the JWT-formatted assertion has expired by parsing the "exp" claim.
+     *
+     * @param jwt JWT string
+     * @return true if the JWT has expired. Otherwise false
+     */
+    boolean hasJwtExpired(String jwt) {
```
sudoapt-getclean: A Wise man once advised me to never write optimizations before:
1.  writing tests that proved that there was something that needed to be optimized. 
2.  writing tests that actually proved the optimization where fast enough to justify the extra code.

anyway, that's my 2 pennies.

sudoapt-getclean (approved):

bgavrilMS (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -81,6 +84,8 @@ private ConfidentialClientApplication(Builder builder) {
     private void initClientAuthentication(IClientCredential clientCredential) {
         validateNotNull("clientCredential", clientCredential);
 
+        this.clientCredential = clientCredential;
```
bgavrilMS: This should not be cached

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -93,23 +98,58 @@ private void initClientAuthentication(IClientCredential clientCredential) {
...
-                    clientId(),
-                    clientCertificate,
-                    this.authenticationAuthority.selfSignedJwtAudience(),
-                    sendX5c,
-                    useSha1).assertion();
+            // Check if the current assertion is null or has expired, and if so create a new one
+            if (this.assertion == null || hasJwtExpired(this.assertion)) {
+                boolean useSha1 = Authority.detectAuthorityType(this.authenticationAuthority.canonicalAuthorityUrl()) == AuthorityType.ADFS;
+
+                this.assertion = JwtHelper.buildJwt(
```
bgavrilMS: 1. The assertion should not be cached. 
2. The assertion contains the issuer, which is the Entra token endpoint. The token endpoint contains the tenant. So you can't resolve the assertion at the Application level, only at the request level, where the user can override the tenant?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClientCertificateTest.java
```diff
@@ -77,9 +77,9 @@ void testIClientCertificateInterface_CredentialFactoryUsesSha256() throws Except
             HttpRequest request = parameters.getArgument(0);
             String requestBody = request.body();
 
-            SignedJWT signedJWT = SignedJWT.parse(cca.assertion);
```
bgavrilMS, line 80: I think we are missing a test

bgavrilMS (dismissed):
assertion should be at request level?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: We discussed this elsewhere but just to reiterate here: I was originally trying to restore the the library's old behavior, where the assertion was cached but refreshed if it expired.

However, as we discussed and as you mention here that old behavior was flawed, so I'll either update this PR or restore @sudoapt-getclean's https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/985 to make it per-request only.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: In the latest commit `ConfidentialClientApplication` now just stores the IClientCredential instance it was originally created with, rather than any actual credential info.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: In the latest commit ConfidentialClientApplication now just stores the IClientCredential instance it was originally created with, rather than any actual credential info.

The assertion for a ClientCredential is now created on a per-request basis, rather than being stored/refreshed like it was in older versions.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ClientCertificateTest.java
Avery-Dunn: Added a test to ensure new assertions are created on each token call, is there another scenario that is missing?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -5,6 +5,8 @@
 
 import org.slf4j.LoggerFactory;
 
+import java.util.Base64;
```
neha-bhargava: Are these imports needed?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -86,25 +86,76 @@ private void addQueryParameters(OAuthHttpRequest oauthHttpRequest) {
...
+    /**
+     * Determines which credential to use for authentication:
+     * - If the request is a ClientCredentialRequest with a specified credential, use that
+     * - Otherwise use the application's credential
+     *
+     * @param application The confidential client application
+     * @return The credential to use, may be null if no credential is available
+     */
+    private IClientCredential getCredentialToUse(ConfidentialClientApplication application) {
+        if (msalRequest instanceof ClientCredentialRequest &&
```
neha-bhargava: Is this flow covered by tests? If not can we add tests to cover this?

Avery-Dunn: There are some end-to-end tests in ClientCredentialsIT that would go through this code path but it wasn't covered that well.

In the latest commit I added a new `testCredentialPrecedenceAndMixing()` unit test to `ClientCredentialTest` to more explicitly show the app-level credential getting overridden at the request level.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: Just removed them, they were just leftovers from an earlier commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
Avery-Dunn: I also wrote this in another comment, but: after some more discussions we agreed that the library's original way of doing it was incorrect, and as of the latest commits we're now making the assertions per-request.

neha-bhargava (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -131,6 +131,40 @@ static void validateAuthority(URL authorityUrl) {
...
+     *
+     * @param originalAuthority The original authority to base the new one on
+     * @param newTenant The new tenant to use in the authority URL
+     * @return A new Authority instance with the specified tenant
+     * @throws MalformedURLException If the new authority URL is invalid
+     * @throws NullPointerException If originalAuthority or newTenant is null
+     */
+    static Authority replaceTenant(Authority originalAuthority, String newTenant) throws MalformedURLException {
+        if (originalAuthority == null) {
+            throw new NullPointerException("originalAuthority");
```
gladjohn: should these be `IllegalArgumentException`?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -86,25 +86,75 @@ private void addQueryParameters(OAuthHttpRequest oauthHttpRequest) {
...
+        if (msalRequest instanceof ClientCredentialRequest) {
+            ClientCredentialParameters parameters = ((ClientCredentialRequest) msalRequest).parameters;
+
+            if (parameters.clientCredential() != null) {
+                credentialToUse = parameters.clientCredential();
+            }
+
+            if (parameters.tenant() != null) {
+                try {
+                    authorityToUse = Authority.replaceTenant(authorityToUse, parameters.tenant());
```
gladjohn, line 123: is there a test for B2C?

gladjohn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
Avery-Dunn: On the public API side we usually throw `IllegalArgumentException` to inform the customer that we expected they might pass it but it's not actually an accepted value, but internally there are a lot of `NullPointerException` for simplicity and to show something unexpected happened.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
```diff
@@ -4,21 +4,66 @@
...
+        if (assertionProvider != null) {
+            try {
+                String generatedAssertion = assertionProvider.call();
+                if (StringHelper.isBlank(generatedAssertion)) {
+                    throw new MsalClientException("Assertion provider returned null or empty assertion",
+                            AuthenticationErrorCode.INVALID_JWT);
+                }
+                return generatedAssertion;
+            } catch (Exception ex) {
+                throw new MsalClientException(ex);
```
trwalke, line 68: So you are throwing a new exception on line 58 already.  Will creating a new exception here change the stack trace? Will this make the previously thrown exception an inner exception?

trwalke (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
Avery-Dunn: Good catch, it would've been an inner exception and made the error less obvious. In the latest commit I made it re-throw an MsalClientException's it catches.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
Avery-Dunn: I had thought there was a test for B2C but I was confusing it with an OBO test. After adding a new test it became clear that the `replaceTenant()` method wouldn't work for B2C authorities.

In the latest commit I refactored that test to handle all authority formats in a simpler way, fixed some other issues with B2C and overriding tenants, and added a new test to confirm the behavior.

gladjohn: great, thanks for adding the tests to improve code coverage.

gladjohn (approved):

bgavrilMS (approved):

PR #988, opened 2025-09-04
Improve behavior related to assertions

Avery-Dunn:
Fixes behavior related to assertions in two areas:
- https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/986 changed the way assertions for `ClientCertificates` were made, now that they are per-request they potentially need the tenant set at the request-level and not just the app-level
- https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/879 and https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/977 both describe an issue when creating an assertion with a callback, the original design only called the callback when the `ClientAssertion` instance was first created

This PR fixes both issues and generally improves the parts of the codebase where these assertions are used:
- Adjusts `ClientAssertion` and `ClientCredentialFactory` to properly call the callback whenever the assertion is needed
- Adjusts `TokenRequestExecutor` to optionally use the tenant from the request to create assertions, and refactors several methods to be simpler and more manageable
- Adds a new method to `Authority` to help with overriding the tenant in an authority
- New tests in `ClientCertificateTest` and `ClientCertificateTest` to cover the new behavior and fill some gaps

This branch was based off the one in https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/986, so it must be handled first.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
@@ -131,6 +131,40 @@ abstract class Authority {
         }
     }
 
+    /**
+     * Creates a new Authority instance with a different tenant.
+     * This is useful when overriding the tenant for a specific request.
+     *
+     * @param originalAuthority The original authority to base the new one on
+     * @param newTenant The new tenant to use in the authority URL
+     * @return A new Authority instance with the specified tenant
+     * @throws MalformedURLException If the new authority URL is invalid
+     * @throws NullPointerException If originalAuthority or newTenant is null
+     */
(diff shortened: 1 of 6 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/Authority.java
```diff
@@ -131,6 +131,40 @@ static void validateAuthority(URL authorityUrl) {
         }
     }
 
+    /**
+     * Creates a new Authority instance with a different tenant.
+     * This is useful when overriding the tenant for a specific request.
```
neha-bhargava: Nit:
```suggestion
     * This is useful when overriding the tenant at request level.
```

Issue #958, opened 2025-05-23
[Bug] MSAL should use the recommended padding in client_credentials


bgavrilMS:
### Library version used

latest

### Java version

8

### Scenario

ConfidentialClient - service to service (AcquireTokenForClient)

### Is this a new or an existing app?

This is a new app or experiment

### Issue description and reproduction steps

The recommended padding is PSS, as per docs https://learn.microsoft.com/en-us/entra/identity-platform/certificate-credentials#header

MSAL Java use RSS - https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/b02c8f2fcd532b762f4d21f45d7e3f6984909b2d/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java#L49

### Relevant code snippets

```csharp

```

### Identity provider

Microsoft Entra ID (Work and School accounts and Personal Microsoft accounts)

PR #990, opened 2025-09-10
Use recommended algorithm in assertions

Avery-Dunn:
Fixes the issue described in https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/958

Microsoft recommends PS256 for assertions and other MSALs follow this, however MSAL Java still used RS256. It seems like PSS support was only added in Java 11, but it was thankfully backported to Java 8.

This PR refactors JwtHelper to use the recommended algorithm and adjusts/expands tests in order to cover the new behavior.

In addition, while making the changes I discovered a weird edge case:
- To create a Signature instance to sign the assertion, we need to provide a `PrivateKey`
- When retrieving a `PrivateKey` from the Windows-MY keystore the underlying type is "sun.security.mscapi.CPrivateKey"
- The `sun.security.mscapi.CPrivateKey` type was accepted by the Signature instance created with "SHA256withRSA", but for some reason the Signature created with "RSASSA-PSS" throws an error about it not specifically being an `RSAPrivateKey` instance

Since our integration tests use a cert stored in Windows-MY simply changing the assertion to use PS256 would've broken many of our tests. I'm sure plenty of our customers retrieve certs in the same way, and I'm not sure what other key types would cause this error.

So to avoid any breaking changes, this PR also implements a mechanism to fallback to the old RS256 style when we get a certain exception, and has extra tests covering this fallback behavior.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
@@ -4,7 +4,10 @@
 package com.microsoft.aad.msal4j;
 
 import java.nio.charset.StandardCharsets;
+import java.security.InvalidKeyException;
 import java.security.Signature;
+import java.security.spec.MGF1ParameterSpec;
+import java.security.spec.PSSParameterSpec;
 import java.util.ArrayList;
 import java.util.Base64;
 import java.util.HashMap;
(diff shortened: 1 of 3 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
```diff
@@ -22,62 +25,183 @@ static ClientAssertion buildJwt(String clientId, final ClientCertificate credent
...
+     *
+     * @param clientId     The client ID to use as the issuer and subject
+     * @param credential   The certificate credential used for signing
+     * @param jwtAudience  The audience claim for the JWT
+     * @param sendX5c      Whether to include the x5c header with certificate chain
+     * @param useSha1      Whether to use SHA-1 hash for thumbprint instead of SHA-256
+     * @return A ClientAssertion containing the signed JWT
+     * @throws Exception If JWT creation or signing fails
+     */
+    private static ClientAssertion generateRs256Jwt(String clientId, ClientCertificate credential,
```
bgavrilMS, line 92: There seem to be a lot of code dup between thjis method and the ps256 version

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
```diff
@@ -22,62 +25,183 @@ static ClientAssertion buildJwt(String clientId, final ClientCertificate credent
...
-            // is not supported or the IClientCredential.publicCertificateHash256() method is not implemented, the library will default to SHA-1.
-            String hash256 = credential.publicCertificateHash256();
-            if (useSha1 || hash256 == null) {
-                header.put("x5t", credential.publicCertificateHash());
-            } else {
-                header.put("x5t#S256", hash256);
+            // First try with PS256 (preferred)
+            return generatePS256Jwt(clientId, credential, jwtAudience, sendX5c, useSha1);
+        } catch (InvalidKeyException e) {
+            // If the key isn't compatible with PSS, fall back to RS256.
```
bgavrilMS, line 31: This is very strange, because the padding algorithm can be always be used with a PSS key

image

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/HelperAndUtilityTests.java
```diff
@@ -187,4 +190,99 @@ void JsonHelper_createIdTokenFromEncodedTokenString_InvalidJsonInToken() {
 
         assertEquals(AuthenticationErrorCode.INVALID_JSON, exception.errorCode());
     }
+
+    @Test
+    void JwtHelper_buildJwt_UsesPSS256WhenSupported() throws Exception {
```
bgavrilMS, line 195: The way it's implemented in MSAL .NET is: 

- for AAD, use SHA2 and PSS
- for ADFS, use SHA1 and PKCS1


I see that the way to specify a certificate in MSAL Java is similar. So why not have some higher level tests, i.e. start from the public API and assert what gets put on the wire?

bgavrilMS (changes requested):
RSA padding should be universal, but I understand that Java8 uses CAPI instead of CNG.

However, for backwards compat, in places where we use SHA1, we should continue to use PKCS1. Does this happen? Can we have better E2E tests?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/JwtHelper.java
bgavrilMS: Read more about this, CNG isn't available on Java 8.

Avery-Dunn:
Closing, will need to investigate cryptography options more.

PR #993, opened 2025-09-12
Use more common "static final" instead of "final static"

Avery-Dunn:
Certain Java conventions suggest "static final" as the ordering of keywords when making a constant, rather than "final static". 

SonarQube considered this a minor issue, and as this PR shows it was quite common in the codebase.

Although this PR affects a lot of files it's the exact same change in each one, and should not affect the behavior of the library in any way.

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenInteractiveIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenInteractiveIT.java
@@ -27,7 +27,7 @@ import java.util.concurrent.ExecutionException;
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class AcquireTokenInteractiveIT extends SeleniumTest {
-    private static final Logger LOG = LoggerFactory.getLogger(AuthorizationCodeIT.class);
+    private static final Logger LOG = LoggerFactory.getLogger(AcquireTokenInteractiveIT.class);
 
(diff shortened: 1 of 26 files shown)


msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenInteractiveIT.java
```diff
@@ -27,7 +27,7 @@
 
 @TestInstance(TestInstance.Lifecycle.PER_CLASS)
 class AcquireTokenInteractiveIT extends SeleniumTest {
-    private final static Logger LOG = LoggerFactory.getLogger(AuthorizationCodeIT.class);
+    private static final Logger LOG = LoggerFactory.getLogger(AuthorizationCodeIT.class);
```
gladjohn:
```suggestion
    private static final Logger LOG = LoggerFactory.getLogger(AcquireTokenInteractiveIT.class);
```

should it be this?

gladjohn (approved):

msal4j-sdk/src/integrationtest/java/com.microsoft.aad.msal4j/AcquireTokenInteractiveIT.java
Avery-Dunn: Yep, must've just missed that on a copy/past a while back.

bgavrilMS (approved):

neha-bhargava (approved):

PR #998, opened 2025-10-09
Deprecate ROPC flow

Ugonnaak1:
[AB#3397688](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3397688)

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IPublicClientApplication.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IPublicClientApplication.java
@@ -14,11 +14,15 @@ import java.util.concurrent.CompletableFuture;
 public interface IPublicClientApplication extends IClientApplicationBase {
 
     /**
-     * Acquires tokens from the authority configured in the application via Username/Password authentication.
+     * Acquires tokens from the authority configured in the application via Username/Password authentication.<br>
+     * <p><b>Deprecated:</b> This API has been deprecated and will be removed in a future release. Use a more secure flow instead.<br>
+     * See <a href="https://aka.ms/msalnet-ropc-migration">https://aka.ms/msalnet-ropc-migration</a> for migration guidance.
      *
      * @param parameters instance of {@link UserNamePasswordParameters}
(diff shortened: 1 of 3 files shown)


rayluo (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IPublicClientApplication.java
```diff
@@ -14,11 +14,15 @@
 public interface IPublicClientApplication extends IClientApplicationBase {
 
     /**
-     * Acquires tokens from the authority configured in the application via Username/Password authentication.
+     * Acquires tokens from the authority configured in the application via Username/Password authentication.<br>
+     * <p><b>Deprecated:</b> This API has been deprecated and will be removed in a future release. Use a more secure flow instead.<br>
+     * See <a href="https://aka.ms/msalnet-ropc-migration">https://aka.ms/msalnet-ropc-migration</a> for migration guidance.
```
Avery-Dunn: It would make sense to put this comment/annotation over the `UserNamePasswordParameters` class also.

This is the only public API that uses it so we may as well get rid of that whole class in the future.

Ugonnaak1: That makes sense. I've added the deprecation comment here

Ugonnaak1: Also, is there a reason the CI pipeline tests aren't running?

Avery-Dunn (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IPublicClientApplication.java
Avery-Dunn: A few years back we realized it could be a risk to let every contributor inject code into our pipelines, so most of our repos don't trigger the pipelines when the PR is based on a fork.

Normally I'd run them manually before merging, but since these are just comments there shouldn't be any problem.

PR #1001, opened 2025-10-29
Deprecate extraQueryParameters

Avery-Dunn:
Deprecates the extraQueryParameters API according to KR [3310905](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3310905), similar to what was done in MSAL.NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5536

This API was meant for niche scenarios which the library did not explicitly cover, and not really meant for production. It can affect the contents and validity of tokens but were not used in our caching scheme which led to bad caching behavior: if the extra query parameters changed between requests we could return tokens that were not valid for the latest set of parameters.

Therefore we are deprecating the current API and, if there is a need for it, may add a new API in the future to perform similar behavior in a more robust way.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
@@ -234,6 +234,10 @@ public class AuthorizationRequestUrlParameters {
         return this.instanceAware;
     }
 
+    /**
+     * @deprecated Not recommended for production scenarios. It will be removed in a future release, and the behavior may be replaced by a new API.
+     */
+    @Deprecated
     public Map<String, String> extraQueryParameters() {
         return this.extraQueryParameters;
     }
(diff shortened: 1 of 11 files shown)


bgavrilMS (approved):

4gust (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -234,6 +234,10 @@ public boolean instanceAware() {
         return this.instanceAware;
     }
 
+    /**
+     * @deprecated Not recommended for production scenarios. It will be removed in a future release, and the behavior may be replaced by a new API.
+     */
+    @Deprecated
```
neha-bhargava, line 240: Should this message be accompanied by the new API? I feel the message is a bit confusing if I am already using this API in production.

gladjohn: agree, should we mimic MSAL .NET PR here as well?

Avery-Dunn: For all the other MSALs we decided to just deprecate the existing API, and not make all the other cache key changes like .NET

Long story short, for .NET we knew there were many customers using a lot of first-party ESTS features for testing (like getting already-expired tokens, simulating an outage, etc.). But for the other languages, the plan is more-or-less to deprecate the existing API and add the new behavior only if a customer specifically asks for it one day.

PR #1004, opened 2026-01-10
Deprecate ResponseMode.QUERY in system browser auth flow, automatically override to FORM_POST with warning

ashok672:
The system browser authentication flow no longer supports query response mode. This change deprecates ResponseMode.QUERY . When developers pass ResponseMode.QUERY, it will be automatically overridden to FORM_POST with a warning log, ensuring backward compatibility while guiding users to the supported mode.


msal4j-persistence-extension/target/test-classes/log4j.properties
```diff
@@ -0,0 +1,5 @@
+log4j.rootLogger=TRACE, stdout
+log4j.appender.stdout=org.apache.log4j.ConsoleAppender
+log4j.appender.stdout.Target=System.out
+log4j.appender.stdout.layout=org.apache.log4j.PatternLayout
+log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd'T'HH:mm:ss.SSS} %-5p [%c] - %m%n
```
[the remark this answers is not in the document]
Avery-Dunn: I agree with Copilot here, looks like this was accidentally committed?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
```diff
@@ -98,32 +98,23 @@ void testBuilder_conflictingParameters() {
     }
 
     @Test
-    void testBuilder_optionalParameters() throws UnsupportedEncodingException {
```
Avery-Dunn, line 101: Why rename the test and remove most of what it was checking?

In the existing test there already was an assert for `assertEquals(queryParameters.get("response_mode"), "query");`, you should be able to just have that look for "form_post" or Response.Mode.FORM_POST instead #Resolved

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -368,7 +377,9 @@ public Builder nonce(String val) {
 
         /**
          * Specifies the method that should be used to send the authentication result to your app.
+         * @deprecated ResponseMode.QUERY is deprecated. If you pass ResponseMode.QUERY, it will be automatically overridden to ResponseMode.FORM_POST.
          */
+        @Deprecated
```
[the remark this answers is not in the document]
Avery-Dunn: I agree with Copilot here, the comment is helpful but having `@Deprecated` here implies the method itself will be fully removed in the future.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
```diff
@@ -113,9 +113,18 @@ private AuthorizationRequestUrlParameters(Builder builder) {
         }
 
         if (builder.responseMode != null) {
-            this.responseMode = builder.responseMode;
-            requestParameters.put("response_mode",
-                    builder.responseMode.toString());
+            // Override QUERY with FORM_POST as QUERY is deprecated
```
Avery-Dunn: Rather than add the logic to check for `Query` here, it might be simpler to add an if statement to the actual builder method:
```
public Builder responseMode(ResponseMode val) {
    // Override QUERY with FORM_POST as QUERY is deprecated
    if (val == ResponseMode.QUERY) {
        LOG.warn("ResponseMode.QUERY is deprecated and will be removed in a future release. " +
                "Automatically overriding to ResponseMode.FORM_POST.");
        this.responseMode = ResponseMode.FORM_POST;
    } else {
        this.responseMode = val;
    }
    return self();
}
```

That way, this constructor stays the same and only valid values get set in the builder.

Avery-Dunn (changes requested):
See above comments: looks like there are some unnecessary additions and changes to the tests.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
ashok672: I didn't realize this was still present when I really wanted to delete this.

msal4j-persistence-extension/target/test-classes/log4j.properties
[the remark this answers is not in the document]
ashok672: yes, it was accidentally committed.

msal4j-persistence-extension/target/test-classes/log4j.properties
```diff
@@ -0,0 +1,10 @@
+# This file is located under the Maven/Gradle build output directory:
```
Avery-Dunn: This file can be removed entirely, it's in the "target" directory which means it's just a file that gets made when building the project and seems to just have some comments from Copilot? #Resolved

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
bgavrilMS: Agreed, we should not change the AuthCodeUri for all scenarios. Or change it, but deprecate the "GetAuthorizationRequestUri" existing API and add a "GetAuthorizationRequestUriWithPost" and fully deprecate ResponseMode param

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParametersTest.java
ashok672: I wanted the test name to be explicit what we are testing but I think I can go with what you suggested as well. I have no strong preference.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthorizationRequestUrlParameters.java
ashok672: I do like Bogdan's suggestion but to keep it simple for now, I will go ahead with minimal change suggested by Avery.

Avery-Dunn (approved):

fransf-anva:
@ashok672 @Avery-Dunn  Do you have a reference for the "no longer supports query response mode" and "due to lack of support" assertions? 

Microsoft's own EntraID documentation (https://learn.microsoft.com/en-us/entra/identity-platform/v2-oauth2-auth-code-flow) don't make any mention of it. The "System browser authentication flow" looks like a normal OAuth2 authorization grant flow, and Microsoft still seems to support `response_mode=query` on that just fine. 

Would really appreciate some clarification here, since this is quite a big, breaking change. Thanks!

PR #1009, opened 2026-02-03
Add copilot-instructions.md and fix tests

Avery-Dunn:
Adds a copilot-instructions.md file. It mainly covers library architecture and supported auth scenarios to help keep future prompts/agents identify and focus on the relevant parts of the library.

This PR also fixes some tests which started failing without any underlying changes to the codebase:
- Fixes a client credential test which was supposed to cover secrets but was using a cert
- Adjusts most client credential tests to use the tenanted authority from the config, since they began to fail with AADSTS7000229 errors
- Disables the two browser-based CIAM tests until it can be investigated more, it seems like the login page's HTML elements may have changed: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/1010

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,239 @@
+# MSAL Java - Copilot Instructions
+
+## Quick Reference
+
+- **Main Module**: `msal4j-sdk/` (focus here for most work)
+- **Main Package**: `com.microsoft.aad.msal4j`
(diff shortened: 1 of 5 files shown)


gladjohn (approved):

gladjohn:
please file a tracking item for this `Disables a browser-based CIAM test until it can be investigated more, it seems like the login page's HTML elements may have changed`

RyAuld (approved):

RyAuld (approved):

PR #1025, opened 2026-04-15
Add Federated Managed Identity (FMI) support for client credentials flow

Avery-Dunn:
This PR adds Federated Managed Identity (FMI) support to MSAL Java's client credentials flow, enabling Leg 1 and Leg 2 of the agent identity protocol. FMI allows a blueprint application to acquire tokens scoped to a specific agent identity by providing an `fmi_path` parameter in the token request.

#### What's included

**New public APIs:**

- `ClientCredentialParameters.fmiPath(String)` — sets the `fmi_path` body parameter sent in the client credentials POST request. Rejects null, empty, and whitespace values (matching MSAL .NET's `WithFmiPath` validation).
- `ClientCredentialFactory.createFromCallback(Function<AssertionRequestOptions, String>)` — creates a client assertion from a context-aware callback function (needed for Leg 2, where the assertion callback must know the FMI path to acquire the correct credential)
- `AssertionRequestOptions` — immutable context object passed to assertion callbacks, providing `clientId()`, `tokenEndpoint()`, and `clientAssertionFmiPath()`

**Cache key isolation (`ext_cache_key`):**

- Tokens acquired with different `fmi_path` values are isolated in the cache using an extended cache key hash
- Hash algorithm matches other MSALs: SHA-256 of sorted key+value concatenation, Base64URL-encoded
- Hash is memoized on the `ClientCredentialParameters` instance (immutable after construction), matching MSAL .NET's approach of computing once at cache item creation
- Credential type changes from `"AccessToken"` to `"AText"` (AccessToken_Extended) for tokens with additional cache key components
- Cache lookups correctly filter bidirectionally: FMI-tagged entries are not returned for non-FMI requests, and vice versa

**Internal changes:**

- `ClientCredentialRequest` injects `fmi_path` into the OAuth2 POST body when set
- `TokenRequestExecutor` passes `clientAssertionFmiPath` to the assertion callback via `AssertionRequestOptions`
- `ClientAssertion` extended to support `Function<AssertionRequestOptions, String>` callbacks (in addition to existing static `String` and `Callable<String>` modes). Dispatch priority: context-aware Function > Callable > static string.
- `TokenCache` extended with `extCacheKeyHashMatches()` bidirectional filter logic
- `AccessTokenCacheEntity` extended with `extCacheKeyHash` field
- `StringHelper.computeExtCacheKeyHash()` implements the SHA-256 hash algorithm
- `AcquireTokenByClientCredentialSupplier` propagates `extCacheKeyHash` through the silent request path

**Tests:**

- `FmiTest.java` — 19 unit tests covering: input validation (null/empty/blank fmiPath rejection), cache key computation and cross-SDK hash verification, assertion callback context propagation (`clientAssertionFmiPath` passed correctly), bidirectional cache isolation (FMI vs non-FMI, different FMI paths), and legacy `Callable` fallback behavior
- `FmiIT.java` — 7 integration tests covering FMI credential acquisition (from cert, from RMA, chained FMI credentials, token from FMI credential, multi-FMI-path cache isolation)
- `AgenticIT.java` — 3 integration tests covering FMI scenarios using RMA (FMI credential acquisition from RMA, token cache hit for repeated requests)

#### Alignment with MSAL .NET

This implementation matches the FMI behavior available on MSAL .NET's `main` branch. Key equivalences:

| .NET | Java |
|------|------|
| `WithFmiPath(string)` | `ClientCredentialParameters.fmiPath(String)` |
| `WithClientAssertion(Func<AssertionRequestOptions, Task<string>>)` | `ClientCredentialFactory.createFromCallback(Function<AssertionRequestOptions, String>)` |
| `AssertionRequestOptions` (ClientId, TokenEndpoint, FmiPath) | `AssertionRequestOptions` (clientId, tokenEndpoint, clientAssertionFmiPath) |
| `CoreHelpers.ComputeAccessTokenExtCacheKey()` | `StringHelper.computeExtCacheKeyHash()` |
| `"ATExt"` credential type | `"AText"` credential type (case differs but cache key is lowercased for lookups in both SDKs) |
| Hash computed once at cache item creation (`InitCacheKey`) | Hash memoized on `ClientCredentialParameters` |
| `string.IsNullOrWhiteSpace` validation on `WithFmiPath` | `ParameterValidationUtils.validateNotBlank` on `fmiPath()` builder setter |

#### Known differences from .NET (intentional)

- Java's `AssertionRequestOptions` exposes `clientId`, `tokenEndpoint`, `clientAssertionFmiPath`. .NET additionally exposes `TenantId`, `Authority`, `Claims`, `ClientCapabilities`, `CorrelationId`. The additional fields can be added in a future PR if needed.
- Java does not yet have `WithFmiPathForClientAssertion` (a separate FMI path passed only to the assertion callback, distinct from the body parameter). This is infrastructure for the composite `AcquireTokenForAgent` API which will be added in a subsequent PR.
- Java does not have `CancellationToken` semantics. The callback is `Function<AssertionRequestOptions, String>` (synchronous) rather than .NET's async `Func<..., CancellationToken, Task<string>>`.
- Java does not yet have `AssertionResponse` / `ClientSignedAssertion` support (assertion + token-binding certificate for mTLS PoP). This can be added in a future PR when jwt-pop scenarios are needed.


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
```diff
@@ -40,20 +43,61 @@ final class ClientAssertion implements IClientAssertion {
...
+
+    /**
+     * Constructor that accepts a context-aware function that provides the assertion string.
+     * The function receives {@link AssertionRequestOptions} containing context such as
+     * the client ID, token endpoint, and FMI path.
+     *
+     * @param contextAwareAssertionProvider A function that receives context and returns a JWT assertion string
+     * @throws NullPointerException if contextAwareAssertionProvider is null
+     */
+    ClientAssertion(final Function<AssertionRequestOptions, String> contextAwareAssertionProvider) {
```
bgavrilMS, line 57: If we go with this callback with input object, we should also future proof things and have an output object. It will future proof mtls pop scenarios. 

`Function<AssertionRequestOptions, AssertionResponse>`

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
```diff
@@ -0,0 +1,438 @@
...
+        DefaultHttpClient httpClientMock = mock(DefaultHttpClient.class);
+
+        when(httpClientMock.send(any(HttpRequest.class))).thenReturn(
+                TestHelper.expectedResponse(HttpStatus.HTTP_OK,
+                        TestHelper.getSuccessfulTokenResponse(new HashMap<>())));
+
+        ConfidentialClientApplication cca =
+                ConfidentialClientApplication.builder("clientId", ClientCredentialFactory.createFromSecret("secret"))
+                        .authority("https://login.microsoftonline.com/tenant/")
+                        .instanceDiscovery(false)
```
bgavrilMS: Why does this need to be turned off? Same for validate authority.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
```diff
@@ -0,0 +1,438 @@
...
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class FmiTest {
+
+    // ========================================================================
+    // §2: fmi_path body parameter
+    // ========================================================================
+
+    @Test
+    void fmiPath_IncludedInTokenRequestBody() throws Exception {
```
bgavrilMS, line 32: We can have intergation tests.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
```diff
@@ -0,0 +1,438 @@
...
+                        .authority("https://login.microsoftonline.com/tenant/")
+                        .instanceDiscovery(false)
+                        .validateAuthority(false)
+                        .httpClient(httpClientMock)
+                        .build();
+
+        ClientCredentialParameters parameters = ClientCredentialParameters
+                .builder(Collections.singleton("api://AzureADTokenExchange/.default"))
+                .fmiPath("agentAppId123")
+                .skipCache(true)
```
bgavrilMS: why skip cache?

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
```diff
@@ -0,0 +1,438 @@
...
+        ClientCredentialParameters params = ClientCredentialParameters
+                .builder(Collections.singleton("api://AzureADTokenExchange/.default"))
+                .fmiPath("agentA")
+                .build();
+        cca.acquireToken(params).get();
+
+        // Assert — cache entry should use "atext" credential type (matching Go/Python convention)
+        assertEquals(1, cca.tokenCache.accessTokens.size());
+        String cacheKey = cca.tokenCache.accessTokens.keySet().iterator().next();
+        assertTrue(cacheKey.contains("-atext-"),
```
bgavrilMS: Pls assert the full cache key against a constant.

4gust (approved):

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
Avery-Dunn: As of the latest commits there is now an AgenticIT and FmiIT file with integration tests that should more-or-less match the coverage of .NET

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
Avery-Dunn: Skipping instance discovery and authority validation is just a convenience used in a lot of Java's unit tests: unless the test is meant to cover authority behavior it's easier to turn it all off than deal with an extra HTTP mock or worrying about correct authority formats.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
Avery-Dunn: Not sure why the agent added it, maybe to ensure the tests are clean? Regardless it wasn't needed, and I just removed it in the latest commit.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
Avery-Dunn: In the latest commit the test now check the full cache key

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/AgenticIT.java
```diff
@@ -0,0 +1,183 @@
...
+ * <p>These tests use MSAL token acquisition APIs (unlike AgenticRawHttpIT which uses raw HTTP).
+ *
+ * <p>Test configuration:
+ * <ul>
+ *   <li>RMA app: {@link #RMA_CLIENT_ID}</li>
+ *   <li>Agent app: {@link #AGENT_APP_ID}</li>
+ *   <li>Tenant: {@link #TENANT_ID}</li>
+ * </ul>
+ *
+ * <p>Flows tested (FMI-only, no FIC/user_fic on this branch):
```
neha-bhargava: Probably AI generated this comment because of the 2 PRs but its a bit weird to have it in the code. "on this branch"

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
neha-bhargava: In .Net we add the mocks, I think we should add the mocks so that it mimics the common scenario

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
```diff
@@ -40,6 +40,13 @@ AuthenticationResult execute() throws Exception {
                         context,
                         null);
 
+                // Propagate ext_cache_key_hash for fmi_path-based cache isolation
+                if (!StringHelper.isBlank(this.clientCredentialRequest.parameters.fmiPath())) {
+                    java.util.TreeMap<String, String> components = new java.util.TreeMap<>();
+                    components.put("fmi_path", this.clientCredentialRequest.parameters.fmiPath());
```
Robbie-Microsoft: `extCacheKeyHash` is computed here and again in `TokenCache.computeExtCacheKeyHashForRequest` — same `TreeMap` + `computeExtCacheKeyHash` logic in two places. If a second component is added to one location and not the other, cache reads and writes will silently diverge. The fully-qualified `java.util.TreeMap` here (instead of an import) suggests this was added in a hurry. Recommend computing the hash once — either in `ClientCredentialParameters` or `ClientCredentialRequest` — and passing it through.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionResponse.java
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Gets the optional token-binding certificate for mutual-TLS Proof-of-Possession (mTLS PoP).
+     * When present, MSAL uses {@code client_assertion_type=jwt-pop} instead of {@code jwt-bearer}.
+     *
+     * @return the binding certificate, or null if not using mTLS PoP
+     */
+    public X509Certificate tokenBindingCertificate() {
+        return tokenBindingCertificate;
```
Robbie-Microsoft: `tokenBindingCertificate()` is read in exactly one place (`TokenRequestExecutor.addAssertionResponseParams`) — to decide between `jwt-pop` and `jwt-bearer`. The certificate is never passed to `MtlsSslContextHelper` or any TLS layer. Is this intentional? If the cert will be wired to actual mutual-TLS in a follow-up PR, that should be documented here. If the only purpose is to signal which assertion type to use, the API is misleading — callers will expect MSAL to do something with the certificate, but it doesn't. A boolean or enum would be more honest about what MSAL actually uses it for.

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/AgenticIT.java
Avery-Dunn: Yeah, agents seem to love putting metacommentary like this into test comments, in an older commit I cleaned up a bunch mentioning the .NET tests it was using as references but this and some other comments slipped by.

In the latest commit the comments should now be more focused.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
Avery-Dunn: In the latest commit I refactored the cache key creation to be more centralized and remove the duplicates.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionResponse.java
Avery-Dunn: Basically, this class and related APIs were added due to this comment about future proofing for mTLS: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1025#discussion_r3174025684

However, after double-checking I've decided to remove it: `AssertionResponse` and everything referencing it was purely additive in this FMI-focused branch and not referenced at all in the FMI and Agentic branches, so when MSAL Java eventually supports mTLS it should be a smooth addition with no risk of breaking APIs

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
Avery-Dunn: I originally added it after your comment, but after thinking about it more and this comment I decided to remove it: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1025#discussion_r3204772801

Unless you're saying we shouldn't have a string-based option at all, the `AssertionResponse` option is purely additive and we shouldn't have any issues implementing it alongside the rest of mTLS support.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionRequestOptions.java
```diff
@@ -0,0 +1,53 @@
...
+package com.microsoft.aad.msal4j;
+
+/**
+ * Context information passed to assertion provider callbacks when MSAL needs a fresh client assertion.
+ * This allows assertion providers to generate context-aware assertions, such as including the FMI path
+ * in Federated Managed Identity (agent identity) scenarios.
+ *
+ * <p>For more details on agent identity scenarios, see the MSAL documentation on FMI/FIC flows.</p>
+ */
+public class AssertionRequestOptions {
```
neha-bhargava: This should be final class

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
```diff
@@ -25,6 +27,7 @@ final class ClientAssertion implements IClientAssertion {
 
         this.assertion = assertion;
         this.assertionProvider = null;
+        this.contextAwareAssertionProvider = null;
```
neha-bhargava, line 30: Since we only expect either one of the assertion/ either providers to exist. Can we have a better pattern for this class. Like strategy?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CredentialTypeEnum.java
```diff
@@ -6,6 +6,7 @@
 enum CredentialTypeEnum {
 
     ACCESS_TOKEN("AccessToken"),
+    ACCESS_TOKEN_EXTENDED("atext"),
```
neha-bhargava: Where do you get the "atext" value from?

neha-bhargava: In .Net it is ATExt. Can you use the same casing?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionRequestOptions.java
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * Gets the FMI (Federated Managed Identity) path for agent identity scenarios.
+     * When set, this indicates which agent identity the assertion is being requested for.
+     * The assertion provider can use this to include the FMI path in its token acquisition logic.
+     *
+     * @return the FMI path, or null if not applicable
+     */
+    public String fmiPath() {
```
neha-bhargava: Name it as `ClientAssertionFmiPath` to be consistent with https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/576cd9f41c448f2e0b48d46562d2ec386b18f764/src/client/Microsoft.Identity.Client/Internal/ClientCredential/CredentialContext.cs#L50

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
Avery-Dunn: An interface hierarchy and multiple new classes might be overkill and out of scope for this PR, it's an internal class and just has three options. 

This might be handled better by a more comprehensive refactoring to follow best practices, other parts of the internal design are also pretty messy.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CredentialTypeEnum.java
Avery-Dunn: The "atext" credential type comes from .NET's doc about cache extensibility: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/docs/cache_extensibility.md#token-schema-changes

As for the casing, I believe .NET, Java, and every other MSAL  just makes the cache key lowercase anyway: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5e571b661f0f837f0ea68842851fac1168b1e1d4/src/client/Microsoft.Identity.Client/Cache/Items/MsalCacheKeys.cs#L49

So it should make a difference, but for consistency with .NET and the other credential types I made it "ATExt" in the latest commit.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/FmiTest.java
Avery-Dunn: Fair enough, in the latest commit I added a new helper method in `TestHelper` to return the mocked instance discovery response (only the FMI tests are using it, and I'll refactor other tests that skip instance discovery separately)

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionRequestOptions.java
Avery-Dunn: In the latest commit it's now final

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionRequestOptions.java
Avery-Dunn: In the latest commit I've refactored `fmiPPath` and it's various references to be `clientAssertionFmiPath`

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -87,6 +90,32 @@ public IClientCredential clientCredential() {
...
+        return this.fmiPath;
+    }
+
+    /**
+     * Computes the extended cache key hash for this request's fmi_path, if set.
+     * Returns an empty string if fmiPath is not set.
+     * This is the single source of truth for the fmi_path cache key hash computation,
+     * used by both cache writes (TokenCache) and cache reads (silent lookup).
+     */
+    String computeFmiCacheKeyHash() {
```
neha-bhargava: Should this be abstracted to some method in case some other additional key needs to be added

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -87,6 +90,32 @@ public IClientCredential clientCredential() {
...
+        if (!StringHelper.isBlank(fmiPath)) {
+            java.util.TreeMap<String, String> components = new java.util.TreeMap<>();
```
neha-bhargava: Nit: `java.util.TreeMap` is fully qualified inline rather than imported

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
```diff
@@ -81,6 +81,28 @@ static boolean isNullOrBlank(final String str) {
         return str == null || str.trim().isEmpty();
     }
 
+    /**
+     * Computes an extended cache key hash from a sorted map of key-value components.
+     * Concatenates sorted key+value pairs, SHA-256 hashes, then Base64URL encodes without padding.
+     * This algorithm is cross-SDK compatible (same output for the same inputs in all MSAL SDKs).
```
neha-bhargava, line 87: Did you validate that the cache key used here matches one generated in .Net?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java
```diff
@@ -554,6 +586,15 @@ private Optional<AccessTokenCacheEntity> getApplicationAccessTokenCacheEntity(
                 .findAny();
     }
 
+    private boolean extCacheKeyHashMatches(AccessTokenCacheEntity accessToken, String expectedHash) {
```
neha-bhargava, line 589: This filter is the security-relevant one — it prevents an FMI-tagged cache entry from being returned to a non-FMI request and vice versa. The integration tests cover the happy paths but no test asserts the filter direction directly. Recommend a focused unit test that seeds the cache with an FMI entry and asserts a non-FMI request misses it (and the inverse).

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
neha-bhargava: I think this is a good reason to have this refactor along with this work. But in case this is time sensitive a follow up would be great

neha-bhargava (approved):
Some minor comments otherwise looks good

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
Avery-Dunn: Could you explain what you mean a bit more? This is an internal method which validates the FMI path isn't blank and calls the more generic `computeExtCacheKeyHash` that actually makes the hash for the cache key, so there's not a whole lot that can be abstracted.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/StringHelper.java
Avery-Dunn: Yes, there is a new test called `fmiPath_HashValueMatchesCrossSDK` which uses the same hardcoded hash values as some tests in .NET's `FmiIntegrationTests`: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/FmiIntegrationTests.cs

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/CredentialTypeEnum.java
```diff
@@ -6,6 +6,7 @@
 enum CredentialTypeEnum {
 
     ACCESS_TOKEN("AccessToken"),
+    ACCESS_TOKEN_EXTENDED("AText"),
```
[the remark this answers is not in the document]
Avery-Dunn: This was both a mistake in the PR description (MSAL .NET also uses "AText"), and something that was adjusted to match MSAL .NET based on a comment: https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1025#discussion_r3235877606

While we try to keep the public API surface consistent across MSALs, the described multi-MSAL persistent cache scenario is going to be very rare (i.e., sharing caches between apps in different programming languages), is not something we consider, and will almost certainly have more inconsistencies than this.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
```diff
@@ -92,6 +167,11 @@ public boolean equals(Object o) {
 
     @Override
     public int hashCode() {
+        // For context-aware providers, use the provider's identity hash code
+        if (contextAwareAssertionProvider != null) {
+            return System.identityHashCode(contextAwareAssertionProvider);
+        }
+
```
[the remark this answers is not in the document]
Avery-Dunn: No change needed, although this might be confusing it matches the existing pattern, and this is some inadvertent historical behavior which may not be great but would be difficult to change now.

This hash code is different from the one the library uses internally for its cache, so it should not affect the vast majority of customers.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
Avery-Dunn: Fixed in the latest commit.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenCache.java
Avery-Dunn: Added three focused cache isolation tests in FmiTest:
- fmiPath_CacheIsolation_FmiTokenNotReturnedForNonFmiRequest
- fmiPath_CacheIsolation_NonFmiTokenNotReturnedForFmiRequest
- fmiPath_CacheIsolation_DifferentFmiPathsNotShared

These directly assert the filter in both directions.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -162,12 +192,30 @@ public ClientCredentialParametersBuilder clientCredential(IClientCredential clie
...
+         * @param fmiPath the FMI path value (typically the agent application ID)
+         * @return builder that can be used to construct ClientCredentialParameters
+         */
+        public ClientCredentialParametersBuilder fmiPath(String fmiPath) {
+            if (fmiPath != null && fmiPath.trim().isEmpty()) {
+                throw new IllegalArgumentException("fmiPath cannot be empty or blank");
+            }
+            this.fmiPath = fmiPath;
+            return this;
+        }
```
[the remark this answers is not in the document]
Avery-Dunn: fmiPath(null) now throws IllegalArgumentException via ParameterValidationUtils.validateNotBlank(), matching MSAL .NET's WithFmiPath which rejects null/empty/whitespace. Having specific 'unset' behavior is not needed and was likely a hallucinated requirement, simply not using the .fmiPath() at all will do the same behavior in a more clear way.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -87,6 +90,32 @@ public IClientCredential clientCredential() {
...
+     * used by both cache writes (TokenCache) and cache reads (silent lookup).
+     */
+    String computeFmiCacheKeyHash() {
+        if (!StringHelper.isBlank(fmiPath)) {
+            java.util.TreeMap<String, String> components = new java.util.TreeMap<>();
+            components.put("fmi_path", fmiPath);
+            return StringHelper.computeExtCacheKeyHash(components);
+        }
+        return "";
+    }
```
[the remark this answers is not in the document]
Avery-Dunn: computeFmiCacheKeyHash() now memoizes the result in a fmiCacheKeyHashCache field, since ClientCredentialParameters is immutable after construction. 

This matches MSAL .NET's approach where the hash is computed once during InitCacheKey().

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -87,6 +90,32 @@ public IClientCredential clientCredential() {
...
+    /**
+     * Computes the extended cache key hash for this request's fmi_path, if set.
+     * Returns an empty string if fmiPath is not set.
+     * This is the single source of truth for the fmi_path cache key hash computation,
+     * used by both cache writes (TokenCache) and cache reads (silent lookup).
+     */
+    String computeFmiCacheKeyHash() {
+        if (!StringHelper.isBlank(fmiPath)) {
+            java.util.TreeMap<String, String> components = new java.util.TreeMap<>();
+            components.put("fmi_path", fmiPath);
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientAssertion.java
```diff
@@ -69,7 +113,33 @@ public String assertion() {
             }
         }
 
-        return this.assertion;
+        // Fall back to non-context-aware assertion
+        return assertion();
```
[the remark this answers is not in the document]
Avery-Dunn: Improved the Javadocs on both assertion() and assertion(AssertionRequestOptions) to explicitly document the
dispatch logic and fallback behavior. The re-entry is intentional — it ensures that a legacy Callable-based assertion always works regardless of whether the caller passes options — but the documentation now makes this clear.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialFactory.java
```diff
@@ -105,4 +106,23 @@ public static IClientAssertion createFromCallback(Callable<String> callable) {
...
+     * @return {@link ClientAssertion} that will invoke the function each time assertion() is called
+     * @throws NullPointerException if assertionProvider is null
+     */
+    public static IClientAssertion createFromCallback(Function<AssertionRequestOptions, String> assertionProvider) {
+        if (assertionProvider == null) {
+            throw new NullPointerException("assertionProvider");
+        }
+
+        return new ClientAssertion(assertionProvider);
+    }
```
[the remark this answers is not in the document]
Avery-Dunn: Updated the PR description after this comment.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
neha-bhargava: Basically can compute cache key with any other additional parameter that requires to be included in cache key

4gust (approved):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
Avery-Dunn: I see what you mean now, good catch: the agent definitely made this part too specific.

In the latest commit I've refactored it to be more generic and expandable.

neha-bhargava (approved):

PR #1040, opened 2026-07-01
Add SN/I certificate support over mTLS Proof-of-Possession (PoP)

Robbie-Microsoft:
## SN/I certificate over mTLS PoP

Confidential client presents its SN/I cert as the **client TLS cert** in the token-endpoint handshake → certificate-bound `token_type=mtls_pop` (`cnf`/`x5t#S256`) instead of a signed assertion. Opt-in, same credential; also covers FIC Leg 1. SNI+Bearer / broker SHR-PoP unchanged. Two-leg FIC stacked in **#1041**.

- **API**: `ClientCredentialParameters.Builder.mtlsProofOfPossession()`; new public `TokenType` (`BEARER`/`MTLS_POP`) + `BindingCertificate` (public material only); `AuthenticationResultMetadata.tokenType()` / `bindingCertificate()`.
- **Internal**: `MtlsClientCertificateHelper` + `MtlsEndpointHelper` (`login.*`→`mtlsauth.*`, region-optional, tenanted-only; legacy usgov+china fail fast); direct-cert path omits `client_assertion`; mTLS transport via injected `DefaultHttpClient`; AT cache keyed by `token_type`+cert KeyId.
- **Tests**: unit 382 green; CI-only E2E `MtlsPopIT` `Credential_X509_Output_{Pop,Bearer}` (+cache isolation) — PoP global asserts `MTLS_POP` and is proven usable by a strict-200 `mtls_pop`-scheme call to `mtlstb.graph.microsoft.com`; Bearer regional (westus3).


.github/copilot-instructions.md
```diff
@@ -137,6 +138,7 @@ MSAL4J supports multiple authentication flows, each with a public `*Parameters`
 - **Parameters**: `ClientCredentialParameters` - App-only authentication (daemon apps)
 - **Internal**: `ClientCredentialRequest` → `AcquireTokenByClientCredentialSupplier`
 - **Key Classes**: `IClientCredential`, `ClientSecret`, `ClientCertificate`, `ClientAssertion`
+- **mTLS Proof-of-Possession (PoP)**: Opt in with `ClientCredentialParameters.builder(...).mtlsProofOfPossession()` to obtain an mTLS-bound PoP token (`token_type=mtls_pop`). The app's SN/I `IClientCertificate` is presented as the client TLS certificate to a rewritten `mtlsauth.*` endpoint (no `client_assertion` on the direct path) instead of signing an x5c assertion (the existing SNI+Bearer path is unchanged). Requires a tenanted authority and an ESTS allow-listed resource; region is optional (global `mtlsauth.microsoft.com` when absent). The result exposes `metadata().tokenType()` and `metadata().bindingCertificate()` (public material only — x5c chain + `x5t#S256`).  - **Key Classes**: `TokenType`, `BindingCertificate`, `MtlsClientCertificateHelper`, `MtlsEndpointHelper`; internal `AuthScheme`. Not supported: US Gov / China clouds, and user-scoped (`user_fic`) FIC over mTLS.
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: Key Classes is now its own nested bullet.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -40,7 +42,7 @@ public class ClientCredentialParameters implements IAcquireTokenParameters {
     // Memoized hash of cacheKeyComponents (computed once since parameters are immutable).
     private String extCacheKeyHashCache;
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: corrected the comment. The hash is lazily memoized and invalidated when the cache-key components change, and both fields are now volatile.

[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: corrected the comment. The hash is lazily memoized and invalidated when the cache-key components change, and both fields are now volatile.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ClientCredentialParameters.java
```diff
@@ -114,6 +117,35 @@ public String fmiPath() {
...
+    void bindingCertificateKeyId(String keyId) {
+        if (StringHelper.isBlank(keyId)) {
+            return;
+        }
+        if (this.cacheKeyComponents == null) {
+            this.cacheKeyComponents = new TreeMap<>();
+        }
+        this.cacheKeyComponents.put("cert_kid", keyId);
+        this.extCacheKeyHashCache = null;
+    }
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: bindingCertificateKeyId is now synchronized and copy-on-write, and the cache-key fields are volatile.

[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: bindingCertificateKeyId is now synchronized and copy-on-write, and the cache-key fields are volatile.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -213,6 +267,13 @@ private AuthenticationResult createAuthenticationResultFromOauthHttpResponse(Htt
             }
             long currTimestampSec = new Date().getTime() / 1000;
 
+            TokenType tokenType = TokenType.BEARER;
+            BindingCertificate bindingCertificate = null;
+            if (isMtlsProofOfPossession()) {
+                tokenType = TokenType.MTLS_POP;
+                bindingCertificate = MtlsClientCertificateHelper.buildBindingCertificate(resolvedBindingCertificate);
+            }
```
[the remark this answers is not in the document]
Robbie-Microsoft: By design: on the direct mTLS path the token type is a property of the request (the caller opts into mtls_pop), not derived from the ESTS response. Leaving as-is.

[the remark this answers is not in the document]
Robbie-Microsoft: By design: on the direct mTLS path the token type is a property of the request (the caller opts into mtls_pop), not derived from the ESTS response. Leaving as-is.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MtlsClientCertificateHelper.java
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    private static List<X509Certificate> decodeCertificateChain(IClientCertificate certificate)
+            throws CertificateException {
+        List<X509Certificate> chain = new ArrayList<>();
+        CertificateFactory certificateFactory = CertificateFactory.getInstance("X.509");
+        for (String encoded : certificate.getEncodedPublicKeyCertificateChain()) {
+            byte[] der = Base64.getDecoder().decode(encoded);
+            chain.add((X509Certificate) certificateFactory.generateCertificate(new ByteArrayInputStream(der)));
+        }
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in d576db6b — `decodeCertificateChain` now catches the `IllegalArgumentException` from Base64 decoding and rethrows it as a `CertificateException`, so callers surface `MsalClientException` with `MTLS_POP_ERROR`. Added regression tests for a non-base64 x5c entry.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MtlsEndpointHelper.java
```diff
@@ -0,0 +1,183 @@
...
+    }
+
+    /**
+     * @return {@code true} if {@code host} is a known public (non-sovereign) Microsoft Entra host that
+     * should resolve to the global public mTLS host {@code mtlsauth.microsoft.com}.
+     */
+    private static boolean isPublicHost(String host) {
+        return AadInstanceDiscoveryProvider.TRUSTED_HOSTS_SET.contains(host)
+                && !AadInstanceDiscoveryProvider.TRUSTED_SOVEREIGN_HOSTS_SET.contains(host);
+    }
```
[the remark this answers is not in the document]
Robbie-Microsoft: By design, and pre-existing (not introduced here). `sts.windows.net` is a documented trusted public host — it's a v1 issuer identifier, never an actual token-endpoint host — so collapsing it to `mtlsauth.microsoft.com` carries no cross-cloud risk. Leaving as-is.

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
```diff
@@ -0,0 +1,205 @@
...
+ * general {@code LabVaultAppID}) is rejected with {@code AADSTS700025: Client is public...}. The lab cert
+ * ({@link KeyVaultSecretsProvider#CERTIFICATE_ALIAS}) is registered for SN/I on this app.
+ */
+@TestInstance(TestInstance.Lifecycle.PER_CLASS)
+class MtlsPopIT {
+
+    // SN/I-allow-listed app and MSI team tenant. mTLS PoP only works on this app/tenant pair; mirrors
+    // MSAL .NET's ClientCredentialsMtlsPopTests. App and tenant IDs are public identifiers, not secrets.
+    private static final String SNI_ALLOWLISTED_APP_ID = "163ffef9-a313-45b4-ab2f-c7e2f5e0e23e";
+    private static final String SNI_ALLOWLISTED_AUTHORITY =
```
gladjohn, line 66: we have so much reliance on this MSI team's test tenant. Can we have our own SNI test setup in IDLABS1?

changelog.txt
```diff
@@ -1,5 +1,6 @@
 Version 1.25.0
 =============
+- Add SN/I certificate support over mTLS Proof-of-Possession (PoP) for confidential clients, presenting the certificate as the client TLS cert to obtain an mTLS-bound token. The response token_type is validated, so a request that is not honored as mtls_pop fails closed rather than surfacing an unbound token
```
Avery-Dunn, line 3: The changes in this PR definitely weren't released a month ago.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/MtlsProofOfPossessionTest.java
```diff
@@ -0,0 +1,298 @@
...
+    void directSniCert_mtlsPop_targetsMtlsEndpoint_omitsClientAssertion_returnsMtlsPopToken() throws Exception {
+        ConfidentialClientApplication app = baseCertAppBuilder().build();
+
+        HttpRequest captured;
+        IAuthenticationResult result;
+        try (MockedConstruction<DefaultHttpClient> mocked = mockConstruction(DefaultHttpClient.class,
+                (m, ctx) -> when(m.send(any(HttpRequest.class))).thenReturn(successResponse("mtls-pop-token", "mtls_pop")))) {
+
+            result = app.acquireToken(ClientCredentialParameters.builder(Collections.singleton("https://graph.microsoft.com/.default"))
+                    .mtlsProofOfPossession()
```
gladjohn, line 116: is there a test that covers the bearer flow using bound creds?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/BindingCertificate.java
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * <p>This type exposes <b>public certificate material only</b> — the X.509 certificate chain
+ * ({@code x5c}) and the SHA-256 thumbprint ({@code x5t#S256}) of the leaf certificate. It never
+ * exposes the private key.
+ *
+ * <p>Instances are available via {@link AuthenticationResultMetadata#bindingCertificate()} for
+ * mTLS Proof-of-Possession results, and are {@code null} for Bearer results. For more details, see
+ * https://aka.ms/msal4j-pop
+ */
+public final class BindingCertificate implements Serializable {
```
gladjohn, line 24: **[Blocker/API]** This result exposes only public certificate information, so it cannot prove possession when calling the resource; the E2E consequently uses the original credential instead of the returned result. #1059 introduces `IMtlsBindingContext` with the exact certificate, key ID and reusable process-local `SSLContext`. Please converge on that as the authoritative binding capability and derive this diagnostics-only view from it.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
```diff
@@ -19,6 +19,17 @@ class AcquireTokenByClientCredentialSupplier extends AuthenticationResultSupplie
 
     @Override
     AuthenticationResult execute() throws Exception {
+        // For mTLS Proof-of-Possession, isolate the access token in the cache by the binding certificate's
```
gladjohn, line 22: **[Blocker]** `mtlsProofOfPossession()` can be combined with `appTokenProvider`. A provider token is returned without an mTLS request or response-token-type validation, saved under the certificate hash, and later forcibly relabelled `MTLS_POP` on a cache hit. Please reject this combination unless the provider contract supplies and validates an atomic binding context and token type.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByClientCredentialSupplier.java
```diff
@@ -19,6 +19,17 @@ class AcquireTokenByClientCredentialSupplier extends AuthenticationResultSupplie
...
+        // KeyId (x5t#S256) in addition to the token_type dimension, so PoP tokens bound to different
+        // certificates never alias. Stamped before the cache lookup so reads and writes hash identically.
+        IClientCertificate resolvedBindingCertificate = null;
```
gladjohn, line 25: **[Blocker]** This writes application-specific certificate identity into a public reusable parameters object. Two CCAs using certificates A and B can concurrently use the same parameters instance, causing one request to compute, read or save using the other application’s partition. Please keep public parameters immutable and carry an acquisition-local binding snapshot/cache key through lookup, execution and cache save.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ExtendedCacheKey.java
```diff
@@ -32,9 +37,30 @@ final class ExtendedCacheKey {
      * when there are none. The result is memoized.
      */
     String computeHash() {
```
gladjohn, line 39: **[Blocker]** `putComponent()` is synchronized but hash computation/publication is not. A thread can compute hash A, another thread can install component B and clear the cache, and the first thread can then republish stale hash A. Please eliminate post-construction mutation or synchronize/version the complete compute-and-publish operation.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -42,13 +48,34 @@ OAuthHttpRequest createOauthHttpRequest() throws MalformedURLException {
                     AuthenticationErrorCode.INVALID_ENDPOINT_URI);
         }
 
+        URL tokenEndpointUrl = requestAuthority.tokenEndpointUrl();
+        IHttpClient mtlsHttpClient = null;
+
+        if (isMtlsProofOfPossession()) {
+            ConfidentialClientApplication application = (ConfidentialClientApplication) msalRequest.application();
```
gladjohn, line 55: **[Blocker]** The binding is independently resolved for the cache key, TLS key manager and result metadata. A rotating or mutable credential can therefore produce certificate A for caching, B for the TLS handshake and C in the result. Please resolve one immutable binding context and use its key ID, certificate and `SSLContext` throughout the acquisition.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -42,13 +48,34 @@ OAuthHttpRequest createOauthHttpRequest() throws MalformedURLException {
...
+            this.resolvedBindingCertificate = resolveBindingCertificate(application);
+            tokenEndpointUrl = MtlsEndpointHelper.deriveMtlsTokenEndpoint(tokenEndpointUrl);
+            SSLSocketFactory mtlsSocketFactory =
+                    MtlsClientCertificateHelper.createMtlsSocketFactory(resolvedBindingCertificate);
+            mtlsHttpClient = new DefaultHttpClient(
```
gladjohn, line 60: **[Blocker]** Constructing a new `DefaultHttpClient` bypasses the configured `IHttpClient`, custom trust roots, proxy and egress controls, certificate pinning, interceptors, tracing and client-specific timeout/cancellation behavior. Please use #1059’s request-specific socket-factory plus `IMtlsCapableHttpClient` path and fail clearly when a custom client cannot honor mTLS.

gladjohn, line 60: **[Blocker/security]** This mTLS client retains `HttpsURLConnection`’s default redirect behavior. A same-protocol redirect can reconnect with the credential-bearing `SSLSocketFactory`, presenting the client certificate to another host; a 307 can also preserve the POST body. Disable redirects for all credential-bound token requests, matching #1059.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -42,13 +48,34 @@ OAuthHttpRequest createOauthHttpRequest() throws MalformedURLException {
...
         }
 
+        URL tokenEndpointUrl = requestAuthority.tokenEndpointUrl();
+        IHttpClient mtlsHttpClient = null;
+
+        if (isMtlsProofOfPossession()) {
+            ConfidentialClientApplication application = (ConfidentialClientApplication) msalRequest.application();
+            this.resolvedBindingCertificate = resolveBindingCertificate(application);
+            tokenEndpointUrl = MtlsEndpointHelper.deriveMtlsTokenEndpoint(tokenEndpointUrl);
+            SSLSocketFactory mtlsSocketFactory =
```
gladjohn, line 58: **[Blocker/availability]** A new `SSLContext`, `SSLSocketFactory` and client are created for every network request. This prevents keep-alive and TLS-session reuse and temporarily strands one-use pool entries, increasing handshake CPU, latency and socket pressure. Please use a bounded per-certificate context/transport cache with explicit idle cleanup.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -177,6 +206,12 @@ private void addCredentialToRequest(Map<String, String> queryParameters,
                 addJWTBearerAssertionParams(queryParameters, clientAssertion.assertion());
             }
         } else if (credentialToUse instanceof ClientCertificate) {
+            if (mtlsPoP) {
+                // For mTLS PoP (direct SN/I cert / FIC Leg 1), the certificate is presented as the client
+                // TLS certificate and authenticates the client; no client_assertion is sent (ESTS resolves
```
gladjohn, line 211: **[Blocker]** `extraQueryParameters` is merged after grant parameters, so callers can overwrite `token_type`, `grant_type` or `scope`, inject `req_cnf`, or inject a `client_assertion` into the pure-certificate branch. Please reject MSAL-owned OAuth and client-authentication parameters before network execution instead of warning and overwriting.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenType.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * @return the wire/string representation of this token type
+     */
+    public String value() {
+        return value;
+    }
+
+    /**
+     * Maps a token_type string (as returned in a token response) to a {@link TokenType}.
```
gladjohn, line 43: **[P1/API]** Every non-`mtls_pop` value—including SHR `pop` and future token schemes—is reported publicly as `BEARER`. Please preserve the raw service value, as #1059 does, or add explicit `POP` and `UNKNOWN` representations. Exact `mtls_pop` matching can remain for fail-closed validation.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MtlsEndpointHelper.java
```diff
@@ -0,0 +1,183 @@
...
+     *   {@code login.microsoft.com}, {@code sts.windows.net}) &rarr; {@code mtlsauth.microsoft.com}.</li>
+     *   <li>Any other {@code login.*} host &rarr; domain-preserving swap ({@code login} &rarr; {@code mtlsauth}),
+     *   keeping the request within the same cloud boundary (mirrors MSAL.NET).</li>
+     * </ul>
+     *
+     * @return the {@code mtlsauth.*} host, or {@code null} if {@code loginHost} is not a recognizable
+     * {@code login.*} host and therefore cannot be safely rewritten
+     */
+    static String deriveMtlsHost(String loginHost) {
+        String lower = loginHost.toLowerCase(Locale.ROOT);
```
gladjohn, line 129: **[Blocker]** Regional sovereign authorities arrive as `<region>.login.microsoftonline.us`, `<region>.login.partner.microsoftonline.cn`, etc. They do not match `.login.microsoft.com`, are not exact trusted-host entries and do not begin directly with `login.`, so this returns `null`. Parse the regional prefix separately, normalize the base authority and preserve both region and sovereign domain during rewriting.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MtlsEndpointHelper.java
```diff
@@ -0,0 +1,183 @@
...
+    private static final String REGIONAL_LOGIN_SUFFIX = ".login.microsoft.com";
+
+    private static final String LOGIN_LABEL = "login";
+    private static final String MTLS_LABEL = "mtlsauth";
+    private static final String LOGIN_LABEL_PREFIX = LOGIN_LABEL + ".";
+
+    // mTLS PoP is unsupported ONLY for these two legacy sovereign hosts (they have no mtlsauth.* endpoint).
+    // Every other login.* host — including Azure Government and the current national clouds — is allowed and
+    // rewritten domain-preserving. Mirrors MSAL.NET's s_unsupportedMtlsHosts denylist. Hosts are lowercase;
+    // callers lowercase the input before lookup.
```
gladjohn, line 53: **[Blocker/documentation]** This denylist runs after preferred-network normalization, which converts these legacy aliases to their modern hosts. The helper-only tests therefore do not represent normal acquisition behavior, and the stated “legacy aliases fail fast” contract is not enforced. Please reject the original authority before normalization or document and test alias normalization as supported.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ExtendedCacheKey.java
gladjohn, line 39: **[Blocker/rebase]** This head predates current `dev` commit `204bdd4`, which replaced delimiter-free `key + value` concatenation with UTF-8 length-prefixed serialization. The old encoding is ambiguous and this PR adds security-sensitive cache dimensions. Please rebase before validating certificate-cache isolation and update the expected hashes and tests.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResultMetadata.java
```diff
@@ -25,10 +25,28 @@ public class AuthenticationResultMetadata implements Serializable {
      */
     private CacheRefreshReason cacheRefreshReason = CacheRefreshReason.NOT_APPLICABLE;
 
+    /**
+     * The type of the access token in the {@link AuthenticationResult}, see {@link TokenType} for possible
+     * values. Defaults to {@link TokenType#BEARER}.
+     */
+    private TokenType tokenType = TokenType.BEARER;
```
gladjohn, line 32: **[Blocker/compatibility]** `IAuthenticationResult` is explicitly `Serializable`, but this public metadata type has no pinned UID. `serialver` changes from `6600853896042983641L` at the PR base to `4902369891410840671L` here, so previously serialized results fail after upgrade. Please pin the previous UID and add a previous-version deserialization fixture.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/MtlsProofOfPossessionTest.java
```diff
@@ -0,0 +1,298 @@
...
+        assertTrue(body.contains("client_assertion"), "Bearer SN/I path still sends a client_assertion");
+        assertTrue(body.contains("jwt-bearer"), "Bearer path uses the jwt-bearer client_assertion_type");
+        assertFalse(body.contains("token_type=mtls_pop"), "Bearer path must not request mTLS PoP");
+
+        assertEquals(TokenType.BEARER, result.metadata().tokenType());
+    }
+
+    @Test
+    void cacheKey_isolatesBearerFromMtlsPopAndByCertificate() {
+        ClientCredentialParameters bearer = ClientCredentialParameters
```
gladjohn, line 198: **[Acceptance blocker]** Comparing input hashes does not prove that cache writes, reads, coexistence and serialization preserve certificate partitions. Please exercise A-network → B-network → A-cache with real entries, serialize and restore into new clients, and add same-key/renewed-leaf coverage. Assert the token, source, token type and exact binding generation on every result.

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
```diff
@@ -0,0 +1,243 @@
...
+                .authority(SNI_ALLOWLISTED_AUTHORITY)   // tenanted authority (required for mTLS PoP)
+                .build();
+
+        IAuthenticationResult result = cca.acquireToken(ClientCredentialParameters
+                        .builder(Collections.singleton(AGENTIC_GRAPH_SCOPE))
+                        .mtlsProofOfPossession()
+                        .build())
+                .get();
+
+        assertMtlsPopResult(result, expectedLabThumbprint());
```
gladjohn, line 126: **[Acceptance blocker]** This resource call uses the original credential rather than the binding returned by MSAL, so it does not satisfy BIND-01/BIND-07. Return `IMtlsBindingContext`, use its exact `SSLContext` for the resource call, verify `cnf.x5t#S256`, and add rejection cells for no certificate, the wrong scheme and a different certificate.

gladjohn (changes requested):

PR #1041, opened 2026-07-06
Add FIC Leg 2 over mTLS Proof-of-Possession (stacked on #1040)

Robbie-Microsoft:
## Summary

Follow-up to #1040. Adds **2-leg Federated Identity Credential (FIC) over mTLS Proof-of-Possession (PoP)**: an app that authenticates with a federated **client assertion** can still **bind the resulting token to a certificate**. The assertion is sent with `client_assertion_type=...:jwt-pop` and the binding certificate is presented on the mTLS handshake, so the final token is bound to the Leg-1 certificate thumbprint.

> **Stacked on #1040** (base `rginsburg/sni-mtls-pop`), so the diff is exactly the Leg-2 additions on top of the core work. Retarget the base to `dev` after #1040 merges.

## Public API

```java
ConfidentialClientApplication.builder(clientId, ClientCredentialFactory.createFromClientAssertion(t1))
        .mtlsBindingCertificate(sniCert)   // binding cert presented on the TLS handshake
        .build();
```

- `ConfidentialClientApplication.Builder.mtlsBindingCertificate(IClientCertificate)`

## Internal changes

- **Request building** (`TokenRequestExecutor`): under mTLS PoP with an assertion credential, the assertion is sent with `client_assertion_type=...:jwt-pop` and the binding cert is presented on the mutual-TLS handshake.
- `MtlsClientCertificateHelper.resolveBindingCertificate` falls back to the configured `mtlsBindingCertificate(...)` **only when the credential is an `IClientAssertion`** (FIC Leg 2); a non-assertion credential (e.g. a client secret) with a stray binding cert fails with `MTLS_POP_ERROR`.
- `AssertionRequestOptions` carries the PoP flag through to assertion construction; adds the `ClientAssertion.ASSERTION_TYPE_JWT_POP` constant.

## Testing

- **Unit:** `ficLeg2_assertionWithBindingCert_usesJwtPopAssertionType` (`MtlsProofOfPossessionTest`); `resolveBindingCertificate` positive (assertion+cert) and negative (secret+cert → `MTLS_POP_ERROR`) tests (`MtlsClientCertificateHelperTest`). Suite green (**385 tests**).
- **E2E:** `acquireTokenFic_TwoLeg_MtlsPop_BothLegsBound` (`MtlsPopIT`, CI-only — needs the lab SN/I cert + an ESTS allow-listed resource).

## Out of scope

Unchanged from #1040: user-scoped (`user_fic`) FIC over mTLS PoP, managed-identity first-leg mTLS, and the two legacy sovereign hosts.


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -183,12 +184,18 @@ private void addCredentialToRequest(Map<String, String> queryParameters,
                 AssertionRequestOptions options = new AssertionRequestOptions(
                         application.clientId(),
                         tokenEndpoint,
-                        clientAssertionFmiPath);
+                        clientAssertionFmiPath,
+                        mtlsPoP);
```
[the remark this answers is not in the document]
Robbie-Microsoft: Acknowledged. This only affects the narrow context-aware-assertion provider + mTLS PoP intersection and is out of scope for this follow-up; can be addressed in a separate change if needed.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/MtlsProofOfPossessionTest.java
```diff
@@ -34,6 +34,8 @@ class MtlsProofOfPossessionTest {
     private static final String PKCS12_RESOURCE = "/mtls_test_cert.p12";
     private static final String PKCS12_PASSWORD = "password";
     private static final String AUTHORITY = "https://login.microsoftonline.com/contoso.onmicrosoft.com/";
+    private static final String JWT_POP_ASSERTION_TYPE =
+            "urn:ietf:params:oauth:client-assertion-type:jwt-pop";
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: the FIC Leg 2 test now uses the JWT_POP_ASSERTION_TYPE constant.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/MtlsProofOfPossessionTest.java
```diff
@@ -132,6 +134,47 @@ void directSniCert_mtlsPop_targetsMtlsEndpoint_omitsClientAssertion_returnsMtlsP
...
+            verify(mocked.constructed().get(0)).send(requestCaptor.capture());
+            captured = requestCaptor.getValue();
+        }
+
+        assertEquals("mtlsauth.microsoft.com", captured.url().getHost());
+
+        String body = captured.body();
+        assertTrue(body.contains("client_assertion=" + leg1Token), "Leg 2 must authenticate with the Leg-1 token");
+        assertTrue(body.contains("jwt-pop"), "Leg 2 must use the jwt-pop client_assertion_type");
+        assertTrue(body.contains("token_type=mtls_pop"));
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed: the test now parses the form body and asserts exact decoded values (client_assertion, client_assertion_type, token_type) instead of substring checks.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -21,6 +21,7 @@ public class ConfidentialClientApplication extends AbstractClientApplicationBase
 
     IClientCredential clientCredential;
     private boolean sendX5c;
+    IClientCertificate mtlsBindingCertificate;
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in 227a04c6 — the field is now `private` (it's only read through the `mtlsBindingCertificate()` getter).

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/MtlsClientCertificateHelper.java
```diff
@@ -56,9 +58,14 @@ static IClientCertificate resolveBindingCertificate(ConfidentialClientApplicatio
             return (IClientCertificate) credential;
         }
 
+        if (application.mtlsBindingCertificate() != null) {
+            return application.mtlsBindingCertificate();
+        }
```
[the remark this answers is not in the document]
Robbie-Microsoft: Fixed in 227a04c6 — the fallback is now restricted to `credential instanceof IClientAssertion`, so only assertion-authenticated apps (FIC Leg 2) pick up the binding cert. A secret (or any non-assertion credential) now fails with `MTLS_POP_ERROR`. Added a negative test.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ConfidentialClientApplication.java
```diff
@@ -139,6 +152,31 @@ public ConfidentialClientApplication.Builder sendX5c(boolean val) {
...
+         * is already an {@link IClientCertificate} (direct SN/I cert or FIC Leg 1), that same certificate is
+         * used as the binding certificate and this option is unnecessary.
+         * <p>
+         * Only the certificate's public material is ever surfaced on the result (see
+         * {@link AuthenticationResultMetadata#bindingCertificate()}); the private key is never exposed.
+         *
+         * @param val the binding certificate
+         * @return instance of the Builder on which method was called
+         */
+        public ConfidentialClientApplication.Builder mtlsBindingCertificate(IClientCertificate val) {
```
gladjohn, line 173: **[Blocker/API]** The assertion and binding certificate are configured through unrelated channels, so they can rotate independently and MSAL cannot prove they represent the same generation. This API also cannot consume #1059’s `IMtlsBindingContext` because it requires an exposed `PrivateKey`. Please replace it with an atomic callback result containing the assertion and exact reusable binding context.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -199,12 +200,18 @@ private void addCredentialToRequest(Map<String, String> queryParameters,
...
```
gladjohn, line 205: **[Blocker]** This derives both `client_assertion_type=jwt-pop` and output `token_type=mtls_pop` from one flag. These are independent dimensions: a binding-context-backed `jwt-pop` assertion can authenticate the client while the requested output token remains Bearer. Derive assertion type from the assertion’s binding material and output token type from the requested authentication scheme.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
```diff
@@ -183,6 +183,7 @@ private void addCredentialToRequest(Map<String, String> queryParameters,
         } else if (credentialToUse instanceof ClientAssertion) {
             // For client assertion, add client_assertion and client_assertion_type parameters
             ClientAssertion clientAssertion = (ClientAssertion) credentialToUse;
```
gladjohn, line 185: **[Blocker]** The request is sent to the resolved global or regional `mtlsauth` endpoint, but the assertion callback receives the normal `login.*` token endpoint. A provider using this value as `aud` will mint an assertion for a different endpoint than the one receiving it. Please pass the exact final endpoint used by `OAuthHttpRequest`.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AssertionRequestOptions.java
```diff
@@ -15,11 +15,17 @@ public final class AssertionRequestOptions {
     private final String clientId;
     private final String tokenEndpoint;
     private final String clientAssertionFmiPath;
+    private final boolean proofOfPossession;
```
gladjohn, line 18: **[P1/API]** A `proofOfPossession` boolean does not identify the certificate generation MSAL will put on the wire. The provider should atomically return a `ClientSignedAssertion` containing the assertion and `IMtlsBindingContext`, rather than returning only a string while relying on separately configured mutable application state.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/TokenRequestExecutor.java
gladjohn, line 185: **[Blocker]** Binding resolution accepts any public `IClientAssertion`, but request serialization handles only the package-private concrete `ClientAssertion`. A valid custom implementation can therefore resolve a binding and reach the endpoint without any `client_assertion` parameters. Please process the public interface consistently or introduce a separate public capability interface.

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
```diff
@@ -223,6 +230,85 @@ void Credential_X509_Output_Pop_And_Bearer_CacheIsolated() throws Exception {
...
+        } catch (ExecutionException e) {
+            if (e.getCause() instanceof MsalClientException
+                    && AuthenticationErrorCode.TOKEN_TYPE_MISMATCH.equals(
+                            ((MsalClientException) e.getCause()).errorCode())) {
+                Assumptions.abort("ESTS returned a non-mtls_pop token_type (downgrade); treating as "
+                        + "inconclusive: " + e.getCause().getMessage());
+            }
+            throw e;
+        }
+    }
```
gladjohn, line 310: **[Acceptance blocker]** This is the only two-leg FIC E2E, but `TOKEN_TYPE_MISMATCH` aborts it as inconclusive. CI can therefore pass without proving that the service honored the requested binding. Run this against a deterministic supported cell and fail on every downgrade.

msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
```diff
@@ -223,6 +230,85 @@ void Credential_X509_Output_Pop_And_Bearer_CacheIsolated() throws Exception {
...
+
+        IAuthenticationResult leg1 = acquireMtlsPopOrSkipOnDowngrade(blueprint, ClientCredentialParameters
+                        .builder(Collections.singleton(TestConstants.AGENTIC_TOKEN_EXCHANGE_SCOPE))
+                        .fmiPath(TestConstants.AGENTIC_AGENT_APP_ID)
+                        .mtlsProofOfPossession()
+                        .build());
+
+        assertNotNull(leg1, "Leg 1 result should not be null");
+        assertNotNull(leg1.accessToken(), "Leg 1 (T1) access token should not be null");
+        assertEquals(TokenType.MTLS_POP, leg1.metadata().tokenType(),
```
gladjohn, line 262: **[Acceptance blocker]** Comparing locally generated metadata to the configured lab certificate does not prove protocol continuity. Decode T1 and T2 and require both `cnf.x5t#S256` values to match the exact returned binding context’s key ID, then call the resource with T2 and that context. Add no-cert, wrong-scheme, wrong-cert, A→B→A and persistent-cache cases.

gladjohn (changes requested):

Issue #935, opened 2025-04-10
[Bug] HttpClient is not fault tolerant for network issues


JozefDropco:
### Library version used

1.15.0

### Java version

21

### Scenario

ConfidentialClient - web api (AcquireTokenOnBehalfOf)

### Is this a new or an existing app?

The app is in production, I haven't upgraded MSAL, but started seeing this issue

### Issue description and reproduction steps

At the moment DefaultHttpClient is not fault tolerant for network issues/glitches. It just handles the situation where HTTP 500 or above is thrown. This is causing troubles for all users not knowing they can "write" their own IHttpClient. This fact is forcing anyone to reinvent the wheel again and again. It would be beneficial if the basic implementation would be more robust to withstand these issues
e.g. Proxy failed with 503 but MSAL4J just failed instead of a retry mechanism to kick in from HttpHelper class
`com.microsoft.aad.msal4j.MsalClientException: java.io.IOException: Unable to tunnel through proxy. Proxy returns "HTTP/1.1 503 Service Unavailable"
	at com.microsoft.aad.msal4j.HttpHelper.executeHttpRequest(HttpHelper.java:56)
	at com.microsoft.aad.msal4j.OAuthHttpRequest.send(OAuthHttpRequest.java:48)
	at com.microsoft.aad.msal4j.TokenRequestExecutor.executeTokenRequest(TokenRequestExecutor.java:39)
	at com.microsoft.aad.msal4j.AbstractClientApplicationBase.acquireTokenCommon(AbstractClientApplicationBase.java:256)
	at com.microsoft.aad.msal4j.AcquireTokenByAuthorizationGrantSupplier.execute(AcquireTokenByAuthorizationGrantSupplier.java:63)
	at com.microsoft.aad.msal4j.AcquireTokenByClientCredentialSupplier.acquireTokenByClientCredential(AcquireTokenByClientCredentialSupplier.java:86)
	at com.microsoft.aad.msal4j.AcquireTokenByClientCredentialSupplier.execute(AcquireTokenByClientCredentialSupplier.java:49)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:69)
	at com.microsoft.aad.msal4j.AuthenticationResultSupplier.get(AuthenticationResultSupplier.java:18)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run(Unknown Source)
[...]

PR #1050, opened 2026-07-27
bugfix #1019 - wrong throttling on clientid.

### LINKED ISSUE #1019: [Bug] Throttling cache affects whole clientId when user provides wrong password in ADFS federation
### Library version used

1.24.0

### Java version

21.0.10

### Scenario

PublicClient (AcquireTokenInteractive, AcquireTokenByUsernamePassword)

### Is this a new or an existing app?

The app is in production, and I have upgraded to a new version of MSAL

### Issue description and reproduction steps

Scenario where multiple users use application with same client id. If some user provides wrong password and 500 error response is returned (we have seen in our case ID3242), then all other users under same client id are affected for any token calls. In our case it was accented by service account performing a lot of operations and a user entering wrong password completely blocked the service account for several seconds.

Relevant Issues from the past: https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/176 https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/721

Potential solution could be adding a useridentifier from request context (if it contains it) into the request thumbprint, thus removing the impact on other users if such situation occurs (since ThrottlingCache is globally shared).

### Relevant code snippets

```csharp

```

### Expected behavior

Wrong password provided should not affect other users except the one who provided wrong password.

### Identity provider

Azure Active Directory Federation Services (ADFS)

### Regression

_No response_

### Solution and workarounds

None known.

gterminator:
Hi,
here is my fix to #1019.

This fix is depending on the (existing) HttpHelper Class. 
Which i refactored on another PR  #935. 

fixes #1019

open for feedback.

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
@@ -125,18 +125,19 @@ class HttpHelper implements IHttpHelper {
 
     private String getRequestThumbprint(RequestContext requestContext) {
         StringBuilder sb = new StringBuilder();
-        sb.append(requestContext.clientId() + POINT_DELIMITER);
-        sb.append(requestContext.authority() + POINT_DELIMITER);
+        sb.append(requestContext.clientId()).append(POINT_DELIMITER);
+        sb.append(requestContext.authority()).append(POINT_DELIMITER);
 
-        IAcquireTokenParameters apiParameters = requestContext.apiParameters();
-
-        if (apiParameters instanceof SilentParameters) {
-            IAccount account = ((SilentParameters) apiParameters).account();
-            if (account != null) {
-                sb.append(account.homeAccountId() + POINT_DELIMITER);
+        UserIdentifier userIdentifier = requestContext.userIdentifier();
+        if (userIdentifier != null) {
+            if (!StringHelper.isBlank(userIdentifier.upn())) {
+                sb.append(userIdentifier.upn()).append(POINT_DELIMITER);
+            } else if (!StringHelper.isBlank(userIdentifier.oid())) {
+                sb.append(userIdentifier.oid()).append(POINT_DELIMITER);
             }
         }
 
+        IAcquireTokenParameters apiParameters = requestContext.apiParameters();
         Set<String> sortedScopes = new TreeSet<>(apiParameters.scopes());
         sb.append(String.join(" ", sortedScopes));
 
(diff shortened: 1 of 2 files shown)


Avery-Dunn:
Thanks for making this fix @gterminator! Overall it looks good and we'll likely be merging it in, however first we're going to do some investigation into the other MSALs (.NET, Python, Go, Node) to make sure 1) this behavior is consistent with our other SDKs and 2) the other SDKs don't have the same issue/need the same fix.

I don't have an ETA, but this should get into the next MSAL Java release.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpHelper.java
```diff
@@ -125,18 +125,19 @@ IHttpResponse executeHttpRequest(HttpRequest httpRequest) {
 
     private String getRequestThumbprint(RequestContext requestContext) {
```
Avery-Dunn, line 126: Adding per-user throttling should help with the issues mentioned in #1019, and avoid a user's bad password or 500 errors from blocking other users.

However, I believe this `getRequestThumbprint` is used for all throttling scenarios in public client flows, not just user-specific ones. If an app is getting 429 errors then all users should be throttled, however with these changes each individual user will need to receive a 429 error to be added to the throttling cache.

In short, this will relieve user-specific throttling issues, but may exacerbate app-wide throttling issues.

gterminator: > Adding per-user throttling should help with the issues mentioned in #1019, and avoid a user's bad password or 500 errors from blocking other users.
> 
> However, I believe this `getRequestThumbprint` is used for all throttling scenarios in public client flows, not just user-specific ones. If an app is getting 429 errors then all users should be throttled, however with these changes each individual user will need to receive a 429 error to be added to the throttling cache.
> 
> In short, this will relieve user-specific throttling issues, but may exacerbate app-wide throttling issues.

good point. I will rework this PR.

Avery-Dunn: I actually just went and created an updated version in MSAL Java, and two other MSALs that seemed to have a similar issue:
https://github.com/AzureAD/microsoft-authentication-library-for-java/pull/1055
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6159
https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/8756

I created the separate Java PR to fix that issue and add some extra tests as I do some more thorough checks in each of the SDKs, and allow it to run on our pipelines just to make sure there aren't any unexpected side effects.

gterminator: > I actually just went and created an updated version in MSAL Java, and two other MSALs that seemed to have a similar issue: #1055 [AzureAD/microsoft-authentication-library-for-dotnet#6159](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6159) [AzureAD/microsoft-authentication-library-for-js#8756](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/8756)
> 
> I created the separate Java PR to fix that issue and add some extra tests as I do some more thorough checks in each of the SDKs, and allow it to run on our pipelines just to make sure there aren't any unexpected side effects.

perfect then i will close this pr.

PR #1051, opened 2026-07-30
#1028 validate region string format before use in URL construction

### LINKED ISSUE #1028: Validate region string format before use in URL construction
The `discoverRegion` method in `AadInstanceDiscoveryProvider.java` returns the region string from either the `REGION_NAME` environment variable or the IMDS endpoint response without validating its format. This region is then used in `getRegionalizedHost` to construct authority URLs via string replacement (e.g., `{region}.login.microsoft.com`).

If the region string contains unexpected characters (dots, slashes, etc.), the resulting URL could be malformed and lead to failed or misdirected requests.

Azure region names follow a consistent pattern of lowercase alphanumeric characters and hyphens (e.g., `eastus`, `westus2`, `east-us-2`).

**Proposed fix:** Validate the region string against a pattern like `^[a-z][a-z0-9-]*$` at discovery time (in `discoverRegion`) and treat invalid values as if no region was detected.

**Reference:** MSAL .NET already validates regions via `RegionManager.ValidateRegion()`. MSAL Go added validation in https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/625.

gterminator:
this should fix #1028

### DIFF
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
@@ -15,6 +15,7 @@ import java.util.TreeSet;
 import java.util.Map;
 import java.util.HashMap;
 import java.util.concurrent.*;
+import java.util.regex.Pattern;
 
 class AadInstanceDiscoveryProvider {
 
@@ -26,6 +27,10 @@ class AadInstanceDiscoveryProvider {
     private static final String SOVEREIGN_HOST_TEMPLATE_WITH_REGION = "{region}.{host}";
     private static final String REGION_NAME = "REGION_NAME";
     private static final int PORT_NOT_SET = -1;
+
+    //Azure region names are lowercase alphanumeric characters and hyphens, starting with a letter (westus, east-us-2, etc.).
+    //Regions are used to build authority hosts, so anything outside of that set could produce a malformed URL
+    private static final Pattern VALID_REGION = Pattern.compile("^[a-z][a-z0-9-]*$");
 
     // For information of the current api-version refer: https://docs.microsoft.com/en-us/azure/virtual-machines/windows/instance-metadata-service#versioning
     private static final String DEFAULT_API_VERSION = "2021-02-01";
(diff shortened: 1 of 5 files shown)


gterminator:
@Avery-Dunn @bgavrilMS need a review.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -303,10 +334,19 @@ static String discoverRegion(MsalRequest msalRequest, ServiceBundle serviceBundl
...
         if (System.getenv(REGION_NAME) != null) {
-            LOG.info("Region found in environment variable: {}", System.getenv(REGION_NAME));
+            String region = System.getenv(REGION_NAME);
+            LOG.info("Region found in environment variable: {}", region);
+
+            //A region is used to build an authority host, so an unexpected value is rejected rather than
+            //  used to make requests against a malformed URL
+            if (!isValidRegion(region)) {
+                currentRequest.regionSource(RegionTelemetry.REGION_SOURCE_FAILED_AUTODETECT.telemetryValue);
+                throw invalidRegionException(region);
```
bgavrilMS: Invalid auto-detected input should behave like failed discovery rather than aborting token acquisition. Please log and return `null` for malformed `REGION_NAME`/IMDS values so MSAL falls back to the global endpoint, as described in #1028.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
```diff
@@ -186,13 +193,37 @@ static void cacheRegionInstanceMetadata(String originalHost, String regionalHost
...
+
     private static String getRegionalizedHost(String host, String region) {
         String regionalizedHost;
 
         if (region == null){
             return host;
         }
 
+        //Last check before the region is used to build a host: regions from any source, such as one set by an
+        //  app developer, must follow the Azure region naming convention
```
bgavrilMS: Could we validate the region once when it enters application state instead of running `VALID_REGION.matcher(...).matches()` during every regional host lookup? Validate explicit builder input during construction and environment/IMDS values when read, then keep `azureRegion` as an already-validated value.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AadInstanceDiscoveryProvider.java
gterminator: changed.

gterminator:
@bgavrilMS done.

May we can (in the future) think about how the discovery is failed. Don't like that we returning null as a business logic.

bgavrilMS (approved):

gterminator:
@bgavrilMS fixed merge conflicts.

Avery-Dunn (approved):

Avery-Dunn:
Looks good to me, thanks for fixing https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/1028! This will be part of our next release.

PR #1054, opened 2026-07-31
Add Bearer-over-mTLS (sendCertificateOverMtls) for confidential clients

Robbie-Microsoft:
## Bearer-over-mTLS (`sendCertificateOverMtls`)

Ports MSAL.NET's `CertificateOptions.SendCertificateOverMtls` (.NET PRs #5849 client-credentials, #6009 OBO/refresh/auth-code).

**What it is:** present the app's SN/I certificate as the client certificate on the mTLS handshake to the token endpoint, route to the mTLS endpoint (`mtlsauth.*`), and receive a plain **Bearer** access token. The certificate authenticates the transport; the token is **not** bound to it.

**How it differs from existing surfaces:**
- **NOT** mTLS PoP (Task 1, #1040): that binds the token (`token_type=mtls_pop`) and fences the cache by certificate thumbprint.
- **NOT** the existing SNI + `private_key_jwt` Bearer path: that signs an assertion to the **regular** token endpoint; the cert is never on the TLS handshake. Bearer-over-mTLS = cert **on** the handshake → Bearer from the **mTLS** endpoint.

### API
`ConfidentialClientApplication.Builder.sendCertificateOverMtls(boolean)` — app-level, defaults `false` (zero behavior change when unset). `build()` fails fast with `MsalClientException` (`AuthenticationErrorCode.CERTIFICATE_REQUIRED_FOR_MTLS`) if the flag is set without a certificate credential.

### Behavior
- Routes to the mTLS endpoint reusing the Task-1 SNI mTLS transport (no re-architecture).
- Keeps `token_type=Bearer` (no `token_type=mtls_pop`, no `req_cnf`); **forces** the x5c chain on the `client_assertion` (regardless of `app.sendX5c()`) so ESTS can do SN/I subject+issuer matching over the mTLS channel; produces a plain (not thumbprint-fenced) Bearer cache entry.
- A per-request `.mtlsProofOfPossession()` always takes precedence over the app-level flag.
- Honored by every confidential flow: client credentials, on-behalf-of, refresh token, authorization code (the flag is read off the application, not a request cast).

### Tests
- **Unit** (`BearerOverMtlsTest`, 13 cells via `mockConstruction(DefaultHttpClient)`): config (default / stored / non-cert fail-fast); client credentials (global `mtlsauth` + Bearer + `client_assertion` with forced x5c and no `mtls_pop`/`req_cnf`; regional; per-request `mtls_pop` wins; plain-Bearer cache); user flows (OBO / refresh / auth-code route to `mtlsauth` with the correct grant + forced-x5c assertion; OBO 2nd call served from cache; OBO without the flag uses the regular login endpoint).
- **Integration** (`MtlsPopIT`): live client-credentials Bearer-over-mTLS cell + cache-hit (SN/I-allow-listed app); a skip-gated OBO cell documenting the user-flow allow-listing gap (AADSTS700027 / AADSTS392189). The token-endpoint wire contract is asserted at the unit level because the mTLS `DefaultHttpClient` is constructed internally (no factory injection point).

### PR topology
Child branch off `rginsburg/sni-mtls-pop` (#1040), **targeting the SNI branch** — a **sibling** to the FIC follow-up #1041, **not** stacked on it. Retarget to `dev` after #1040 merges. Does not touch the FIC test or its skip-hatch.

Draft until #1040 merges to `dev`.

### DIFF
--- a/msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
+++ b/msal4j-sdk/src/integrationtest/java/com/microsoft/aad/msal4j/MtlsPopIT.java
@@ -40,6 +40,10 @@ import static org.junit.jupiter.api.Assertions.assertNull;
  * <p>The primary scenario is covered:
  * <ul>
  *   <li><b>Direct SNI cert &rarr; mTLS PoP</b> (client credentials), global and regional endpoints.</li>
+ *   <li><b>Bearer-over-mTLS</b> (Task 3): the same SNI cert is presented on the TLS handshake to the mTLS
+ *       token endpoint via {@code sendCertificateOverMtls(true)}, but a plain (unbound) {@code Bearer}
+ *       token is returned. Live client-credentials cell + a skip-gated OBO cell documenting the
+ *       user-flow allow-listing gap.</li>
  * </ul>
  *
  * <p><b>Testability gate (SME note A):</b> ESTS gates mTLS PoP on the <i>final resource audience</i>,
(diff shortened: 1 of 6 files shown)


gterminator:
Hi,
looks fine. 

I have only one problem. We are adding and adding more functionality in our tokenrequestexecutor. This will lead to a massive degeneration of our code in the future when everybody does that.

How about we do a refactoring after this PR and use the composite pattern to have a functionality based class like mTLS that add functionality to an object?

@Robbie-Microsoft How is your opinion on that?

bgavrilMS:
Why not have the same CertificateOptions property bag like in .NET ?

gterminator:
> Why not have the same CertificateOptions property bag like in .NET ?

sounds like a great idea. I didn't looked into the .NET code yet but maybe this is a better solution than this here :D

Issue #644, opened 2023-05-02
Create sample and documentation for WAM


siddhijain:

Avery-Dunn:
Duplicate of https://github.com/AzureAD/microsoft-authentication-library-for-java/issues/720

PR #1057, opened 2026-08-10
Support Azure Arc user-assigned managed identity (fail closed)

gladjohn:
## Summary

Adds **Azure Arc user-assigned managed identity (UAMI)** support to MSAL Java, bringing it to parity with MSAL .NET ([#6128](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6128)) and MSAL Go ([#644](https://github.com/AzureAD/microsoft-authentication-library-for-go/pull/644)).

Previously `AzureArcManagedIdentitySource` rejected any user-assigned identity at construction. Azure Arc agents that support UAMI accept a `client_id` / `object_id` / `msi_res_id` selector on the token request and echo the identity they used in the response.

## Changes

- **Forward the selector** on the Arc request (`client_id` / `object_id` / `msi_res_id`). Remove the `USER_ASSIGNED_MANAGED_IDENTITY_NOT_SUPPORTED` rejection.
- **Use `msi_res_id` (not `mi_res_id`) for Azure Arc** resource-id selection. A live Azure Connected Machine agent silently ignores `mi_res_id` and falls back to the system-assigned identity; it honors `msi_res_id`.
- **Fail closed**: after a 200, verify the echoed identity matches the requested selector (case-insensitive; `mi_res_id` accepted as a safety net). If it is missing or does not match, throw `USER_ASSIGNED_MANAGED_IDENTITY_NOT_CONFIRMED` so MSAL never returns a token for a different identity than requested. System-assigned is unaffected; genuine service errors surface as before.
- `ManagedIdentityResponse` now parses the `object_id` / `msi_res_id` / `mi_res_id` echo fields.

## Tests

`ManagedIdentityTests` (nested `AzureArc`): Arc is no longer in the "user-assigned not supported" set; added honored `client_id` / `object_id`, `msi_res_id` forwarding, and fail-closed cases. **11/11 Arc tests pass** (`mvn -pl msal4j-sdk test`).

## Note

Like the .NET/Go/JS PRs, this is gated on broad Azure Connected Machine agent support for UAMI

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,6 @@ Version 1.25.1
 - Add claimsFromClient API for client-originated claims on confidential-client flows (#1039)
 - Migrate region discovery to IMDS /compute JSON endpoint (#1038)
 - Bump Azure Arc HIMDS api-version from 2019-11-01 to 2020-06-01 (#1045)
-- Support Azure Arc user-assigned managed identity, forwarding the client_id/object_id/msi_res_id selector and failing closed if the token response does not confirm the requested identity (#1057)
 - Extract shared ExtendedCacheKey helper to de-duplicate cache-key plumbing (#1046)
 - Fix refreshed tokens not staying in their client-claims cache partition (#1039)
 - Fix extended cache-key hash collision by using a length-prefix encoding of the sorted components (#1047)
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@ Version 1.25.1
 - Add claimsFromClient API for client-originated claims on confidential-client flows (#1039)
 - Migrate region discovery to IMDS /compute JSON endpoint (#1038)
 - Bump Azure Arc HIMDS api-version from 2019-11-01 to 2020-06-01 (#1045)
+- Support Azure Arc user-assigned managed identity, forwarding the client_id/object_id/msi_res_id selector and failing closed if the token response does not confirm the requested identity (#1057)
 - Extract shared ExtendedCacheKey helper to de-duplicate cache-key plumbing (#1046)
 - Fix refreshed tokens not staying in their client-claims cache partition (#1039)
 - Fix extended cache-key hash collision by using a length-prefix encoding of the sorted components (#1047)
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
@@ -174,9 +174,20 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
             case OBJECT_ID:
                 echoed = response.getObjectId();
                 break;
-            case RESOURCE_ID:
-                echoed = response.getResourceId();
+            case RESOURCE_ID: {
+                // Azure Arc returns the resource id under msi_res_id; mi_res_id is accepted as a
+                // safety net. If both aliases are present and disagree, the echo cannot be trusted,
+                // so fail closed rather than accept one alias that matches while the other contradicts it.
+                String msiResId = response.getMsiResId();
+                String miResId = response.getMiResId();
+                if (!StringHelper.isNullOrBlank(msiResId) && !StringHelper.isNullOrBlank(miResId)
+                        && !msiResId.equalsIgnoreCase(miResId)) {
+                    echoed = null;
+                } else {
+                    echoed = !StringHelper.isNullOrBlank(msiResId) ? msiResId : miResId;
+                }
                 break;
+            }
             default:
                 echoed = null;
                 break;
--- a/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
+++ b/msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
@@ -64,13 +64,6 @@ class AzureArcManagedIdentitySource extends AbstractManagedIdentitySource{
     private AzureArcManagedIdentitySource(URI endpoint, MsalRequest msalRequest, ServiceBundle serviceBundle){
         super(msalRequest, serviceBundle, ManagedIdentitySourceType.AZURE_ARC);
         this.MSI_ENDPOINT = endpoint;
-
-        ManagedIdentityIdType idType =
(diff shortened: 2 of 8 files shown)


msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
```diff
@@ -40,6 +42,13 @@ public static ManagedIdentityResponse fromJson(JsonReader jsonReader) throws IOE
                     case "client_id":
                         response.clientId = reader.getString();
                         break;
+                    case "object_id":
+                        response.objectId = reader.getString();
+                        break;
+                    case "msi_res_id":
+                    case "mi_res_id":
+                        response.resourceId = reader.getString();
```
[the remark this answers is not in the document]
gladjohn: Fixed in `4349bf7`. `ManagedIdentityResponse` now parses `msi_res_id` and `mi_res_id` into **separate** fields (`msiResId` / `miResId`) instead of collapsing them into one order-dependent field. In `verifyUserAssignedIdentityWasHonored` the resource-id case now fails closed when both aliases are present and disagree, rather than letting a matching alias rescue a contradictory canonical echo. Added tests that assert fail-closed for conflicting aliases in **both** JSON orders (`userAssignedResourceId_ConflictingAliasesMsiResIdFirst_FailsClosed` / `...MiResIdFirst_FailsClosed`), plus a missing-echo fail-closed test and a test that the `mi_res_id` alias is accepted when it matches.

msal4j-sdk/src/test/java/com/microsoft/aad/msal4j/ManagedIdentityTests.java
```diff
@@ -906,6 +907,82 @@ void invalidPathWithRealFile(String authHeaderKey)
...
 
+        @Test
+        void userAssignedClientId_Honored() throws Exception {
+            ManagedIdentityId id = ManagedIdentityId.userAssignedClientId(ManagedIdentityTestConstants.CLIENT_ID);
+            setUpCommonTest(AZURE_ARC, ManagedIdentityTestConstants.AZURE_ARC_ENDPOINT, id);
+
+            when(httpClientMock.send(expectedRequest(AZURE_ARC, ManagedIdentityTestConstants.RESOURCE, id)))
+                    .thenReturn(expectedResponse(HttpStatus.HTTP_OK,
+                            getSuccessfulResponseWithUserAssignedId(ManagedIdentityTestConstants.RESOURCE,
+                                    "client_id", ManagedIdentityTestConstants.CLIENT_ID)));
```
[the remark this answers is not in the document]
gladjohn: Fixed in `4349bf7`. Added Windows-guarded tests that exercise the real Azure Arc two-request flow (initial request → 401 + `WWW-Authenticate` secret-file challenge → authenticated second request → 200):
- `userAssignedResourceId_ChallengeFlow_RetainsMsiResIdOnAuthenticatedRequest_Honored` — captures the second request and asserts it still carries `msi_res_id=` (and not `mi_res_id=`), and that a matching echo returns the token.
- `userAssigned_ChallengeFlow_MismatchedEcho_FailsClosed` — a mismatched echo on the authenticated second response fails closed.

These use a real `.key` secret file under `%ProgramData%\AzureConnectedMachineAgent\Tokens` (created/cleaned up per test), so the production 401→retry path — not just the direct-200 path — is now covered. All 17 AzureArc tests pass.

changelog.txt
```diff
@@ -3,6 +3,7 @@ Version 1.25.1
 - Add claimsFromClient API for client-originated claims on confidential-client flows (#1039)
 - Migrate region discovery to IMDS /compute JSON endpoint (#1038)
 - Bump Azure Arc HIMDS api-version from 2019-11-01 to 2020-06-01 (#1045)
+- Support Azure Arc user-assigned managed identity, forwarding the client_id/object_id/msi_res_id selector and failing closed if the token response does not confirm the requested identity (#1057)
```
Avery-Dunn: Unless we're planning to go back in time to the last release, this changelog update seems like a mistake.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
```diff
@@ -144,10 +143,63 @@ public ManagedIdentityResponse handleResponse(
...
+        switch (idType) {
+            case CLIENT_ID:
+                echoed = response.getClientId();
+                break;
+            case OBJECT_ID:
+                echoed = response.getObjectId();
+                break;
+            case RESOURCE_ID: {
+                // Azure Arc returns the resource id under msi_res_id; mi_res_id is accepted as a
+                // safety net. If both aliases are present and disagree, the echo cannot be trusted,
```
Avery-Dunn: > If both aliases are present and disagree, the echo cannot be trusted

As far as I can tell from the other PRs, the other MSALs simply prefer `msi_res_id` and ignore conflicting `mi_res_id`, and only MSAL Java considers conflicting values untrustworthy. Is this intentional?

Avery-Dunn (changes requested):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AzureArcManagedIdentitySource.java
gladjohn: Good call — not intentional, and you're right that it made Java the outlier. Fixed in `c4bcc1d` to match the other MSALs: `verifyUserAssignedIdentityWasHonored` now prefers `msi_res_id` and falls back to `mi_res_id`, with no conflict-rejection:

`echoed = !StringHelper.isNullOrBlank(response.getMsiResId()) ? response.getMsiResId() : response.getMiResId();`

This mirrors MSAL .NET (both spellings map to one `ResourceId`), Go (`[msi_res_id, mi_res_id]` first non-empty), JS (`msi_res_id || mi_res_id`), and Python (`msi_res_id or mi_res_id`). A wrong canonical `msi_res_id` still fails closed even if a `mi_res_id` alias matches (we prefer the canonical value), so behavior is identical to the other four in every real Arc scenario.

changelog.txt
gladjohn: You're right — removed in `c4bcc1d`. The entry was mistakenly added under the already-released `1.25.1` section; I'll leave the changelog for the maintainers to place under the correct release when this is cut (the PR is gated on broad Connected Machine agent support anyway).

Avery-Dunn (approved):

PR #1059, opened 2026-08-16
Add Managed Identity v2 KeyGuard mTLS PoP support

gladjohn:
## Summary

Adds production-shaped Managed Identity v2 mTLS support for Java using Windows
VBS KeyGuard and optional Microsoft Azure Attestation.

Java retains native interop only for KeyGuard/CNG signing and attestation.
JCA/JSSE owns TLS, and mTLS PoP callers receive a reusable
`IMtlsBindingContext` for independent downstream Java HTTP calls.

**Reviewer entry point:** [`managed-identity-v2-mtls-pop-review-guide.md`](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/gladjohn/java-msi-v2-mtls-binding-context-local/msal4j-sdk/docs/managed-identity-v2-mtls-pop-review-guide.md)

## Architecture

```text
ManagedIdentityApplication
  -> IMDS v2 platform metadata
  -> KeyGuard non-exportable RSA key
  -> optional MAA attestation
  -> IMDS v2 binding certificate
  -> normal MSAL OAuth token pipeline over JSSE mTLS
  -> mtls_pop token + reusable IMtlsBindingContext
```

Native signing remains limited to:

```text
JSSE
  -> X509ExtendedKeyManager
  -> CngRsaPrivateKey
  -> CngSignatureSpi
  -> NCryptSignHash
  -> VBS KeyGuard
```

## Changes

- Adds managed-identity mTLS PoP and binding-aware token caching keyed by token
  type, complete leaf-certificate DER SHA-256 identity, and attestation mode.
- Routes the ESTS mTLS exchange through the normal MSAL token pipeline for
  claims, client capabilities, telemetry, correlation IDs, retries, error
  handling, and response parsing.
- Adds an optional Windows extension for CNG, KeyGuard, CSR generation,
  attestation, certificate rotation, and reusable JSSE binding contexts.
- Bundles and verifies the Microsoft-signed
  `Microsoft.Azure.Security.KeyGuardAttestation` 1.1.5
  `AttestationClientLib.dll`.
- Requires exact `token_type=mtls_pop`, HTTPS token endpoints, IMDS-specific
  retry behavior, validated IMDS response origin, and disabled redirects for
  credential-bound requests.
- Defines `IMtlsCapableHttpClient` for custom transports and fails fast when a
  custom client does not consume request-specific mTLS configuration.
- Detects and recreates stale per-boot KeyGuard keys using a private signing
  liveness probe.
- Keeps the manual Key Vault E2E application in a separate profile-only module.

## .NET parity follow-up

The following parity changes are included in this PR branch:

- Moves the public attestation opt-in out of core and into the optional
  `com.microsoft.azure:msal4j-key-attestation` artifact through
  `ManagedIdentityAttestationExtensions.withAttestationSupport(...)`, matching
  the MSAL.NET package boundary.
- Adds bearer-over-mTLS through `withRequestOverMtls()`. It authenticates the
  ESTS connection with KeyGuard, requests `token_type=bearer`, uses a distinct
  cache partition, and intentionally returns no downstream binding context.
- Adds SLF4J-backed native attestation logging with correlation IDs, safe native
  callback lifetime, and suppression of raw native debug payloads.
- Adds `MSAL_MI_DISABLE_IMDS_V2`; `true` or `1` disables v2, capability
  discovery reports no mTLS binding, and explicit PoP or bearer-over-mTLS
  requests fail before metadata, key, CSR, certificate, or attestation work.
- Keeps the KeyGuard key container persistent but the issued certificate and
  binding context process-memory only. No Windows certificate-store persistence
  is required.
- Does not add token-revocation reminting, claims-triggered certificate
  replacement, or `bypass_cache`.

The optional bounded capability-discovery timeout proposed in MSAL.NET issue
#6180 remains a separate follow-up. It is useful for credential-chain failover,
but is not currently shipped by MSAL.NET and requires a Java cancellation and
transport-budget contract rather than a superficial future timeout.

## Validation

- Full `msal4j-sdk` tests pass.
- Targeted optional-package API and native logger tests pass.
- Production classes remain Java 8 compatible.
- The production extension and shaded E2E JAR contain the verified native DLL;
  the production extension excludes the manual E2E app.
- Live validation passed on a Windows Server 2025 Trusted Launch Azure VM:
  - acquired an attested `mtls_pop` token;
  - returned the binding certificate and reusable JSSE `SSLContext`;
  - verified JWT `cnf.x5t#S256` against SHA-256 of the complete leaf certificate
    DER;
  - independently called a token-binding-enforced Azure Key Vault with
    `HttpsURLConnection` and received HTTP 200;
  - retried the same token without its binding certificate and received HTTP
    401 `Unauthorized`;
  - reacquired through MSAL and received `TokenSource.CACHE` with the matching
    live binding context;
  - force-refreshed and received `TokenSource.IDENTITY_PROVIDER`, then completed
    another HTTP 200 Key Vault call;
  - acquired an ordinary `Bearer` token over attested mTLS, exposed no binding
    context, and verified the second acquisition was a cache hit;
  - enabled `MSAL_MI_DISABLE_IMDS_V2` and verified mTLS PoP failed before native
    attestation work.

## Security properties

- Private-key bytes are never exported into Java memory.
- TLS is performed by Java JCA/JSSE, not WinHTTP or Schannel as the HTTP stack.
- Attestation failures fail closed when attestation is requested.
- Ordinary bearer, bearer-over-mTLS, mTLS PoP, attested, and unattested cache
  entries cannot cross-hit.
- Same-key certificate renewal creates a new token-cache partition because the
  complete certificate DER changes.
- Credential-bound redirects are disabled.
- Custom HTTP clients fail fast unless they support request-specific mTLS.
- Retired native key handles are cleaned after certificate expiry.

## MSAL.NET parity gap analysis

A feature-by-feature comparison, intentional platform differences, validation evidence, and prioritized remaining release gaps are documented in [managed-identity-v2-dotnet-java-parity-gap-analysis.md](https://github.com/AzureAD/microsoft-authentication-library-for-java/blob/gladjohn/java-msi-v2-mtls-binding-context-local/msal4j-sdk/docs/managed-identity-v2-dotnet-java-parity-gap-analysis.md).


msal4j-mtls-extensions-e2e/src/main/java/com/microsoft/aad/msal4j/mtls/e2e/ManagedIdentityMtlsPopKeyVaultDevApp.java
```diff
@@ -0,0 +1,362 @@
...
+import java.net.URL;
+import java.nio.charset.StandardCharsets;
+import java.security.MessageDigest;
+import java.util.Base64;
+import java.util.UUID;
+
+/**
+ * Manual Java 8 validation for attested managed identity v2 mTLS PoP with Key Vault.
+ */
+public final class ManagedIdentityMtlsPopKeyVaultDevApp {
```
g2vinay, line 28: The unit coverage on the signing and PKCS10 pieces is solid, but the end-to-end path of binding to SSLContext to a real mTLS handshake is only exercised by this manual dev app, which cannot run in CI because it needs KeyGuard hardware. Consider an automated test that stands up a local TLS server requiring client auth and drives it with a software-backed KeyManager double implementing the same contract, so regressions in the SSLContext wiring (protocol, key manager, cert chain) are caught in CI even though the native KeyGuard path cannot run there.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
```diff
@@ -118,12 +155,41 @@ public ManagedIdentityParametersBuilder claims(String claims) {
             return this;
         }
 
+        /**
+         * Requests the KeyGuard managed identity v2 mTLS PoP flow.
+         */
+        public ManagedIdentityParametersBuilder withMtlsProofOfPossession() {
```
g2vinay, line 192: There is currently no way to ask whether mTLS PoP is actually available before calling. A credential chain like Azure Identity's ManagedIdentityCredential needs to decide up front whether to attempt the mTLS path or fall back, and today the only signal is to call and catch. Consider a lightweight capability probe (for example a static isMtlsPopSupported() or a getManagedIdentityCapabilities() view) that reflects platform, provider presence via ServiceLoader, and the IMDS v2 source, so consumers can branch without triggering a full token attempt.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * Process-local mTLS binding capability associated with an mTLS PoP access token.
+ *
+ * <p>The private key is not exportable. Applications use the returned {@link SSLContext}
+ * with a Java JSSE HTTP stack when calling the downstream resource.</p>
+ */
+public interface IMtlsBindingContext {
+
+    SSLContext sslContext();
```
g2vinay, line 36: Consider also exposing the underlying KeyManager (or a KeyManagerFactory / X509ExtendedKeyManager) here, not just a fully built SSLContext. Most Azure SDK data-plane clients on Java run on Netty (reactor-netty), Vert.x, or the JDK HttpClient, and those stacks want to build their own SslContext from a KeyManager rather than consume a prebuilt JSSE SSLContext. KeyGuardMtlsBindingContext already constructs a CngX509ExtendedKeyManager internally, so surfacing it would let non-JSSE transports wire the binding without reflection or rebuilding. Keeping sslContext() is still useful for the JSSE / HttpsURLConnection path.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -25,8 +26,10 @@ final class AuthenticationResult implements IAuthenticationResult {
     private final String scopes;
     private final AuthenticationResultMetadata metadata;
     private final Boolean isPopAuthorization;
+    private final String tokenType;
+    private final transient IMtlsBindingContext mtlsBindingContext;
```
g2vinay, line 32: Marking the binding context transient is correct since it is process-local and non-serializable. Worth confirming the persisted-cache path explicitly: when an mtls_pop entry is written to a persistent cache and later deserialized, this field comes back null. Please confirm that path fails closed, that is, the SDK never hands back an mtls_pop token without a live binding context (which would look like a usable token but cannot complete the mTLS handshake). Including key_id in the cache key hash is a good guard; a short test around deserialize-then-use would lock the behavior in.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
```diff
@@ -0,0 +1,136 @@
...
+                        e);
+            }
+            loadedLibrary = library;
+            return library;
+        }
+    }
+
+    static Path extractBundledLibrary() throws MtlsMsiException {
+        String architecture = System.getProperty("os.arch", "");
+        if (!"amd64".equalsIgnoreCase(architecture)
```
g2vinay, line 68: The clear typed failure for non-x64 is good and much better than a confusing link error. Since Windows on ARM is increasingly common for Azure VMs, it would help to note the arm64 roadmap here (bundle a second native or document the intended timeline) so consumers know whether they need to design a fallback for that platform today.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/KeyGuardMtlsBindingContext.java
```diff
@@ -0,0 +1,77 @@
...
+    void closeNativeKey() {
+        privateKey.close();
+    }
+
+    private static SSLContext createSslContext(
+            CngRsaPrivateKey privateKey,
+            X509Certificate certificate) {
+        try {
+            CngProvider.installIfAbsent();
+            SSLContext context = SSLContext.getInstance("TLSv1.2");
```
g2vinay, line 74: This pins the binding to TLSv1.2 and passes null trust managers. Two consumption concerns: (1) hardcoding TLSv1.2 prevents TLSv1.3 negotiation for the downstream mTLS call, which several services now prefer; consider requesting the protocol-agnostic "TLS" context and letting the platform negotiate the highest version. (2) null trust managers means the JVM default trust store is used implicitly, which is usually correct but should be documented, so consumers who need a custom trust anchor know they must build their own SSLContext from the exposed KeyManager (see the related note on IMtlsBindingContext).

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
```diff
@@ -0,0 +1,136 @@
...
+
+            directory = Files.createTempDirectory("msal4j-keyguard-attestation-");
+            library = directory.resolve("AttestationClientLib.dll");
+            try (DigestInputStream verifiedResource =
+                         new DigestInputStream(resource, digest)) {
+                Files.copy(verifiedResource, library);
+            }
+
+            String actualHash = toHex(digest.digest());
+            if (!SHA256.equals(actualHash)) {
```
g2vinay, line 102: Good that the extracted DLL is integrity-checked. One gap: a pinned SHA-256 constant that ships in the same JAR as the DLL only proves the bytes match what this build embedded. If the JAR is repackaged, both the DLL and this constant move together, so the check still passes for a substituted binary. For a native library that performs attestation, consider also verifying the Authenticode signature (WinVerifyTrust) so the loaded DLL is proven Microsoft-signed independent of JAR integrity.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
```diff
@@ -0,0 +1,136 @@
...
+        Path library = null;
+        try (InputStream resource =
+                     AttestationLibraryLoader.class.getResourceAsStream(RESOURCE_PATH)) {
+            if (resource == null) {
+                throw new MtlsMsiException(
+                        "Bundled Microsoft.Azure.Security.KeyGuardAttestation " + VERSION +
+                                " native library is missing from the extension JAR.");
+            }
+
+            directory = Files.createTempDirectory("msal4j-keyguard-attestation-");
```
g2vinay, line 93: Extracting the native library to a temp directory and loading it by absolute path is the right approach and avoids PATH-based planting of the DLL itself. The residual Windows risk is dependency sideloading: when AttestationClientLib.dll loads, its own dependency DLLs can still be resolved from the load directory (this temp folder) under the default search order. Consider restricting the extraction directory ACLs to the current user and/or using the safe DLL search directory APIs, so a co-located planted dependency cannot be picked up.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -136,9 +180,168 @@ private AuthenticationResult createFromManagedIdentityResponse(ManagedIdentityRe
...
+            IManagedIdentityMtlsProvider provider,
+            ManagedIdentityMtlsRequest request) {
+        try {
+            return provider.getOrCreateBinding(request);
+        } catch (MsalException e) {
+            throw e;
+        } catch (RuntimeException e) {
+            MsalClientException wrapped = new MsalClientException(
+                    "The managed identity mTLS provider failed.",
+                    MsalError.MANAGED_IDENTITY_MTLS_REQUEST_FAILED);
```
g2vinay, line 330: Several very different conditions all collapse into MsalClientException with MANAGED_IDENTITY_MTLS_REQUEST_FAILED: permanent setup or config errors (wrong MI source, unsupported selector, a custom client missing the marker) and potentially transient runtime failures (the provider or attestation call failing here). A consumer that wants to fall back in a credential chain cannot distinguish "this will never work, stop" from "this failed once, maybe retry or move on" without string matching. Consider mapping permanent errors to a distinct code from transient ones (you already have MANAGED_IDENTITY_MTLS_PROVIDER_UNAVAILABLE, so extending that taxonomy would help) and preserving the underlying MtlsMsiException as the cause so the specific reason is not lost.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsCapableHttpClient.java
```diff
@@ -0,0 +1,13 @@
...
+
+package com.microsoft.aad.msal4j;
+
+/**
+ * Marker contract for custom HTTP clients that honor request-specific mTLS settings.
+ *
+ * <p>Implementations must use {@link HttpRequest#sslSocketFactory()} when it is present
+ * and must not automatically follow redirects for that credential-bound request.</p>
+ */
+public interface IMtlsCapableHttpClient extends IHttpClient {
```
g2vinay, line 17: This marker means any existing custom IHttpClient fails closed for mTLS PoP until it is changed to implement IMtlsCapableHttpClient. That is a reasonable safety default, but it is a real friction point for consumers like Azure Identity, whose HttpPipelineAdapter implements IHttpClient and would otherwise silently lose mTLS support. Two asks: (1) call this out clearly in the changelog and migration notes as a required opt-in for custom clients, and (2) consider shipping a small default decorator that wraps a plain IHttpClient and applies HttpRequest.sslSocketFactory(), so integrators do not each reimplement the same socket-factory plumbing.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
```diff
@@ -100,6 +103,18 @@ public String body() {
         return this.body;
     }
 
+    /**
+     * Returns a request-specific JSSE socket factory, when the request requires mTLS.
+     */
+    public SSLSocketFactory sslSocketFactory() {
```
g2vinay, line 111: SSLSocketFactory is a blocking, java.net.Socket-shaped abstraction. Custom IHttpClient implementations built on async or SSLEngine-based stacks (Netty, the JDK HttpClient, OkHttp) cannot consume an SSLSocketFactory directly, so this contract effectively limits request-scoped mTLS to blocking JSSE clients. Consider carrying the binding as an SSLContext (or the KeyManager) on the request so an async client can derive its own engine, or clearly document that mTLS PoP requires a JSSE-based IHttpClient.

msal4j-mtls-extensions-e2e/src/main/java/com/microsoft/aad/msal4j/mtls/e2e/ManagedIdentityMtlsPopKeyVaultDevApp.java
gladjohn: Added a software-backed local mTLS server/client handshake test for CI coverage; the real KeyGuard flow was also validated manually on the VM because the Java pipeline cannot attach that specialized VM.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityParameters.java
gladjohn: Added getManagedIdentityCapabilities() with IMDS v2/provider/KeyGuard discovery, NONE/SOFTWARE/KEY_GUARD strength reporting, retryable failed probes, and a minimum-strength request floor for credential chains.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
gladjohn: Exposed the CNG-backed X509ExtendedKeyManager alongside SSLContext so Netty, JDK HttpClient, and other transports can build their own TLS context.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
gladjohn: Added fail-closed deserialization so an mtls_pop result cannot be restored or returned without its live process-local binding context, with coverage for the persisted-result path.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
gladjohn: Documented Windows ARM64 as unsupported in this release; adding it requires an upstream ARM64 AttestationClientLib native artifact, so callers should retain a platform fallback for now.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/KeyGuardMtlsBindingContext.java
gladjohn: Kept TLS 1.2 intentionally because the service does not yet request the required client certificate during TLS 1.3 negotiation; we are working with the service team on TLS 1.3 support, and default/custom trust behavior is now documented.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
gladjohn: Added fail-closed WinVerifyTrust Authenticode validation before Native.load in addition to the pinned SHA-256 check, and locally confirmed the bundled DLL signer is Microsoft Corporation.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
gladjohn: Restricted the extraction directory ACL to the current Windows user and now fail closed when Windows ACL support is unavailable, while continuing to load the verified DLL by absolute path.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
gladjohn: Added distinct permanent error codes for unsupported source/selector and custom HTTP clients, a dedicated minimum-strength error, retryable capability discovery, and preserved provider failures as the cause.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsCapableHttpClient.java
gladjohn: Documented the required IMtlsCapableHttpClient opt-in and migration path; I did not add a universal decorator because it cannot safely retrofit request-specific TLS into arbitrary Netty, OkHttp, or JDK transports.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/HttpRequest.java
gladjohn: HttpRequest now carries the request-specific SSLContext as well as the legacy SSLSocketFactory, and the binding exposes its X509ExtendedKeyManager for SSLEngine-based transports.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public String chooseClientAlias(
+            String[] keyTypes,
+            Principal[] issuers,
+            Socket socket) {
+        return supportsAnyKeyType(keyTypes) ? ALIAS : null;
+    }
+
+    @Override
+    public String chooseEngineClientAlias(
```
g2vinay, line 40: Good to see `chooseEngineClientAlias` overridden here. That is the path our default HTTP transport (reactor-netty) and the JDK `HttpClient` actually use, since both drive TLS through an `SSLEngine` rather than `SSLSocket`. Could we add (or point me to) a test that exercises alias selection through a real `SSLEngine` handshake, not just `SSLSocket`? That is the path most Azure SDK consumers will hit, and it is easy for the engine path to regress without a test that catches it.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
```diff
@@ -0,0 +1,95 @@
...
+        this.certificateChain = new X509Certificate[]{certificate};
+    }
+
+    @Override
+    public String[] getClientAliases(String keyType, Principal[] issuers) {
+        return supportsKeyType(keyType) ? new String[]{ALIAS} : null;
+    }
+
+    @Override
+    public String chooseClientAlias(
```
g2vinay, line 32: Question on alias selection: do `chooseClientAlias` and `chooseEngineClientAlias` return the binding alias regardless of the `issuers` argument? The MI leaf is attested/self-issued and will not chain to a public CA, so during the handshake the server often advertises an empty or non-matching CA list. If selection is filtered by `issuers`, these methods would return null and no client cert gets sent, which fails silently as an empty client Certificate message. Want to confirm the selection is unconditional here.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -136,9 +185,192 @@ private AuthenticationResult createFromManagedIdentityResponse(ManagedIdentityRe
...
+        }
+
+        String scope = managedIdentityParameters.resource().endsWith("/.default")
+                ? managedIdentityParameters.resource()
+                : managedIdentityParameters.resource().replaceAll("/+$", "") + "/.default";
+        Map<String, String> body = new HashMap<>();
+        body.put("grant_type", "client_credentials");
+        body.put("client_id", binding.clientId());
+        body.put("scope", scope);
+        body.put("token_type", "mtls_pop");
```
g2vinay: Two contract questions from the consumer side (azure-identity / azure-core), since this is where the `mtls_pop` flow is set up:

1. Does the same client cert need to be presented on the downstream data-plane call to the resource (RFC 8705 `cnf` / `x5t#S256` matching), or only here on the token-acquisition call?
2. What `Authorization` scheme should we send with an `mtls_pop` token on the resource call, still `Bearer <token>`, or a PoP/mtls scheme?

Both answers decide how we set the header and pin the cert in our pipeline, so it would help to have them stated explicitly.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -25,8 +28,19 @@ final class AuthenticationResult implements IAuthenticationResult {
...
```
g2vinay, line 32: Marking the binding `transient` is right for not serializing key material, but it means a result restored from the persistent token cache comes back with `mtlsBindingContext == null`. I see the constructor guard just below throws when `tokenType` is `mtls_pop` and the binding is null, which is good, but does deserialization actually go through that constructor, or could a cached `mtls_pop` token resurface without its binding? Want to make sure we cannot silently end up with an `mtls_pop` token and no cert on the wire after a cache reload.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
```diff
@@ -0,0 +1,95 @@
...
+package com.microsoft.aad.msal4j.mtls;
+
+import javax.net.ssl.SSLEngine;
+import javax.net.ssl.X509ExtendedKeyManager;
+import java.net.Socket;
+import java.security.Principal;
+import java.security.PrivateKey;
+import java.security.cert.X509Certificate;
+
+final class CngX509ExtendedKeyManager extends X509ExtendedKeyManager {
```
g2vinay, line 13: Thread-safety check: these KeyManager callbacks get invoked concurrently when a connection pool opens several handshakes in parallel. Is this class, and the underlying CNG / native signing path it delegates to, safe for concurrent use from multiple threads? If there is shared native state per key handle, it would be good to confirm the signing calls are either stateless or synchronized.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
```diff
@@ -0,0 +1,46 @@
...
+     * context can combine this key manager with their own trust configuration.</p>
+     */
+    X509ExtendedKeyManager keyManager();
+
+    X509Certificate bindingCertificate();
+
+    /**
+     * Base64URL-without-padding SHA-256 digest of the complete leaf certificate DER.
+     */
+    String keyId();
```
g2vinay, line 53: For connection pooling on our side we need a stable identity for the binding so we can key/segment pools and detect rotation. A few things to confirm about `keyId()`: is it stable for the life of a token and unique per binding (it looks derived from the full leaf cert DER, so renewal changes it, which is what we want)? On rotation, is the previous native key handle disposed/closed so we do not leak CNG handles? And does `keyManager()` return a stable instance for a given token rather than a new one per call?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
```diff
@@ -0,0 +1,46 @@
...
+     */
+    default MtlsBindingStrength bindingStrength() {
+        return MtlsBindingStrength.SOFTWARE;
+    }
+
+    /**
+     * Returns a ready-to-use JSSE context. The context uses the JVM's default trust
+     * managers.
+     */
+    SSLContext sslContext();
```
g2vinay, line 36: Confirming the intended split: `sslContext()` is the ready-to-use context for MSAL's own token call (and for simple JSSE consumers), while for resource calls that need custom trust or a different transport we should take `keyManager()` and build our own context, pairing it with the platform default trust managers. If that is right, a line in the javadoc steering consumers away from reusing the token-acquisition `SSLContext` for data-plane traffic would help, since reusing it would also apply MSAL's trust configuration to resource calls.

g2vinay (commented):
Re-reviewed after the latest fix commit. The mTLS PoP surface looks solid. Most of my remaining notes are consumer-side integration questions from the Azure SDK for Java side (azure-identity / azure-core), where these bindings will be wired into the HTTP pipeline across pluggable transports.

The higher-priority items are the SSLEngine alias path and issuer-independent alias selection in `CngX509ExtendedKeyManager` (our default reactor-netty and JDK HttpClient transports are engine-based and MI certs are self-issued), the resource-facing contract (whether the same cert must ride the data-plane call and what Authorization scheme to use for `mtls_pop`), and the transient-binding cache-reload path. The rest are clarifications on thread-safety, binding identity/rotation lifecycle, and the `keyManager()` vs `sslContext()` trust scope. None are blocking; mostly confirmations and a couple of test/javadoc asks. Thanks.

Robbie-Microsoft (dismissed):

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
```diff
@@ -93,8 +98,13 @@ HttpURLConnection openConnection(final URL finalURL)
...
 
-            if (sslSocketFactory != null) {
-                httpsConnection.setSSLSocketFactory(sslSocketFactory);
+            SSLSocketFactory effectiveSslSocketFactory =
+                    requestSslSocketFactory != null ? requestSslSocketFactory : sslSocketFactory;
+            if (effectiveSslSocketFactory != null) {
+                httpsConnection.setSSLSocketFactory(effectiveSslSocketFactory);
+            }
+            if (requestSslSocketFactory != null) {
+                httpsConnection.setInstanceFollowRedirects(false);
```
Robbie-Microsoft: Could we add a request-scoped no-redirect policy for credential-bearing IMDS calls? A 307 can replay `/issuecredential`'s CSR and attestation token to another host.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/ImdsV2Client.java
```diff
@@ -0,0 +1,264 @@
...
+            throw new MtlsMsiException(
+                    "KeyGuard attestation failed; no attestation token was produced.");
+        }
+        String body = "{\"csr\":\"" + escapeJson(csr) + "\""
+                + (isBlank(attestationToken)
+                    ? ""
+                    : ",\"attestation_token\":\""
+                        + escapeJson(attestationToken) + "\"")
+                + "}";
+        String response = execute(
```
Robbie-Microsoft: Could we call `validateImdsOrigin` before consuming the `/issuecredential` response? It controls the binding certificate and token endpoint; the marker provides useful defense-in-depth.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
```diff
@@ -37,6 +37,10 @@ public static ManagedIdentityResponse fromJson(JsonReader jsonReader) throws IOE
                     case "expires_on":
                         response.expiresOn = reader.getString();
                         break;
+                    case "expires_in":
+                        response.expiresOn = String.valueOf(
+                                (System.currentTimeMillis() / 1000) + reader.getLong());
```
Robbie-Microsoft: `expires_in` may be a JSON string, so `getLong()` can fail. Please accept string or numeric values and use `expires_in` only when `expires_on` is absent.

msal4j-mtls-extensions/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <keyguard.attestation.version>1.1.5</keyguard.attestation.version>
+        <keyguard.attestation.sha256>90dfcce20e1a74519b49796eeee17e6e59a257c3acf754f454a49380d28a568b</keyguard.attestation.sha256>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>com.microsoft.azure</groupId>
+            <artifactId>msal4j</artifactId>
+            <version>1.26.0</version>
```
Robbie-Microsoft: Published `msal4j:1.26.0` lacks these mTLS SPI classes. Please depend on the first coordinated, unreleased SDK version containing them.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngProvider.java
```diff
@@ -0,0 +1,83 @@
...
+        super(PROVIDER_NAME, PROVIDER_VERSION, PROVIDER_INFO);
+
+        putSignature("SHA256withRSA", CngSignatureSpi.Sha256WithRsa.class,
+                "SHA-256withRSA");
+        putSignature("SHA384withRSA", CngSignatureSpi.Sha384WithRsa.class,
+                "SHA-384withRSA");
+        putSignature("SHA512withRSA", CngSignatureSpi.Sha512WithRsa.class,
+                "SHA-512withRSA");
+        putSignature("RSASSA-PSS", CngSignatureSpi.RsaSsaPss.class,
+                "SHA256withRSAandMGF1", "SHA384withRSAandMGF1", "SHA512withRSAandMGF1");
```
Robbie-Microsoft: The SHA-384/512 PSS aliases use an SPI initialized with SHA-256 defaults. Please add digest-specific implementations or remove these aliases.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -136,9 +228,217 @@ private AuthenticationResult createFromManagedIdentityResponse(ManagedIdentityRe
...
+        Map<String, String> body = new HashMap<>();
+        body.put("grant_type", "client_credentials");
+        body.put("client_id", binding.clientId());
+        body.put("scope", scope);
+        String requestedTokenType = managedIdentityParameters.requestOverMtls()
+                ? "bearer" : "mtls_pop";
+        body.put("token_type", requestedTokenType);
+        AuthenticationResult result;
+        try {
+            result = tokenRequestExecutor.executeTokenRequest(
```
Robbie-Microsoft, line 514: Rejected bindings remain cached, and the claims-challenge path at lines 147–153 also reuses the current certificate. Can we invalidate and remint with `bypass_cache=true`, including repeated remediation for qualifying `invalid_client` responses?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
```diff
@@ -129,6 +145,78 @@ Boolean isPopAuthorization() {
...
+                .accessToken(accessToken)
+                .expiresOn(expiresOn)
+                .extExpiresOn(extExpiresOn)
+                .refreshToken(refreshToken)
+                .refreshOn(refreshOn)
+                .familyId(familyId)
+                .idToken(idToken)
+                .accountCacheEntity(accountCacheEntity)
+                .environment(environment)
+                .scopes(StringHelper.isBlank(scopes) ? defaultScopes : scopes)
```
Robbie-Microsoft: Could both `withScopes` and `withMtlsBindingContext` always use the original managed-identity resource? Retaining `resource/.default` causes later bearer-over-mTLS and PoP cache misses.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
gladjohn: Addressed: `IMtlsBindingContext.keyManager()` now exposes the stable `X509ExtendedKeyManager` for custom-trust and non-JSSE transport integration.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/KeyGuardMtlsBindingContext.java
gladjohn: Documented the trust split: `sslContext()` uses JVM-default trust, while custom trust is composed with `keyManager()`. TLS 1.2 remains intentional because the current service path did not reliably request the client certificate under TLS 1.3.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
gladjohn: Added real client/server `SSLEngine` mutual-TLS coverage in `KeyGuardMtlsBindingContextSslTest`, including verification of the presented client certificate.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
gladjohn: Confirmed and covered by tests: client alias selection ignores issuer hints and filters only for compatible RSA key types, including empty and non-matching issuer lists.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
gladjohn: Covered fail-closed: mTLS acquisition resolves the live binding before cache lookup and reattaches it on cache hits. Java deserialization also rejects an `mtls_pop` result without its process-local binding context, with persisted-cache coverage.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngX509ExtendedKeyManager.java
gladjohn: Addressed: native signing and key close share the same per-key synchronization boundary, so a handle cannot be released during `NCryptSignHash`. Added concurrency coverage for close waiting on an in-flight operation.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
gladjohn: Confirmed: `keyId()` is SHA-256 over the full leaf-certificate DER, and `keyManager()`/`sslContext()` are stable per binding generation. Retired generations are cleaned after certificate expiry; rejected-generation retention is now bounded.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/IMtlsBindingContext.java
gladjohn: Yes. The Javadocs now describe `sslContext()` as the ready-to-use JVM-default-trust path and `keyManager()` as the integration point for custom trust or other transports.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/DefaultHttpClient.java
gladjohn: Addressed in `9bda984`: IMDS v2 requests now carry an explicit request-scoped no-redirect policy, including plain-HTTP `/issuecredential`, with tests.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/ImdsV2Client.java
gladjohn: Addressed in `9bda984`: `/issuecredential` now validates the IMDS `Server` marker before consuming the certificate or token endpoint, with negative coverage.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/ManagedIdentityResponse.java
gladjohn: Addressed in `9bda984`: `expires_in` accepts JSON string or number forms and is used only when `expires_on` is absent. Both field orders are covered.

msal4j-mtls-extensions/pom.xml
gladjohn: Addressed in `9bda984`: the SDK and dependent key-attestation/E2E modules now coordinate on the unreleased `1.27.0-SNAPSHOT` SDK version.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/CngProvider.java
gladjohn: Addressed in `9bda984`: SHA-256/384/512 PSS aliases now use digest-specific SPI implementations with matching digest and salt defaults, with tests.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
gladjohn: Addressed in `9bda984`: qualifying `invalid_client`, TLS, and connection-reset failures invalidate the rejected generation, remint, and retry exactly once. Rejected-generation retention is bounded. We intentionally did not add repeated remediation or `bypass_cache=true`; current MSAL.NET still tracks that query parameter as a TODO.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AuthenticationResult.java
gladjohn: Addressed in `9bda984`: both result helpers now preserve the caller original managed-identity resource rather than retaining the ESTS-normalized `/.default` scope, with regression coverage.

changelog.txt
```diff
@@ -1,3 +1,17 @@
+Unreleased
```
Avery-Dunn: No need to update the changelog with unreleased stuff (though this might be fine while we're just doing preview/snapshot releases)

build/run-java-msi-v2-mtls-devapp.ps1
```diff
@@ -0,0 +1,88 @@
+[CmdletBinding()]
```
Avery-Dunn, line 1: Could this script be moved to the 'build' folder along with the other script in this PR? Seems odd to have a script like this in the root folder.

changelog.txt
gladjohn: Removed the unreleased changelog entry in b58b8de.

build/run-java-msi-v2-mtls-devapp.ps1
gladjohn: Moved the script to build/run-java-msi-v2-mtls-devapp.ps1 and updated all references in b58b8de.

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthHttpRequest.java
```diff
@@ -40,6 +44,11 @@ public HttpResponse send() throws IOException {
                 this.url.toString(),
                 httpHeaders,
                 this.query);
+        if (sslContext != null) {
```
Robbie-Microsoft, line 47: Could we set `followRedirects(false)` whenever this request carries a request-specific `SSLContext` or `SSLSocketFactory`? `IMtlsCapableHttpClient` requires credential-bound requests not to redirect, but the request currently exposes `followRedirects=true`, giving custom transports conflicting policy signals.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
```diff
@@ -0,0 +1,164 @@
...
+            String actualHash = toHex(digest.digest());
+            if (!SHA256.equals(actualHash)) {
+                Files.deleteIfExists(library);
+                Files.deleteIfExists(directory);
+                throw new MtlsMsiException(
+                        "Bundled Microsoft.Azure.Security.KeyGuardAttestation " + VERSION +
+                                " failed SHA-256 verification.");
+            }
+
+            AuthenticodeVerifier.verify(library);
```
Robbie-Microsoft, line 110: Please clean up the extracted DLL and temporary directory when Authenticode verification fails, because that `MtlsMsiException` bypasses both the `IOException` cleanup path and the subsequent `deleteOnExit` registration, leaving another extraction behind after each failed attempt.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/KeyGuardManagedIdentityMtlsProvider.java
```diff
@@ -0,0 +1,364 @@
...
+                retained.add(generation);
+            }
+        }
+        return retained;
+    }
+
+    static void trimRejectedGenerations(
+            List<BindingGeneration> generations) {
+        while (generations.size() > MAX_REJECTED_GENERATIONS) {
+            generations.remove(0).context.closeNativeKey();
```
Robbie-Microsoft: `RETIRED` contains both normally rotated and rejected generations, so this count-based cap can close an unexpired key still referenced by a previously returned `mtls_pop` result; could we retain each generation until certificate expiry or track that no returned result can still use it before releasing the native key?

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
Robbie-Microsoft: `WinVerifyTrust` confirms that the DLL has a trusted Authenticode signature, but it does not verify that Microsoft is the signer. Could we validate the signer identity as well, or narrow the Microsoft-signed assurance accordingly?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/AcquireTokenByManagedIdentitySupplier.java
```diff
@@ -136,9 +180,137 @@ private AuthenticationResult createFromManagedIdentityResponse(ManagedIdentityRe
...
+                + (managedIdentityParameters.attestationSupport()
+                        ? ":att1" : ":att0");
+        ManagedIdentityMtlsRequest request = new ManagedIdentityMtlsRequest(
+                queryName,
+                queryValue,
+                bindingCacheKey,
+                requestContext.correlationId(),
+                httpClient,
+                managedIdentityParameters.attestationSupport());
+        return ManagedIdentityMtlsProviderLoader.load().getOrCreateBinding(request);
```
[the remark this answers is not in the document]
Robbie-Microsoft: One remaining failure path is provider discovery: `ServiceLoader.hasNext()` or `next()` can throw `ServiceConfigurationError`, which is an `Error` and bypasses the `RuntimeException` translation. Could we map that specifically to `MANAGED_IDENTITY_MTLS_PROVIDER_UNAVAILABLE` as well?

msal4j-sdk/src/main/java/com/microsoft/aad/msal4j/OAuthHttpRequest.java
gladjohn: Fixed in `5144a82`: OAuth requests carrying a request-specific `SSLContext` or `SSLSocketFactory` now expose `followRedirects=false` to every transport. Added a regression test that captures the emitted `HttpRequest` and verifies the no-redirect policy.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/AttestationLibraryLoader.java
gladjohn: Fixed in `5144a82`: runtime failures from Authenticode verification now clean up both the extracted DLL and its temporary directory before the original exception is rethrown. Added cleanup coverage.

msal4j-mtls-extensions/src/main/java/com/microsoft/aad/msal4j/mtls/KeyGuardManagedIdentityMtlsProvider.java
gladjohn: Fixed in `5144a82`: removed count-based retirement of rejected generations. Returned binding contexts now retain their native key until certificate expiry, when normal retired-generation cleanup closes the handle.

